Issue #879, opened 2014-09-02
Changing the cursor over clickable elements is broken


PragTob:
Normally the cursor should change to a hand or something, when over a clickable element (such as a link).

It doesn't do that for me right now on master (tried with the menu of simple-manual). As I've been fooling around with swt 4.4 (although I uninstalled it), could someone please verify that it is indeed broken or it's just me/Linux.

Thanks.

Tobi

Issue #441, opened 2013-09-26
Errors with new glib version


PragTob:
Linux Mint Debian Edition had a new update pack released yesterday and that has a new GLib version, on top of which SWT works. 

Ever since then I get these errors:

```
(SWT:4528): GLib-GObject-CRITICAL **: g_closure_add_invalidate_notifier: assertion `closure->n_inotifiers < CLOSURE_MAX_N_INOTIFIERS' failed
```

It seems like it might be fixed upstream. So when I get to it I'm gonna work on updating the SWT gem, send dan a pull request and hopefully he'll merge it.

Because right now with this, I can't really run the SWT specs, which sucks :-(

Cheers,
Tobi

jasonrclark:
Been a couple years, any chance this is fixed (if easy to confirm?)

I don't have a desktop linux around to test against, although I may remedy that if need be via VirtualBox...

PragTob:
Sadly no. After each system glib update and each SWT update (as in the current PR) I run the specs and am all hopeful.. just to be disappointed.

I also had multiple full OS reinstalls in between on different machines (all Linux Mint though) so there is 0 chance that it's just a local me problem :( 

I guess what they fixed was something else :|

PR #880, opened 2014-09-02
Update Swt to 4.4

PragTob:
Title says it all. Note the versioning change in the swt gem, which now mirrors the actual SWT version :-)

First tests indicate that it works fine (at least with some of our big apps). Unfortunately it does not seem to get rid off #441 ...

Also, for me on Linux my scrollbar looks slightly awkward but no biggie (see screenie)

screenshot from 2014-09-02 19 39 55

At first I thought it also introduced #879 but that seems to happen on master as well.

if you want try it out, please go ahead :-) (of course you will have to do bundle install to get the newest version)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ notifications:
     on_failure: change # [always|never|change] # default: always
 before_script:
   - "export DISPLAY=:99.0"
+  - "export SWT_GTK3=0"
   - "sh -e /etc/init.d/xvfb start"
   - "sleep 3"
 after_script:


ccoupe:
Gtk3 is **highly** version dependent - what works in Gtk-3.6. may not work in Gtk-3.15. Fix that, It may fail on Gtk-3.20.  Fix that and it won't work on Gtk-3.6 or Gtk-3.15. 

Gtk 3.20 in particular changed the rules for Shoes 3 (and the swt developers). I'm sorry I moved Shoes 3.3 to Gtk 3.

PR #989, opened 2014-12-23
Add shoes-package (extracted from furoshiki)

wasnotrice:
This is the companion to shoes/furoshiki#14. It extracts the shoes-specific code from furoshiki, so that we have no more circular dependencies. Furoshiki is generic and doesn't depend on Shoes.

Not completely certain that I've got the abstractions right, or that the use of inheritance vs mixin vs duck typing is ideal yet. Mainly, I'm shooting for separation between shoes and furoshiki. We can improve the interface at a later point. The idea is that Furoshiki has the general logic for packaging jars and apps (so far), and shoes-package injects the shoes-specific requirements/handling into the furoshiki workflow. Other frameworks could inject different requirements/handling

I suspect this needs some love regarding the versioning fixes that @jasonrclark recently pushed.

Comments???


wasnotrice:
@jasonrclark thanks for testing! Given the fact that packaging on master is totally broken, I think we can head this toward merge sooner than later. I do still want to get to packaging Shoes apps as Mac .apps (which did work previously) and then of course .exes for Windows and .runs or whatever format we decide is best for Linux.

But let's not hold this up for what's undone. @PragTob does this work for you now?

Happy New Year! :tada:

jasonrclark:
So do you want to get apps working again before we do pre3? Any way I can help?

While I'd love to get Windows and Linux done too, I'm keen to get the new gem breakdown into the wild. Are you cool with that work carrying over to pre4?

Happy 2015 :tada: :rocket:

wasnotrice:
Let's merge this in and not hold up a release. I think it's fine to carry over to next release.

PragTob:
It.. works! :)

Thanks so much @wasnotrice !

So I take it packaging .app is broken atm as well? Would kind of like to get that in pre3 but not necessary as it was always marked as experimental :)

wasnotrice:
Well, lo and behold I packaged a Shoes app as a Mac .app just to see how badly I had broken it and...it worked. :boom:

jasonrclark:
:metal: :rocket:

Issue #820, opened 2014-07-30
List Box should update items when style[:items] gets updated


KCErb:
As noted in the mailing list you can update the `:items` of a list_box like a normal array (via `<<` for example) but the gui isn't catching the change.

Take a look at this:
```ruby
Shoes.app do
  l = list_box items: ["a", "b", "c"]
  p l.items
  l.items << "d"
  p l.items
end
```

PragTob:
Should be done, right @KCErb ?

KCErb:
Right, nearly forgot!

PR #998, opened 2014-12-25
Using a ProxyArray to get ListBox items handled correctly

KCErb:
Meant to not make a new branch, but I accidentally screwed up the old one when trying to revert it and decided to delete it. (See #992).

Anyways, to repeat: this intends to address #820.

This time my approach is an ArrayProxy that inherits from BasicObject and delegates everything to Array. 

Please let me know if this approach is good, needs tweaking, or needs a totally new direction (for example, how about `class ProxyArray < Array`?).

I have some questions / comments. I'll put those in-line before :poultry_leg: :sweet_potato: :fork_and_knife:  time!


jasonrclark:
Hey @KCErb sorry for the delay in getting back to you on this!

Tested it out, and I think that `after_do` is just getting in your way here. It's wrapping was making it so that you'd get an `Enumerator` back from the `super` call to `map!` which then ended up not getting evaluated.

In fact, I think we can simply update your `method_missing` to look more like this:

```ruby
    def method_missing(method, *args, &block)
      res = super(method, *args, &block)
      gui.update_items #if ARRAY_MODIFYING_METHODS.include? method

      case res
      when ProxyArray, Array
        self
      else
        res
      end
    end
```

After we `super`, we just call `gui.update_items` directly--no need for the extra callback `after_do` brings to the table.

You'll note too that I commented out the `if` clause there... I think we might want to just blanket all the methods, because I could spot at least a half dozen additional ones beyond your list that would need to be included, and that seems like a losing game to me in the long run. Plus, since this only happens when app code manipulates the list, I don't see that there's much perf concern at this stage without something demonstrating it's a problem.

How's that all sound?

KCErb:
Hey that's great! Thanks Jason, I was really stumped there! Looking at it now, it seems obvious that by adding method missing to the Proxy, `after_do` becomes unnecessary. I played around with ditching the `after_do` callback and it seems to behave as expected.

>  I don't see that there's much perf concern at this stage without something demonstrating it's a problem.

Agreed! I think I heard a saying somewhere about premature optimization being the root of all evil? 

Now I'll just write up a spec or two and get this in.

KCErb:
Alright, a little slow on this one, but I think it's ready to go. I fixed it up as per our discussion and added a couple of specs that look for `update_items` to get called on `gui`. I came up with these after playing around on master and finding specs that failed there but worked on my branch.

Sorta reverse TDD I guess :)

Let me know if this could use more work or is ready to :arrow_forward:

shoes-swt/lib/shoes/swt/list_box.rb
```diff
@@ -28,7 +28,7 @@ def initialize(dsl, parent)
       end
 
       def update_items
-        @real.items = @dsl.items.map(&:to_s)
+        @real.items = @dsl.items.to_a
```
KCErb: Looking back at this, I'm not sure if the `to_s` here is just a thoughtful guard or actually important. After the `to_a` I could add back the map. Opinions?

jasonrclark: Looks to me like Shoes 3 did coercion to string, so probably best keep it for compatibility. I can imagine someone passing things that aren't strings along (in particular primitives like numbers), and it'd be frustrating if it didn't do the "obvious" thing.

jasonrclark:
Apart from the `to_s` bit, I only had one other question that occurred to me as I read through... Is the `SimpleDelegator` actually buying us anything since we're intercepting `method_missing` anyway? Rather than calling `super(method, *args, &block)`, you'd just call `@array.public_send(method, *args, &block)`, and grab `@array = array` in initialize instead of `super(array)`.

Main reason I suggest this is that on further reflection, it's a bit weird that your `method_missing` trick works here. If we're delegating, why is the method missing? The reason happens to be that the underlying `Delegator` class is implemented in terms of `method_missing` (see https://github.com/ruby/ruby/blob/trunk/lib/delegate.rb#L82-L96)! So your override just steps in front of the version of `method_missing` that `SimpleDelegator` itself uses. In the unlikely case that this behavior changed (i.e. the underlying delegator starts defining methods for you for performance reasons), our code here would break.

All right, that was a bit meta and long-winded, but once those two bits are done I think we're set to merge!

KCErb:
Hi Jason, if I understand your comment correctly it essentially takes us back to [my first commit](https://github.com/shoes/shoes4/commit/10603aa246469063dc892a71db79a75896659c13) right? 

If so, I think we picked up SimpleDelegator because of the goodies it gives us in terms of making ProxyArray equal to Array. So if we don't use SimpleDelegator we'd have to roll those ourselves.

Or am I off?

shoes-swt/lib/shoes/swt/list_box.rb
KCErb: Yup, that totally makes sense, I'll fix it.

jasonrclark:
Had forgotten the equal bit was part of why we used `SimpleDelegator`, but you're totally right. In light of that, carry on! I'd say just make the `to_s` change, and then I'm happy to see this :ship:

KCErb:
OK, I added the string conversion and a spec for it. 

btw I was getting a weird bug with the specs. After both shoes-core and swt specs had run (and passed) this error was thrown:
```
LoadError: no such file to load -- furoshiki/configuration
```

I'll play around with it for a bit (try stuff like `bundle`) and open a proper issue if the problem persists.

shoes-swt/spec/shoes/swt/list_box_spec.rb
```diff
@@ -26,10 +26,10 @@
   end
 
   it "should call 'items' when updating values" do
-    subject # creation already calls update_items once
     allow(dsl).to receive(:items).and_return ["hello"]
     subject.update_items
-    expect(real).to have_received(:items=).with(["hello"])
+    # creation already calls update_items once
+    expect(real).to have_received(:items=).with(["hello"]).twice
```
KCErb: Changed this because it made more sense to me. Agreed?

jasonrclark: That's fine by me.

jasonrclark:
With the `furoshiki` thing, what versions of that gem do you have installed, and where's the `Gemfile` and `shoes-swt` (or `shoes-package`?) gemspec pointing to `furoshiki`?

I think that class moved with the 0.3.0 release, so maybe you're just behind/mismatched in some way?

KCErb:
Hmm, I think I know what happened.

I started this PR before the pre3 release and pull / merge / `bundle`d on `master`. Then I merged `master` into my `list-box` branch. 

I'm guessing that I should have `bundle`d again after the merge while on the `list-box` branch. I'm not sure I really understand why that would make a difference, but I just tried to `bundle` again, only this time while I was on the `list-box` branch and it acted as though grabbing `furoshiki 0.3.0` was a new thing (`Installing` instead of `Using`).

Anyways, my specs run as expected now. :)

jasonrclark:
Hurray! Nice work on this @KCErb. Lots if metaprogrammatic back and forth, but I like where it landed a lot. :rocket: :sparkles:

Issue #1009, opened 2015-01-01
Clipboard doesn't work, and scoping is funky


wasnotrice:
Looking at clipboard to help out with [this Stack Overflow question](http://stackoverflow.com/questions/27514457/shoes-clipboard-access-from-app), I noticed a couple of problems with clipboard:

- [x] Clipboard spec is being skipped because of #398, but we really need to be running this spec, because
- [x] Clipboard spec fails because Java doesn't like Ruby-style accessors in this context for some reason
- [ ] the `clipboard=` method isn't accessible without `app` or `self` inside a block

Take this example app:

```ruby
Shoes.app do
  edit_box do |clip|
    clipboard = clip.text
  end
end
```

This should copy the contents of the edit_box to the clipboard on every change...but it doesn't. You have to add a self:

```ruby
Shoes.app do
  edit_box do |clip|
    self.clipboard = clip.text
    # Or, alternatively
    # app.clipboard = clip.text
  end
end
```

Surprising, isn't it? I think we probably want clipboard to act like element methods (`para`, etc) and be accessible within blocks, don't you agree?

wasnotrice:
I should add that `app.clipboard = clip.text` works just like `self.clipboard = clip.text`

PragTob:
Isn't the last one kind of a ruby feature? I mean the inaccessibility there? The Ruby parser can't tell whether you create a local variable or want to call a method so it chooses the local variable route if you don't use an explicit receiver.

I fear there's not much we can do there... :-(

wasnotrice:
Yeah I think you're right. That's kind of a failure of the DSL I think, but let's not worry about that one for now.

PR #1010, opened 2015-01-01
Fix SWT clipboard accessors and uncomment spec

wasnotrice:
Addresses #1009, starting with the spec/accessor issues


PragTob:
Cool thanks!

PR #1011, opened 2015-01-02
Handle relative and missing image paths

### LINKED ISSUE #826: Image with bad or relative file path
**Edit - rewrote question so that this doesn't partially become a duplicate of #292**

As noted [here](https://github.com/shoes/shoes4/pull/819#issue-39096262)
`image` should throw an error if it's given a bad file path, and it'd be nice if it would accept relative file paths as well.

This might go nicely with #292

lumeet:
* Add support for relative (to Shoes::DIR) image paths
  * Add FileNotFoundError and raise it if an image file cannot be found
  * Slightly rename initializer argument to better indicate its contents

This hopefully fixes #826.


jasonrclark:
This looks pretty great to me. Thanks @lumeet :heart:

My only question is about the relative paths. Do we really want them relative to the Shoes::DIR? For most people that'll be the gem path... Not super helpful. Seems more useful to people if it was relative to current working for (ie File.expath_path)

What do you think?

PragTob:
I have another idea - not only File.expand_path but the current working directory (from where the script was started)

shoes-core/lib/shoes/image.rb
```diff
@@ -21,9 +23,31 @@ def path
...
+      return path_or_data if raw_image_data?(path_or_data)
+      return path_or_data if url?(path_or_data)
+      absolute_file_path(path_or_data)
+    end
+
+    def absolute_file_path(path)
+      path = File.join(Shoes::DIR, path) unless Pathname(path).absolute?
+      fail FileNotFoundError, "#{path} not found." unless File.exist?(path)
+      path
+    end
```
PragTob: I believe some of the above methods are copied from the SWT/backend side of things or semi copied, right? If so it'd be nice to remove them from there and use those methods here or another way to avoid duplication

PragTob: I'm stupid, of course you did that. Sorry.

PragTob:
Looks good - thanks a lot for your work! :) Please ignore my comments, I can be stupid :P

lumeet:
Damn it! I had similar thoughts but never really did anything about it. :) So, it's going to be `Dir.pwd` instead of `Shoes::DIR`? I guess that's what at least @PragTob has in mind.

jasonrclark:
Pretty sure that if you just call File.expand_path(path) it takes care of the Dir.pwd internally.

wasnotrice:
I agree with @PragTob...looks great, and we should create the absolute path relative to Dir.pwd. The way you've set it up, if I want an image that's bundled with Shoes, I can pass `File.join(Shoes::DIR, 'path/to/my_image.jpg')` and that'll work too. Thanks!!!

jasonrclark:
On the build failure that's not you. A merge/rebase to current master should fix it.

lumeet:
Yes, it seems that `File.join(Dir.pwd, relative_path) == File.expand_path(relative_path)`. Should I change it once more? :)

jasonrclark:
I'm not bothered either way. Will give this a quick test when I can and get this merged. Thanks again :sparkles:

jasonrclark:
This all checks out. Thanks again for the contribution @lumeet!

PR #1012, opened 2015-01-03
Move default args

### LINKED ISSUE #968: Move default arguments from dsl.rb to the DSL classes
As discussed [in another issue](https://github.com/shoes/shoes4/pull/961#discussion_r21145161) we probably want to move the assignment of default values for parameters from our dsl.rb over to the specific domain objects where appropriate.

As @jasonrclark put it:

> Keeping the dsl.rb layer as thin as possible seems like a win.

And I agree.

Marked as newcomer friendly as it's a good way to get to know the interplay between dsl.rb and the domain objects. Also it does not have too many intersections with other code.

Happy to help, clarify and review stuff :)

graves:
Hoping to resolve #968

Hi, I'm new here. How does this look? Rect and Oval were the only two I saw that really needed to be cleaned up but I went ahead and moved the button text anyway.

### DIFF
--- a/shoes-core/lib/shoes/dsl.rb
+++ b/shoes-core/lib/shoes/dsl.rb
@@ -346,27 +346,16 @@ class Shoes
     def oval(*opts, &blk)
       oval_style = pop_style(opts)
       oval_style = style_normalizer.normalize(oval_style)
-      case opts.length
-        when 3
-          left, top, width = opts
-          height = width
-        when 4
-          left, top, width, height = opts
-        when 0
-          left = oval_style[:left] || 0
-          top = oval_style[:top] || 0
-          width = oval_style[:diameter] || oval_style[:width] ||
-                  (oval_style[:radius] || 0) * 2
-          height = oval_style[:height] || width
-        else
-          message = <<EOS
+
+      left, top, width, height = opts
+
+      message = <<EOS
 Wrong number of arguments. Must be one of:
   - oval(left, top, diameter, [opts])
   - oval(left, top, width, height, [opts])
   - oval(styles)
 EOS
-          fail ArgumentError, message
-      end
+      fail ArgumentError, message unless [0, 3, 4].include? opts.size
       create Shoes::Oval, left, top, width, height, oval_style, blk
     end
 
--- a/shoes-core/lib/shoes/dsl.rb
+++ b/shoes-core/lib/shoes/dsl.rb
@@ -241,7 +241,7 @@ class Shoes
       create Shoes::Stack, opts, blk
     end
 
-    def button(text = 'Button', opts = {}, &blk)
+    def button(text = nil, opts = {}, &blk)
       create Shoes::Button, text, opts, blk
     end
 
--- a/shoes-core/lib/shoes/dsl.rb
+++ b/shoes-core/lib/shoes/dsl.rb
@@ -343,6 +343,7 @@ class Shoes
     #   @option styles [Integer] height (0) the height
     #   @option styles [Integer] top (0) the y-coordinate of the top-left corner
     #   @option styles [Boolean] center (false) is (left, top) the center of the oval
+    OVAL_ALLOWED_ARG_SIZES = [0, 3, 4]
     def oval(*opts, &blk)
       oval_style = pop_style(opts)
       oval_style = style_normalizer.normalize(oval_style)
@@ -355,7 +356,7 @@ Wrong number of arguments. Must be one of:
   - oval(left, top, width, height, [opts])
   - oval(styles)
 EOS
-      fail ArgumentError, message unless [0, 3, 4].include? opts.size
+      fail ArgumentError, message unless OVAL_ALLOWED_ARG_SIZES.include? opts.size
       create Shoes::Oval, left, top, width, height, oval_style, blk
     end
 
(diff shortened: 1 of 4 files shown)


KCErb:
Hi there, thanks for contributing! 

I'm just now looking at #968 (so @jasonrclark  and @PragTob will please correct me if I'm wrong) and I think we want the default arguments to end up in the element's `STYLES` hash as done [here](https://github.com/shoes/shoes4/blob/master/shoes-core/lib/shoes/arc.rb#L9).

It's not clear to me though, because it looks like the conversation about star (that I missed :blush:) resulted in doing [this](https://github.com/shoes/shoes4/blob/master/shoes-core/lib/shoes/star.rb#L15-L18) as you did.

So my question going forward is: when do we want to use `STYLES` and when do we prefer `||=`?

jasonrclark:
Good point about the `STYLES` convention for defining defaults. I had completely forgotten that when I did those changes on `star` which @graves was following along to.

I'm happy to use `STYLES` as the way to define the one-and-only-default-value for a given type if that makes sense. Will it give us the result we want, i.e. that after `initialize` the DSL element will have the default set for the given accessor?

Also, does it make sense for the dimensioning values like left, top, etc. to have default values in the `STYLES`, since those are a lot of the defaulting that was happening up in `dsl.rb`.

What do you think @KCErb? @graves is it clear what's being proposed there, and how we could test out if it works right on your branch?

Also, :heart: for getting involved! We love when new folks pitch in on Shoes, so please ask away if you have any questions.

shoes-core/lib/shoes/dsl.rb
```diff
@@ -346,27 +346,16 @@ def line(x1, y1, x2, y2, styles = {}, &blk)
...
-        when 4
-          left, top, width, height = opts
-        when 0
-          left = oval_style[:left] || 0
-          top = oval_style[:top] || 0
-          width = oval_style[:diameter] || oval_style[:width] ||
-                  (oval_style[:radius] || 0) * 2
-          height = oval_style[:height] || width
-        else
-          message = <<EOS
```
PragTob: love how this cuts down on stuff!

shoes-core/lib/shoes/dsl.rb
```diff
@@ -346,27 +346,16 @@ def line(x1, y1, x2, y2, styles = {}, &blk)
...
+
+      message = <<EOS
 Wrong number of arguments. Must be one of:
   - oval(left, top, diameter, [opts])
   - oval(left, top, width, height, [opts])
   - oval(styles)
 EOS
-          fail ArgumentError, message
-      end
+      fail ArgumentError, message unless [0, 3, 4].include? opts.size
```
PragTob: The Array should probably get to be a constant, more readable and efficient (as the array does not have to be recreated)

PragTob:
Thanks for getting involved! :heart: 

I must say, I also forgot about the `STYLES` defaults and wonder how that works out - we'll see. Probably a better place - although not all values can go there (i.e. dynamic values can't be there).

@KCErb would be cool if you could refer to some example code as guidance - even I would have to search for some default handling to copy :D

graves:
What values do you see living in the STYLES hash? These are overwritten in the styles_init method by matching keys in either of the hashes passed to it, meaning you'd have to dance around passing in nil as seen by my latest commit.

KCErb:
Hi all,

It seems to me that the question here is what to do with *argument* styles, so my example of how to default Arc's `wedge` style is not applicable. I think I misunderstood the problem when I first commented, sorry about that. 

After playing around a bit with this I'm realizing that merging a hash overwrites non-`nil` values with `nil` ones, which is not what I had expected.

I would lean towards something that does that, i.e. something that merges styles only if non-nil. If we used that over in `common/style.rb` then we could just pass all the args in as styles and not worry about them being nil.

The only objection I can think of is if someone actually wants to set a style to nil.

Ultimately @graves is right, we'll need to do some nil dancing since the `style` code relies on `merge` to decide which style wins (default vs arg).

Does that answer your question about dimensioning values? I see dimensions as just glorified styles.

shoes-core/lib/shoes/dsl.rb
graves: A constant or a variable? What should it be called?

jasonrclark: Pretty sure @PragTob meant a full constant outside the method. Maybe `ALLOWED_OVAL_ARGUMENT_COUNT` or something of that variety would be a good name?

PragTob:
Yeah my problem here is that people might actually want to pass `nil` values if I remember correctly there are some places in the shoes DSL where this is valid. Don't have an example right now... sorry.

If I'm right and that's the case we could say that people can only pass `false`... not sure, thoguh :(

@graves sorry for the confusion, we clearly missed something here earlier on. Thanks for pointing that out etc. :) :+1:

jasonrclark:
Hey all! So over in #1030 I'm looking to do some consolidation on the various DSL initialize methods, so I'd like to get this merged in.

While there's definitely some good stuff we could do around not having multiple ways of declaring defaults, I think what @graves has here is a nice step forward.

Any objections to me rolling with this? @KCErb or @PragTob, if you have specific changes you'd like to see subsequently, maybe we could get a separate issue in for them?

KCErb:
Yeah, it seems wise to go with what we have here for now. I'd really like to see a more uniform answer to this question in the future, but it shouldn't hold #1030 up.

PragTob:
Agreed. Sorry @graves for letting this lie around here for some time!

graves:
Hey no problem, happy to help in any way possible :smile:

jasonrclark:
Sweet, thanks all! I'll get this merged, hopefully later on today.

jasonrclark:
Once again, sorry for the delay, but this all worked out nicely. Thanks @graves :sparkles:

PR #1017, opened 2015-01-06
Smart spec_helper for root

jasonrclark:
This commit adds a smart spec_helper at the root of the project which
allows us to just do things like `rspec
shoes-swt/spec/shoes/swt/text_block_spec.rb` and have it work properly
(mod the `JRUBY_OPTS=-XstartOnFirstThread` for OS X users).

It works by just taking the first `shoes-.*` segment of any paths that are
passed, and requires their respective `spec/spec_helper`.

Note that there are potential side-effects if running an odd set of
tests across projects, but it does make the easy case easy.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,4 @@
 --color
 --format progress
 --order random
+--require spec/spec_helper.rb
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,0 +1,15 @@
+# This is a helper to let `rspec test-file` run directly from the root dir by
+# pulling in the right spec helpers from the subdirectories
+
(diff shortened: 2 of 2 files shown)


wasnotrice:
:heart: this idea. To clarify for anyone as forgetful as I am, to use this on OS X:

```
$ JRUBY_OPTS=-J-XstartOnFirstThread rspec shoes-swt/spec/shoes/swt/text_block_spec.rb
```

And of course you can export `JRUBY_OPTS` in your environment as well.

PragTob:
We should probably add that to the README/wiki/comment it :)

wasnotrice:
@PragTob good point. Added instructions to the README in 899b46a2be

PR #1018, opened 2015-01-06
Move some stray requires to be with their friends

wasnotrice:
No reason to have these requires all off by themselves like they're not
as good as all of the rest :D

Making these requires consistent also allows us to only have to modify
one file to load shoes-core for the atom backend.


jasonrclark:
Looks good to me :+1:

PragTob:
:+1:

PR #1022, opened 2015-01-07
Add method for adding details to inspect string

wasnotrice:
For consistency, readability, and compatibility with Opal's unmodifiable strings.

I was never really comfortable with the `insert` copy-pasted all over the code. I think this takes us at least a step in the right direction.

### DIFF
--- a/shoes-core/lib/shoes/app.rb
+++ b/shoes-core/lib/shoes/app.rb
@@ -86,11 +86,11 @@ class Shoes
     alias_method :fullscreen?, :fullscreen
 
     def to_s
-      inject_title super
+      add_detail_to_inspect super, " \"#{@__app__.app_title}\""
     end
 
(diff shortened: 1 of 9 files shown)


jasonrclark:
Totally makes sense and is cleaner anyway.

Noticed that a few object types seem to just be reporting their dimensions on `inspect` and not their own class names. Have you noticed this @wasnotrice?

wasnotrice:
No I haven't noticed that. But next time you notice it, throw up an issue, because it's easy to fix, and having specs for those strings makes it so much easier to refactor 🌱

jasonrclark:
Next time I'm looking at it will file an issue. Thanks for the cleanup! :rainbow:

PragTob:
Love the new approach! Thanks @wasnotrice ! =)

Issue #353, opened 2013-07-25
Curate samples


PragTob:
I believe it would be good to curate the samples before the rc1 is released. People will look at samples to get a grasp of shoes, therefore they should be as understandable as possible :-)

What do I mean by curate?
- **Make them clearer to understand** - rename some variables/methods or maybe even extract some methods or eventually add some comments.
- **make the sample names clearer** - nobody knows what `sample15` is, but with a better name one might know that. If you rename them make sure to also rename them in `samples/README`
- **Delete duplicated samples** - over time for one reason we amassed some samples that essentially do the same, find them and delete one of them! 

Thanks!

PragTob:
Updated to be more clear about what's happening and added the Newcomer Friendly tag

PR #1023, opened 2015-01-07
Delete practically duplicated samples

PragTob:
* nks_poem in favor of sample34
* simple-font in favor of sample45
* simple-button-animate in favor of simple-move (this is a bit arguable, as one used move and the other uses displace)


jasonrclark:
Yeah! Have definitely thought more than once that we should curate these better. Maybe even a newcomer friendly item to name things more reasonably (i.e. sampleX -> something meaningful) and clean out any other dups... What do you think @PragTob?

wasnotrice:
@PragTob I was noticing this issue too, but I didn't do anything about it! Thank you!

I think you are thinking of simple-displace instead of simple-move? But you're right, they're both similar to simple-move. They do illustrate the difference between move and displace though.

BTW we don't need to hold this up, but I think there are even more duplicates in there :)

PragTob:
Yeah there are a lot more I forgot some, will hunt them down the next time around :)

There are also two wildly different calculator implementations that look the same. One is more OO and imo easier to understand, the other has just half the loc which might be a metric for some people.

Also there is an issue for curating samples I'll edit it to make renaming them to have better file names as well. Issue is #353 - I added the newcomer friendly tag :)

Issue #926, opened 2014-10-25
Loop on resizing with percentage width stacks


jasonrclark:
Reported originally at http://librelist.com/browser//shoes/2014/10/20/shoes-4-loops-uninterruptibly-after-resize-when-stack-width-is-stated-as-a-percentage/

Given this program:

```ruby
#!    /usr/bin/env jruby
require 'shoes'

Shoes.app do
   flow do
     stack(width: '50%') do
       button "One"
     end
     stack(width: '50%') do
       button "Two"
     end
   end
end
```

Grab the bottom of the window and drag it up, resizing the window, as 
far as possible, then drag it down again. Notice the black area when you 
pull it down. Shoes loops uninterruptibly when this happens.

Running on jruby 1.7.16 (1.9.3p392) 2014-09-25 575b395 on OpenJDK 64-Bit Server VM 
1.7.0_65-b32 +jit [linux-amd64]
Built using rvm.
OS is Debian Jessie, which is a pre-release.

Some images of the problem, which we haven't entirely duplicated yet:

Just after starting the app: http://algoram.com/tmp/shoes/1.png
Bottom pulled up as far as it will go: http://algoram.com/tmp/shoes/2.png
Pull it down again, and oops, it’s looping and not responding: 
http://algoram.com/tmp/shoes/3.png

jasonrclark:
Pinged mailing list about this, as the recently merged #1024 might give us some diagnostics, but I'm unable to repro it on OS X.

PragTob:
We don't have feedback --> closing until we get some

PR #1024, opened 2015-01-08
Fix for hang during content alignment with text

### LINKED ISSUE #982: Hanging layout/redraw loop in class-book-adjusted
If I run `samples/class-book-adjusted.rb` and drag it down to the narrowest possible width, the app hangs and gives me a beachball (clear sign that it's in some type of infinite loop).

Was happening as far back as `pre2`, so I'm not going to block `pre3` for it, but it's on the short list after that.

Not sure whether it matters or not yet, but I'm running on OS X.

jasonrclark:
This fixes #982 in two ways:

* Calculations which were failing when text blocks got squeezed too small have been individually tightened up to not raise. These are the changes you see in `swt/text_block/fitter.rb` and `swt/text_block.rb` which are being more careful around 0 height/width and empty segments.
* Also, because the behavior is so strange and unhelpful, we also now trap exceptions in the slot positioning code rather **loudly** but carry on so we don't bork SWT.

:boom: => :hushed:

### DIFF
--- a/shoes-core/lib/shoes/slot.rb
+++ b/shoes-core/lib/shoes/slot.rb
@@ -150,6 +150,13 @@ class Shoes
       else
         current_position
       end
+    rescue => e
+      puts "SWALLOWED POSITIONING EXCEPTION ON #{element} - go take care of it: " + e.to_s
+      puts e.backtrace.join("\n\t")
+      puts 'Unfortunately we have to swallow it or risk SWT hanging.'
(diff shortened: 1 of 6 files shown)


jasonrclark:
The general rescue code *might* help out with #926 (although the specific text block stuff doesn't apply since there aren't any text blocks in the sample app in question). Once this is landed on master, may follow up with the reporter and see if that helped it out at all.

PragTob:
Looks great, thanks!

Issue #911, opened 2014-10-05
Download of non file URLs does not work


PragTob:
With our `simple-download.rb` I tried downloading a couple of URLs that are not directly a file. 

So for instance, downloading a file is fine e.g. this URL points to a small version of my profile pic: https://avatars3.githubusercontent.com/u/606517?v=2&s=40

Works fine.

However http://github.com - does not seem to download anything for me :-(

Even worse, https://github.com causes a crash :-(

```
tobi@big-one:~/github/shoes4$ bin/ruby-shoes samples/simple-downloader.rb 
java/lang/reflect/Method.java:601:in `invoke': org.eclipse.swt.SWTException: Failed to execute runnable (org.jruby.exceptions.RaiseException: (TypeError) NilClass can't be coerced into Fixnum)
	from /home/tobi/.rvm/gems/jruby-1.7.16@shoes/gems/swt-4.4/lib/swt/minimal.rb:109:in `run_event_loop'
	from /home/tobi/.rvm/gems/jruby-1.7.16@shoes/gems/swt-4.4/lib/swt/minimal.rb:92:in `event_loop'
	from /home/tobi/github/shoes4/lib/shoes/swt/app.rb:46:in `open'
	from /home/tobi/github/shoes4/lib/shoes/internal_app.rb:103:in `open_gui'
	from /home/tobi/github/shoes4/lib/shoes/app.rb:49:in `initialize'
	from /home/tobi/github/shoes4/lib/shoes/app.rb:19:in `app'
	from /home/tobi/github/shoes4/samples/simple-downloader.rb:1:in `(root)'
	from org/jruby/RubyKernel.java:1081:in `load'
	from /home/tobi/github/shoes4/lib/shoes/ui/cli.rb:1:in `(root)'
	from /home/tobi/github/shoes4/lib/shoes/ui/cli.rb:79:in `execute_app'
	from bin/ruby-shoes:6:in `(root)'
	from bin/bin/ruby-shoes:6:in `(root)'
Caused by:
org/jruby/RubyFixnum.java:559:in `/': org.jruby.exceptions.RaiseException: (TypeError) NilClass can't be coerced into Fixnum
	from /home/tobi/github/shoes4/lib/shoes/download.rb:47:in `percent'
	from /home/tobi/github/shoes4/samples/simple-downloader.rb:18:in `(root)'
	from org/jruby/RubyProc.java:271:in `call'
[...]

PR #1025, opened 2015-01-08
Download fixes

jasonrclark:
So there were two separate things reported over in #911 that I address here (at least partially).

https://github.com was a problem because the response we receive from them didn't have a `Content-Length` header, so we didn't set the content length instance variable properly and crashed. I fixed this simply by making `percent` more resilient to those values being `nil`--other uses of them generally looked ok to me, although I wasn't too exhaustive about reading up on them.

http://github.com returns a 301 redirect to https://github.com... but unfortunately `open-uri` disallows redirecting from `http` elsewhere for security reasons (see https://github.com/ruby/ruby/blob/trunk/lib/open-uri.rb#L243-L252 for the gory details). So while I can't fix that without changing our use of `open-uri`, what I did was made it so the error actually propagates up. It was happening on a background thread and silently dying, so now when the download thread gets a problem it shuttles it off to the parent thread via the standard `eval_block` used in all the download code to get work done on the main UI thread.

Certainly up for discussion of whether we should use a different HTTP library, but these changes seem valuable on their own regardless where we head with that.

### DIFF
--- a/shoes-core/lib/shoes/download.rb
+++ b/shoes-core/lib/shoes/download.rb
@@ -69,17 +69,27 @@ class Shoes
     def start_download
       require 'open-uri'
       @thread = Thread.new do
-        uri_opts = {}
-        uri_opts[:content_length_proc] = content_length_proc
-        uri_opts[:progress_proc] = progress_proc if @progress_blk
+        begin
+          uri_opts = {}
+          uri_opts[:content_length_proc] = content_length_proc
+          uri_opts[:progress_proc] = progress_proc if @progress_blk
 
-        open @url, uri_opts do |download_data|
-          @response.body = download_data.read
-          @response.status = download_data.status
-          @response.headers = download_data.meta
-          save_to_file(@opts[:save]) if @opts[:save]
-          finish_download download_data
+          open @url, uri_opts do |download_data|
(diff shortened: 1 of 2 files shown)


PragTob:
Looks good :+1: 

Very open to a http library, it means another dependency which is a bit "meh" on the other hand I don't want us to waste time time on things that are not our competency. Something miniml would be cool.

Otherwise we could role our own minimal thing, maybe, to share with furoshiki as we already solved the redirecting problem over there... or maybe that's in `shoes-package` now, could not find it quickly and have to get out of the train now :)

jasonrclark:
I think the biggest problem with a replacement is finding one with a progress proc... don't know of one off the top of my head, but haven't looked hard yet.

I'll probably log a separate issue when I get this merged later today about the following redirects so we can get #911 closed out as it stands. Thanks for the double-check @PragTob.

wasnotrice:
@jasonrclark great work here. Love the simplicity! And totally agree about the separate http lib, pros and cons.

jasonrclark:
Thanks for the merge and the issue @wasnotrice!

wasnotrice:
Trying to lighten your secretarial load so you have more energy for awesome coding :star2:

PR #1027, opened 2015-01-11
Support click on shape

jasonrclark:
Just what it says on the tin :wink: 

The first problem was that the backend was lacking the support for click, which was easily remedied by adding the `Shoes::Swt::Common::Clickable` module.

That unearthed that the standard `Dimensions` provided `in_bounds?` check wasn't quite right, because `Shape` doesn't participate in normal positioning (being an art element), and has its own bounds instance variables kept up to date throughout drawing. Given that, a custom `in_bounds?` didn't seem, ahem, out of bounds to me so I went with it.

### DIFF
--- a/shoes-core/lib/shoes/shape.rb
+++ b/shoes-core/lib/shoes/shape.rb
@@ -100,6 +100,14 @@ class Shoes
       self
     end
 
+    # Determines if a given point is in the boundary of the shape. Given the
+    # many possibilities of what a shape could contain, this just checks the
+    # outer bounding box of the shape, nothing more sophisticated.
+    def in_bounds?(x, y)
+      (@left_bound..@right_bound).include?(x) &&
+      (@top_bound..@bottom_bound).include?(y)
+    end
+
     private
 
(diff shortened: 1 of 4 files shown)


PragTob:
hey @jasonrclark - art elements should participate in normal positioning. At least all others do (rect etc). And shape should rely on all the Dimension logic as of #737 - so this got me kind of wondering :)

Looks good otherwise :)

jasonrclark:
Fair point @PragTob--I overstated things saying it isn't normally positioned. However, even though shape runs through the normal positioning, it doesn't seem to end up with values assigned to it that result in the built-in `Dimensions#in_bounds?` working.

I can dig into that further if you'd like before merging here, or we could take the mystery for the moment (given the already existing around `Shoes::Shape#update_bounds` and move on. What do you think?

PragTob:
Seems good with me, just wanted to check that out to make sure no misconceptions exist on either side :)

The values it uses are not standard Dimensions values, however absolue_left/right/top/bottom should work... well should this is few code enough that this is a good solution imo - thanks! :)

Issue #808, opened 2014-07-25
All elements should support the hidden style


PragTob:
As noted by @KCErb in #806 all elements should support a `hide` style where one can style them to hide them.

For this it might be nice to add a method to the visibility module (https://github.com/shoes/shoes4/blob/master/lib/shoes/common/visibility.rb) where one can just say something like `hide(true)` or `hide(false)` but that name right now collides with the existing method.

jasonrclark:
Closed via #1028

PR #1028, opened 2015-01-11
Support the hidden style on all elements

jasonrclark:
Per #808 and compatibility with Shoes 3.

To accomplish this, instead of using a separate instance variable in the `Shoes::Common::Visibility` module (`@hidden`), we now back onto `style[:hidden]` as the stateful source of all truth about whether to show ourselves or not.

This required a couple of additional changes:

* ~~Several elements needed to have a call to `update_visibility` when they were done. These were primarily anything without a painter (since painters were taking visibility into account properly), because the backends didn't consult the visibility flags directly. I actually like this well enough, since another backend will get the proper calls, and we don't bake things. Note that we can't do this in `style_init` because the backend objects don't actually exist yet at that point in time. We'd need another "done with initializing" hook to try to do it generically.~~ Revised in light of #1030 so only one the `Common::Initialization#initialize` method needed `update_visibility` added. Hurray!
* For testing purposes `Shoes::Common::Visibility#update_visibility` does a `respond_to?` check since now it gets hit against the mocks, which don't all individually have that method. This seemed cleaner than scattering more of this business across the mocks to me at the moment.
* `Shoes::FakeElement` got a little bit of style treatment, since it uses `Visibility` but not `Style` (too much app-level stuff in that module to use from this fake object).

### DIFF
--- a/shoes-core/lib/shoes/common/visibility.rb
+++ b/shoes-core/lib/shoes/common/visibility.rb
@@ -3,12 +3,12 @@ class Shoes
     module Visibility
       # Hides the element, so that it can't be seen. See also #show and #toggle.
       def hide
-        @hidden = true
+        style[:hidden] = true
         update_visibility
       end
 
       def hidden?
-        @hidden
+        style[:hidden]
       end
 
       alias_method :hidden, :hidden?
@@ -19,21 +19,21 @@ class Shoes
 
       # Reveals the element, if it is hidden. See also #hide and #toggle.
       def show
-        @hidden = false
+        style[:hidden] = false
         update_visibility
       end
 
       # Hides an element if it is shown. Or shows the element, if it is hidden.
       # See also #hide and #show.
       def toggle
-        @hidden = !@hidden
+        style[:hidden] = !style[:hidden]
         update_visibility
       end
 
       private
 
       def update_visibility
-        gui.update_visibility
+        gui.update_visibility if gui.respond_to?(:update_visibility)
         self
       end
     end
--- a/shoes-core/lib/shoes/common/visibility.rb
+++ b/shoes-core/lib/shoes/common/visibility.rb
@@ -33,7 +33,7 @@ class Shoes
       private
 
       def update_visibility
-        gui.update_visibility if gui.respond_to?(:update_visibility)
+        gui.update_visibility
         self
       end
     end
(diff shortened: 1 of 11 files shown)


shoes-core/lib/shoes/common/visibility.rb
```diff
@@ -19,21 +19,21 @@ def visible?
...
-        @hidden = !@hidden
+        style[:hidden] = !style[:hidden]
         update_visibility
       end
 
       private
 
       def update_visibility
-        gui.update_visibility
+        gui.update_visibility if gui.respond_to?(:update_visibility)
```
PragTob: I'd be against having this check in here just for the mock backend. There should be a `CommonMethods` somewhere in the mock world, I'd just jam the `update_visibility` method in there.

PragTob:
Looks good to me, thanks @jasonrclark :)

I'm not the biggest of repeating the `update_visibility` call around the initialize as it kind of duplicates the concept that is entailed there... it's not bad, though.

it's also a bit hard to think of a solution. I have two things in mind:

* Have a common shoes super class, that takes care of this - like `Shoes::GuiObject` or something as there's a lot that we do in basically every class now (style_init, update_visibility, creating the backend instance) 
* Use `after_do` to update the visibility - reduces duplication but also decouples it further so that it might be harder to track down (it's kind of the `after_initialize` hook you were mentioning) :)

wasnotrice:
another possibility that I have been entertaining is to do some of this initialization in `Shoes::Dsl#create`. For all elements that are created "for real", I think this is (or is meant to be) the creation path. Benefit: we have a place to hang method calls that need to happen after unit but before use. In other words, we may be trying to do too much in initializers.

PragTob:
Interesting thought @wasnotrice - so looks like we got 3 options now, anyone care to add some more? :)

jasonrclark:
Posted #1030 with some proposed code for resolving the duplication. If we like it, I'd be inclined to hold this until that's finished and revise this in light of the new paradigm.

jasonrclark:
Rebased this on master for #1030 and revised to make `update_visibility` unconditional per @PragTob's excellent suggestion. There were fewer things needed in the mocks than I expected, *and* it uncovered where `Shoes::Swt::Shape` didn't have `update_visibility` either. :metal: 

Assuming tests pass on this and everyone's happy, I think it's good to go.

PragTob:
Looks good to me :+1:

Issue #752, opened 2014-06-09
Everyone gets app, parent, gui


KCErb:
Over in #736 we discussed giving all of the elements `@app` `@parent` and `@gui`. Probably would put that in `Shoes::CommonMethods` or establish a base class.

Thoughts? Which is better base class or stick it in the module?

Issue #741, opened 2014-05-30
Unify code for generating backend objects


wasnotrice:
Right now, we have two different methods that create backend objects. Some objects call `Shoes.backend_for` (which then calls `Shoes::Configuration#backend_with_app_for`), and some call `Shoes.configuration.backend_for` (Shoes::Configuration#backend_for). See [lib/shoes/configuration.rb](https://github.com/shoes/shoes4/blob/master/lib/shoes/configuration.rb)

This was a transition that didn't get completed, but we ought to have a single code path here. The main question here is whether a DSL object should reach into its app/parent and pass a gui object to the backend factory, or whether the backend factory should reach into the DSL object for the gui object.

Any preferences?

PR #1030, opened 2015-01-12
Proposal for consolidating DSL initialization

jasonrclark:
Over in #1028 there's discussion of reducing duplication in `initialize` for the DSL classes. This came up from my spiking a call to `update_visibility` into several elements. I wasn't hugely satisfied with that myself, but it was late when I pushed that PR :zzz: :grin: 

Here's a proposal for resolving that duplication. It's similar to @PragTob's option 1) in #1028 (call it a 3.5-th option :stuck_out_tongue_winking_eye:)

**Pros**
* No required base-class (:arrow_down: with inheritance!)
* Gets consistency across DSL types, resolving #752 along the way
* Can be applied in stages, then pulled up to `Shoes::Common::UIElement` once all elements are on board
* Allows cross-cutting additions to DSL initialization at any point of the process, not just at the beginning/end like a `create` override would

**Cons**
* Not as clear (especially to beginners) where a DSL instance's creation actually happens
* Might be a surprise to some that `initialize` can be pulled in from a module like this

Honestly not attached to the approach, so fire away with your thoughts and other proposals!


KCErb:
Hey I think that's great! :+1:

The cons both seem to be related to how a beginner might be tripped up by this sort of thing, but to me (a beginner) it seems we do this kind of stuff all over the code base, i.e. move important methods out of the class into a common container.

So while the `Initialization` module would be a surprise to me, it would also be a good intro to a better way of thinking about how Shoes is structured.

I might point out that along with this hitting #752, this seems to pick up #741.

wasnotrice:
I like this approach. As @jasonrclark noted, it is more-or-less the same approach that @PragTob proposed. While it doesn't introduce a common superclass, it is an inheritance-based strategy, even if it doesn't *look* that way.

Still, given the architecture as it exists, this is probably the best solution.As I explore alternate backends (i.e., javascript), I am becoming more conscious of the fact that we have made some assumptions in our core implementation that may make it more difficult to implement alternate backends.

Here, for instance, you can see that the most basic thing you could possibly do (instantiate an object) is intimately tied to a specific notion of belonging to an app, belonging to a parent, and instantiating a backend object that represents this object (and which, in the SWT backend, involves a number of additional SWT objects). Which is just to say that we are doing quite a lot during initialization, and that we require a lot of other objects (i.e. dependencies) to be in place simply to instantiate a single object. This works for SWT, but may not necessarily work for other backends (no specific cases yet). I think we all feel this web of dependences in the specs, too. I'm wondering if ultimately we might want to separate the notion of initializing an object from the notion adding it to an app, or to a parent. Maybe even using another object to "prepare/configure" the object to be added? I'm not sure exactly how this would look, and I don't think we are ready to do it yet.

Anyway, I'm not sure that doing all of this stuff during initialization will create problems down the road, but it sure seems like it might. In any case, this pull request helps in the sense that it consolidates all of that logic into one place. If at some point we decide that a different approach would be more flexible, it will be easier to change after this pull request than it would have been before. 

:+1:

shoes-core/lib/shoes/common/initialization.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      def before_initialize(styles)
+      end
+
+      def after_initialize
+      end
+
+      def dimensions_class
+        AbsoluteDimensions
+      end
```
PragTob: This is not really needed, is it? Looks like an earlier attempt at another way to `create_dimensions` :)

shoes-core/lib/shoes/rect.rb
```diff
@@ -8,14 +9,9 @@ class Rect
     style_with :angle, :art_styles, :curve, :common_styles, :dimensions
     STYLES = { angle: 0 }
 
-    def initialize(app, parent, left, top, width, height, styles = {}, blk = nil)
-      @app = app
-      @parent = parent
-      style_init styles
+    def create_dimensions(args)
+      left, top, width, height = args
```
PragTob: not super sure, but couldn't we destructure this in the parameter list already?

shoes-core/lib/shoes/common/initialization.rb
jasonrclark: Yep, good eye!

PragTob:
Looks good to me - I think it helps with overall initialization and will probably be continuously refined once it's used for more classes :)

About the assumptions - I'm sure we will run into problems sooner or later somewhere somehow :) Great to have someone on board who has experience with this :) So far I believe that having an app and a parent is part of the shoes DSL and the instantiation of the backend object is how we imagine the backend working so far... it most definitely feels clumsy but I think it's ok, and it's great to have it all in one place now :)

:+1:

shoes-core/lib/shoes/rect.rb
jasonrclark: As long as args has the right number of arguments for any given DSL class, I think you might be right. Will experiment with it and report back.

jasonrclark:
So sounds like everyone's on board with the approach, so I'll address @PragTob's immediate comments and start pushing this through other DSL classes to see what comes up in the course of that. Thanks all for the swift feedback!

jasonrclark:
First set of changes through, I tackled the art elements (since I'd started with the `Rect`). This required some minor revisions. The big revelation being that `shape` takes a block but doesn't use it as a `click` (duh) so that needed a bit more generalizing.

I expect some of the other elements will probably be even less consistent, but it's time to take a break for the evening :zzz:

shoes-core/lib/shoes/common/initialization.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        before_initialize(styles)
+
+        @app = app
+        @parent = parent
+
+        style_init(styles)
+        create_dimensions(*args)
+
+        @parent.add_child self
```
PragTob: Did I mention how I love to have this here and not add it EVERY DAMN TIME?

Thank you :heart:

shoes-core/lib/shoes/common/initialization.rb
```diff
@@ -0,0 +1,40 @@
...
+        @gui = Shoes.backend_for self
+
+        handle_block(blk)
+
+        after_initialize
+      end
+
+      def create_dimensions(args)
+        raise NotImplementedError.new
```
PragTob: As per rubocop we should probably `fail` here :)

PragTob:
Love what I'm seeing :tada: :smile: Keep :metal: :+1:

jasonrclark:
Pushed through the rest of the changes on this. :fireworks: 

Everything that included `UIElement` got the `Initialization` treatment, and at the end I pulled that up along with attr_readers in common ala #752.

Going to run through the samples on it now, but pretty happy with how it came together. There was less wild variety than I was concerned about, and lots of duplication cut.

jasonrclark:
Hurray, samples all passed! If anyone has time to give this a glance, I'd appreciate it since it does cover a lot of ground, but it's very similar to what you all saw on it earlier.

wasnotrice:
@jasonrclark I really like how this turned out!

PragTob:
Will take a look and most likely merge once I'm through - thanks a lot @jasonrclark ! :)

shoes-core/lib/shoes/common/initialization.rb
```diff
@@ -0,0 +1,57 @@
...
+      # necessary methods are there.
+      #
+      # Override if DSL element uses that block for something else (i.e. slot)
+      def handle_block(blk)
+        return unless blk
+        register_click blk
+      end
+
+      # Final method called in initialize. Intended for any final setup after
+      # the rest of the object has been set up fully.
```
PragTob: Thanks for the docs, very helpful! :)

PragTob:
Looks great, we should have done this earlier :)

PR #1029, opened 2015-01-12
Corrected two rubocop offenses

davidenglishmusic:

PR #1031, opened 2015-01-12
Fix gem name check on uninstall

jasonrclark:
Missed in the Great Gem Shuffle (Part 2) of 2014

Result was that uninstall `shoes-core` potentially left around a stale `shoes` (or `shoes.bat` for the Windows-inclined) and `shoes-backend` executable.

Noticed while taking a peek at #1029. Double goodness from one PR!


PragTob:
Hey,

thanks for the fixup - as this happened a second time, can you think of  trivial way to test this? I can't right now, just don't want it to slip once more in the future

jasonrclark:
Hmm, can't think of anything trivial, but could definitely wrap something around it. Will mull it over and see what I can come up with.

PragTob:
Don't spend too much time on it - I hope the gem structure is not soo likely to change again and the fix is easy. So the effort for a spec might be too high to be worth it - although automatic detection would be nice.

Issue #593, opened 2014-02-04
Assign elements their position right after adding them and update as needed


PragTob:
There are a couple of issues related to the position of an element not being set immediately after it is added and continuously running contents_alignment for all elements to ensure that everything is set. Some of them are:

* #491
* #590 
* trying to base the position of subsequent elements on the previous ones
* all of the nil guards that make the Dimensions class harder to read and maintain #549 

A proposed algorithm could just add elements as they are created. E.g. the process for contents alignment is linear either way - start with the first element, position it, then the next etc. That's not a big problem - hopefully.

What's a bigger problem is appending/prepending elements. When prepending elements all subsequent elements have to be positioned. Same goes for appending - but for appending those are in the next slot (if there is a next slot). 
The easiest solution for the start would be, that whenever prepend/append is called we run a full contents alignment. However it should still be faster and called less. Later on the approach could be refined to only reposition the elements that need repositioning. This could go very far - e.g. has the prepend/append operation changed the `absolute_bottom` of a slot? If yes then reposition elements in the next slot. If not, stop.

Or well - reposition elements as long as the position of an element changes. As the position of the next element is based on the position of the previous element we should be good as long as it is positioned the same.

Bu yeah there are probably lots of subtle issues with this, I don't expect it to be easy going but it looks like a good thing to do for the overall architecture :)

PR #1032, opened 2015-01-12
Move guard against empty text segments further up in the chain

### LINKED ISSUE #636: FittedTextLayoutCollection - layouts.first can be nil
I had to add the following guard statement:

```ruby
        return [] unless @layouts.first # TODO WTF
```

to the code for simple-downloader.rb to work. See [this comment](https://github.com/shoes/shoes4/pull/612#issuecomment-39989776).

It seems to work for @jasonrclark and @KCErb though. So maybe it's linux only...

I'd love for more people to try it out though.

Steps to reproduce is basically:

1. start samples/simple-downloader.rb
2. Enter like any URL
3. hit go
4. BOOOOOM

Help/Investigation appreciated.

PragTob:
It has been shown that adding some elements triggers a redraw
while positioning has not run yet and causes failures later on.
This fixes #636 as we have found the root cause of the problem,
which can be resolved by solving #593 :)

Thanks @jasonrclark ! :)

### DIFF
--- a/shoes-swt/lib/shoes/swt/text_block/painter.rb
+++ b/shoes-swt/lib/shoes/swt/text_block/painter.rb
@@ -15,6 +15,10 @@ class Shoes
         def paintControl(paint_event)
           reset_graphics_context(paint_event.gc)
           return if @dsl.hidden?
+          # See #636 for discussion, contents_alignment may not run or if the
+          # space is very narrow we might squish things down to be very narrow.
+          # If paint is triggered then, code later on will crash.
+          return if @dsl.gui.segments.empty?
 
           draw_layouts(paint_event.gc)
         end
--- a/shoes-swt/lib/shoes/swt/text_block/text_segment_collection.rb
+++ b/shoes-swt/lib/shoes/swt/text_block/text_segment_collection.rb
@@ -82,7 +82,6 @@ class Shoes
         # be in either, or both, of the segments. This method figures out which
         # segments apply, and what the relative ranges within each segment to use.
         def segment_ranges(text_range)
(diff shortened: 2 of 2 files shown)


jasonrclark:
Only comment I have is on the comment... it's not just when `contents_alignment` hasn't run, but in some very small window scenarios, we can end up with empty segments. Is that worth calling out too while we're being specific about it?

PragTob:
Totally. Will update and squash (not worth an extra commit)

PragTob:
Or wait a second, does that really guard against that? I don't know what you mean exactly, but can we ever have a `TextBlock` that has no segmens at all? That would mean like no comments, doesn't it?

What I can imagine is that the first segment itself is an empty segment, but we have segments - so this guard does not guard against them. But we guard against them elsewhere :)

Is that correct? Which one is it? TextBlock isn't really my area of most experience :)

jasonrclark:
So the segments represent how we broke the text up across multiple SWT layout objects for actual display. When we squish things down too small to even show, the fitter will give up and just return an empty set of segments to signify "Nothing I can show here." (Note that the `TextBlock` will still have text, just the "fitted segments" will be empty)

The relevant line is https://github.com/shoes/shoes4/blob/master/shoes-swt/lib/shoes/swt/text_block/fitter.rb#L98-L100 which was introduced during #1024.

PragTob:
I see, thanks for the explanation. Always learning, always learning :)

PragTob:
Updated, hope the comment correctly expresses what actually happens :)

jasonrclark:
Yep, that gets it. Thanks!

PR #1037, opened 2015-01-25
Factor margins into TextBlock#desired_width

### LINKED ISSUE #750: Text blocks not respecting margin_right
Margins in all other directions are fine, but it doesn't pay attention to the right margin if applied. This also holds whether we set `margin_right` directly, or if it's set as part of the overriding `margin` style instead.

```
Shoes.app  height: 250 do
  para 'Hello Shoes 4 ' * 20, fill: lightblue, margin: 100
end
```

screen shot 2014-06-08 at 11 15 38 pm

jasonrclark:
This was preventing the right margins from being properly respected.

Fixes #750

### DIFF
--- a/shoes-core/lib/shoes/text_block_dimensions.rb
+++ b/shoes-core/lib/shoes/text_block_dimensions.rb
@@ -31,7 +31,7 @@ class Shoes
 
     # This is the width the text block initially wants to try and fit into.
     def desired_width(containing = containing_width)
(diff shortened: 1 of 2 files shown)


PragTob:
Oh my I haven't looked at this yet, I'm so bad :( Right on it!

PragTob:
Really good, thanks! =)

Issue #777, opened 2014-07-03
One-pixel gaps between flowing TextBlock segments


jasonrclark:
When a `TextBlock` flows into separate layouts we align the text correctly, but because of how we're using the underlying SWT elements there's a gap between them when a background color is present (see sample app following).

There's also a vertical gap between text elements which probably shouldn't be there.

I tried one naive solution by altering the second SWT `TextLayout`'s spacing, but it didn't work right for all the spacing possibilities in this test app (requires resizing of the app to see all the possible ways the text blocks can split across lines):

```
Shoes.app width: 250, height: 500 do
  COLORS = [red, green, blue, yellow, orange, purple]
  3.times do
    flow do
      COLORS.each do |color|
        para "Boo Hoo Hoo Hoo", fill: color, size: 24
      end
    end
  end
end
```

![screen shot 2014-05-24 at 8 12 02 pm](https://cloud.githubusercontent.com/assets/130504/3076184/6053c8aa-e3ba-11e3-8522-baad3b0e7605.png)

jasonrclark:
So present behavior on this is a bit different. The one-pixel gap remains between the text blocks, which deserves to get fixed with this item, but weirder offset seen above doesn't happen.

![image](https://cloud.githubusercontent.com/assets/130504/5952886/3d138440-a735-11e4-8077-e88f33755a0a.png)

However, not all is rosy. There's been changes to exactly when we choose to bump to the next line, and apparently some of the math is off for positioning the next text.

![image](https://cloud.githubusercontent.com/assets/130504/5952913/8f4dff4c-a735-11e4-9ed6-bd5da5610b12.png)

PragTob:
Thanks for the update @jasonrclark ! :)

jasonrclark:
Posted #1038 with fixes for some of the off-by-ones. Indicated there that I see other problems from this test:

**Text clipped on right** I suspect this is something wrong with width calculations, maybe given certain font sizes, because doesn't happen when text is set to default size.
![image](https://cloud.githubusercontent.com/assets/130504/5991239/a91a6ea8-a995-11e4-9e61-fb69c082f91b.png)


**Background not filling to edge** Would look a lot nicer if this didn't happen:
![image](https://cloud.githubusercontent.com/assets/130504/5991241/d336e9e6-a995-11e4-8855-01c7388a769a.png)


Suppose I could also file separate issues for them, but either way didn't want to lose track of those.

PragTob:
[...]

PR #1038, opened 2015-02-01
Fixes for one-pixel off problems with text blocks

jasonrclark:
This addresses a variety of different off-by-one issues we had with the
exact layout of text blocks. This shows most prominently when text
blocks are lined up with backgrounds. See #777 for an example.

Note that I'm not in love with how these changes went, but from testing
they are right in appearance, even if I can't precisely explain all the
fiddly bits off the top of my head. It deals with:

* One-pixel between text blocks horizontally
* One pixel between text blocks vertically
* An additional horizontal pixel off when flowing to second section of a text block
* Text blocks of same height bumping down two lines instead of lining up perfectly

There are still some other issues I've uncovered based on #777, so I won't
be closing it on this, but these seemed like a valid step to take.

### DIFF
--- a/shoes-swt/lib/shoes/swt/text_block.rb
+++ b/shoes-swt/lib/shoes/swt/text_block.rb
@@ -44,10 +44,11 @@ class Shoes
       end
 
       def adjust_current_position(current_position)
-        current_position.y = @dsl.absolute_bottom
+        current_position.y = @dsl.absolute_bottom + 1
 
         last_segment = segments.last
         if last_segment && !@bumped_to_next_line
+          current_position.x -= 1
           current_position.y -= last_segment.last_line_height
         end
       end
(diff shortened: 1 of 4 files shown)


shoes-swt/lib/shoes/swt/text_block.rb
```diff
@@ -44,10 +44,11 @@ def contents_alignment(current_position)
       end
 
       def adjust_current_position(current_position)
-        current_position.y = @dsl.absolute_bottom
+        current_position.y = @dsl.absolute_bottom + 1
```
PragTob: In `Slot` we have a `NEXT_ELEMENT_OFFSET` constant and I think another one in Dimension, maybe we should reuse one of them or just have one here as well - makes it a bit more readable :)

shoes-swt/lib/shoes/swt/text_block.rb
```diff
@@ -44,10 +44,11 @@ def contents_alignment(current_position)
...
 
         last_segment = segments.last
         if last_segment && !@bumped_to_next_line
+          current_position.x -= 1
```
PragTob: This looks like it might be hiding another problem, I'm not sure - might be worth to put a comment or something there :)

PragTob:
Looks good, although I also don't know why some of those +-1s are there :)

Thanks @jasonrclark !

jasonrclark:
At long last got back to actually touch up those bits you mentioned @PragTob. Good advice.

Will get this merged as soon as the tests are green on Travis.

PR #1043, opened 2015-02-05
Update app.yaml

bliaxiong:
Add example of adding your ruby file. This will be helpful.


PragTob:
This is helpful, thanks  never mind the failing appveyor build I have yet tos et it up correctly which turned out more complicated than I thought.

Thanks!

Issue #1047, opened 2015-02-06
Text background not filling to edge when wrapping


jasonrclark:
Sample app (drawn from #777)

```
Shoes.app width: 250, height: 500 do
  COLORS = [red, green, blue, yellow, orange, purple]
  3.times do
    flow do
      COLORS.each do |color|
        para "Boo Hoo Hoo Hoo", fill: color, size: 24
      end
    end
  end
end
```
![image](https://cloud.githubusercontent.com/assets/130504/5991241/d336e9e6-a995-11e4-8855-01c7388a769a.png)

Would look a lot nicer if the color on the first text layout extended to the furthest edge.

PragTob:
Huh, it probably would look nicer but I guess it migt cause problems - what we would need for this to work is to make the first text block (when we break them up) extend all the way to the edge (increase width). Hm Maybe it's not that difficult after all :)

PR #1048, opened 2015-02-07
Text background and wrapping fix

### LINKED ISSUE #1046: Text clipped on right instead of wrapping
Sample app (drawn from #777)

```
Shoes.app width: 250, height: 500 do
  COLORS = [red, green, blue, yellow, orange, purple]
  3.times do
    flow do
      COLORS.each do |color|
        para "Boo Hoo Hoo Hoo", fill: color, size: 24
      end
    end
  end
end
```

image

Note how the text is clipping on the right. I suspect this is something wrong with width calculations, maybe given certain font sizes, because doesn't happen when text is set to default size.

jasonrclark:
This fixes #1046 and #1047 (turns out fix for #1047 addressed the other :tada:)

I've looked back at the history, and I don't see why we were bothering to try and restrict the width on the SWT `TextLayout` object if the text already fit inside the width the fitter passed. Taking that away addresses the resizing problem in #1046.

The story on #1047 is sadder and more annoying. Turns out that SWT's `TextLayout` only puts the background around the characters that it draws, even when the layout is explicitly set to a greater width.

When we're filling in the first of multiple text segments, we actually want the full space filled even though the text probably doesn't eat it all up. Given that, we have to perform the fill ourselves more directly.

Don't have 1000 more words, so here's a couple pictures:

image

image

### DIFF
--- a/shoes-swt/lib/shoes/swt/text_block/text_segment.rb
+++ b/shoes-swt/lib/shoes/swt/text_block/text_segment.rb
@@ -14,7 +14,8 @@ class Shoes
       class TextSegment
         DEFAULT_SPACING = 4
 
-        attr_reader :layout, :element_left, :element_top
+        attr_reader   :layout, :element_left, :element_top
+        attr_accessor :fill_background
 
         extend Forwardable
         def_delegators :@layout, :text, :text=, :bounds, :width, :spacing,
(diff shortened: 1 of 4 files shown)


shoes-swt/lib/shoes/swt/text_block/text_segment.rb
```diff
@@ -95,6 +96,15 @@ def last_line_width
         end
 
         def draw(graphics_context)
+          # Why not use TextLayout's background? Unfortunately it doesn't draw
+          # all the way to the edges--just around the literal text. This leaves
+          # things jagged when we flow, so do it ourselves.
```
PragTob: thanks for the comment!

PragTob:
Looks great, thanks heaps!

Issue #1049, opened 2015-02-08
Changelog output problems on OS X


KCErb:
Hi all, I'm noticing that `rake changelog` doesn't work as expected for me.

Here's what `inspect` on `Changelog.new.generate` returns

```
"Improvements (5)\n----------------\n\n* Support the hidden style on all elements\r [63afeca]\n* Merge branch 'master' into move-default-args [81092b9]\n* Move guard against empty text segments further up in the chain\r [f121cac]\n* Support click on shape\r [83bfd8a]\n* Change ListBox to use a ProxyArray internally. This makes it so that ListBox's items forward Array methods to Array and then update the gui appropriately.\r [36546ab]\n\nBug Fixes (4)\n-------------\n\n* Fix text wrapping behavior - text isn't clipped and the background extens all the way\r [1e0d948]\n* Fix gem name check on uninstall avoiding stale executables\r [3707655]\n* Fix for hang during content alignment with text\r [e2a8bfd]\n* Fix download crash on missing Content-Length header\r [81f2ebb]\n\nContributors (7)\n----------------\n\nJason R. Clark, Tobias Pfeiffer, KC Erb, Eric Watson, Thomas Graves, David English, bx10000"
```

And here's what `puts` generates

```
Improvements (5)
----------------

 [63afeca]the hidden style on all elements
* Merge branch 'master' into move-default-args [81092b9]
 [f121cac]rd against empty text segments further up in the chain
 [83bfd8a]click on shape
 [36546ab]istBox to use a ProxyArray internally. This makes it so that ListBox's items forward Array methods to Array and then update the gui appropriately.

Bug Fixes (4)
-------------

 [1e0d948] wrapping behavior - text isn't clipped and the background extens all the way
 [3707655]name check on uninstall avoiding stale executables
 [e2a8bfd]hang during content alignment with text
 [81f2ebb]load crash on missing Content-Length header

Contributors (7)
----------------

Jason R. Clark, Tobias Pfeiffer, KC Erb, Eric Watson, Thomas Graves, David English, bx10000
```

If I `gsub` the `\r`s into `\n` then I get the expected output:
```
Improvements (5)
----------------

* Support the hidden style on all elements
 [63afeca]
* Merge branch 'master' into move-default-args [81092b9]
* Move guard against empty text segments further up in the chain
 [f121cac]
* Support click on shape
 [83bfd8a]
[...]

PR #1050, opened 2015-02-08
Get rid of \r from Changelog task

KCErb:
This addresses #1049. 

Does this merit a pull request? What's the custom here?


PragTob:
Overall this is tiny enough to do straight on master, on the other hand it's strange that they "\r" are there in the first place so a PR is nice to check :)

I figure I might have accidentally left them there :fearful: 

Thanks!
Tobi

PR #1051, opened 2015-02-08
First Appveyor Windows build running but failing

### LINKED ISSUE #1040: Have a Windows CI
Right now none of us is regularly running WIndows builds, which has caused us a bit of pain. As I learned at FOSDEM (yaaayy) there seems to be a free CI system for Windows: http://www.appveyor.com

I'll check that out and hopefully make it do what I want it to do - let's see how smooth getting JRuby there goes.

Greetings!
Tobi

EDIT: appveyor is already in there for PRs but it is not correctly configured yet, installing JRuby on Windows is... harsh :P

PragTob:
(Getting it into this stage took me a lot of trail and error - Windows and CLI.. though - squashed it though for your convenience)

Builds run on appveyor: http://www.appveyor.com/
Repository url: https://ci.appveyor.com/project/PragTob/shoes4/

The build is running now, in the sense that it installs and the tests run... however the specs fail with: 

```
Failure/Error: Unable to find matching line from backtrace
Java::OrgEclipseSwt::SWTError:
No more handles
```

I don't think it's like that on a real Windows machine, so maybe the server edition the tests are running on has less handles. Or maybe something else.

Solution strategies?

* Find a way to get rid of more handles during test runs
* Find a way to increase the number of handles on the appveyor systems
* Cry and give up on Windows builds :)

At some point this should fix #1040 

Tobi

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,4 +10,4 @@ build: off
 shallow_clone: true
 
 test_script:
-  - rake spec
+  - rake spec:shoes
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,13 @@
+install:
+  - appveyor DownloadFile https://s3.amazonaws.com/jruby.org/downloads/1.7.19/jruby-bin-1.7.19.zip
+  - 7z x jruby-bin-1.7.19.zip -y
+  - SET PATH=C:\projects\shoes4\jruby-1.7.19\bin;%PATH%
+  - gem install bundler
+  - bundle
+
+build: off
+
+shallow_clone: true
+
+test_script:
+  - rake spec


jasonrclark:
Probably just missing something, but where are you seeing that error? I just see a log which ends saying `io/console not supported; tty will not be manipulated`

Given that, do some of the specs run before cutting off? If so, then reducing handles might be an approach, although I had a hard time quantifying the handle counts when I was looking at leaks before. Definitely can imagine that they're on a Windows server with smaller (non-existent?) UI resource allocations available, though.

Alternatively, do the core specs run since they wouldn't be using any SWT handles? If so, that'd be better than nothing.

PragTob:
Super weird, the PR test doesn't seem to run the specs... here is one that had those: https://ci.appveyor.com/project/PragTob/shoes4/build/24-appveyor

So yeah, a couple of specs run. And yes the core specs do run :)

jasonrclark:
Ah, I see that. Didn't look far enough back in the history. I know I've run the specs on Windows before (although it's been a while), and didn't get those sort of handle failures. Betting it's something with AppVeyor's environment.

That said, it's also possible that our tests aren't being quick about cleaning up their SWT resources. Maybe a reasonable approach would be to configure things to run with the core specs, and get an issue logged to dig deeper on what's going on. I can look into whether we're "leaking" handles during the test runs at some point when I get more :athletic_shoe: time then, but would love to have even the basics running on Windows sooner rather than later.

Also might be worth a quick question to the AppVeyor folks. They might have seen this type of issue before.

PragTob:
The problem you saw with no output/test seems to be more serious... it just happens and the build hangs forever. We can't have a CI like that atm, I switched the builds over to my fork and opened some issues to see if anyone can find a solution witth this. I don't want to confuse people with broken builds, closing this for now.

jasonrclark:
That makes sense. Bummer that it's been so much work, but appreciate you taking it on @PragTob. You're a champ! :tada: :dancer: :rocket:

Issue #772, opened 2014-06-25
Gutter width with scrollbar not operating system independent


glenn-murray-bse:
Specs `Shoes::App#gutter when app has no scrollbar has gutter of 16` and `Shoes::App#gutter when app has a scrollbar has gutter of 16` fail on Windows. Windows returns '17' for gutter size.

```
1) Shoes::App#gutter when app has no scrollbar has gutter of 16
   Failure/Error: expect(app.gutter).to eq(16)

   expected: 16
        got: 17

   (compared using ==)
 # ./spec/shoes/app_spec.rb:300:in `(root)'

2) Shoes::App#gutter when app has a scrollbar has gutter of 16
   Failure/Error: expect(app.gutter).to eq(16)

   expected: 16
        got: 17

   (compared using ==)
 # ./spec/shoes/app_spec.rb:292:in `(root)'
```

I could fix this with operating system dependent test cases if testing for pixel perfect widths is in fact required. Thought may need to be given to best way forward for times with the windowing layer is different (I could only find a few fails_on_osx tests as examples of this)

seethemhigh:
Hey, guys. Running specs on Ubuntu 14.04 gives the gutter size of... 0! Here's the output of rspec:
```
Failures:

  1) Shoes::App#gutter when app has a scrollbar has gutter of 16
     Failure/Error: expect(app.gutter).to be_within(2).of(16)
       expected 0 to be within 2 of 16
     # ./shoes-core/spec/shoes/app_spec.rb:292:in `(root)'

  2) Shoes::App#gutter when app has no scrollbar has gutter of 16
     Failure/Error: expect(app.gutter).to be_within(2).of(16)
       expected 0 to be within 2 of 16
     # ./shoes-core/spec/shoes/app_spec.rb:300:in `(root)'

Finished in 2 minutes 4.2 seconds (files took 14.67 seconds to load)
2068 examples, 2 failures

Failed examples:

rspec ./shoes-core/spec/shoes/app_spec.rb:291 # Shoes::App#gutter when app has a scrollbar has gutter of 16
rspec ./shoes-core/spec/shoes/app_spec.rb:299 # Shoes::App#gutter when app has no scrollbar has gutter of 16
```
Here's how it looks like when there's no scrollbar...
![image](https://cloud.githubusercontent.com/assets/10567153/6106507/137227b2-b073-11e4-85e9-888d35df7ed3.png)

..and when there's one
[...]

PR #1058, opened 2015-02-11
Remove gutter specs, they are too flaky

PragTob:
We had to always gradually adjust them, and now it seems like
Ubuntu/Unity has an entirely different behavior,
see https://github.com/shoes/shoes4/issues/772#issuecomment-73503674

Therefore I think it's best to get rid of those specs

Input on whether or not that's the best route welcome :)

### DIFF
--- a/shoes-core/spec/shoes/app_spec.rb
+++ b/shoes-core/spec/shoes/app_spec.rb
@@ -283,25 +283,6 @@ describe Shoes::App do
     end
   end
 
-  describe "#gutter" do
-    context "when app has a scrollbar" do
-      let(:input_opts) { {width: 100, height: 100} }
-      let(:input_block) { Proc.new { para "Round peg, square hole" * 200 } }
-
-      it "has gutter of 16" do
-        expect(app.gutter).to be_within(2).of(16)
-      end
-    end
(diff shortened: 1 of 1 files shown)


jasonrclark:
I'm :+1: on this. Given how much these have moved from OS dependent stuff, they're costing more than they're gaining us.

PragTob:
A :+1: form @jasonrclark is enough for me to move forward with this, if anyone disagrees, it's easy to revert should that be needed :)

wasnotrice:
Pragmatic, Tobi :D It's too bad there's not a better way to specify this, but I agree we should remove them at least for now.

KCErb:
I think a :+1: from @jasonrclark is a form of currency in some regions . . .

PragTob:
Fun fact, PragTob in fact stands for "Pragmatic Tobi" - or at least that's what I envisioned the name to stand for when I thought of it. Back then I was just starting at Mendicant University and the `Pragmatic Programmer` was (still is) one of my favorite books. Together with its authors who are PragmaticAndy and PragDave on twitter :) Oh and the owners of my favorit epublisher - "PragProg"

PR #1059, opened 2015-02-12
Adjustments for rubocop offences

seethemhigh:
Hello. I've adjusted some rubocop offences further on. It's a pity that this PR comes so soon after you've merged in the similar one before :)

### DIFF
--- a/shoes-core/lib/shoes/common/inspect.rb
+++ b/shoes-core/lib/shoes/common/inspect.rb
@@ -9,7 +9,7 @@ class Shoes
       # Example:
       #   (Shoes::App:0x01234abc "Hello")
       def inspect
-        "(#{self.class.name}:#{'0x%08x' % (object_id * 2)}#{inspect_details})"
+        sprintf('(%s:%s%s)', self.class.name, hexy_object_id, inspect_details)
       end
 
       private
@@ -22,6 +22,10 @@ class Shoes
       # Additional details to include in the to_s representation.
       def to_s_details
         ''
+      end
+
+      def hexy_object_id
+        sprintf('0x%08x', object_id * 2)
       end
     end
   end
--- a/shoes-core/lib/shoes/app.rb
+++ b/shoes-core/lib/shoes/app.rb
@@ -114,8 +114,6 @@ class Shoes
       inspect_details
     end
 
-    public
-
     DELEGATE_BLACKLIST = [:parent, :app]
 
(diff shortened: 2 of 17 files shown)


shoes-core/lib/shoes/common/inspect.rb
```diff
@@ -9,7 +9,7 @@ def to_s
       # Example:
       #   (Shoes::App:0x01234abc "Hello")
       def inspect
-        "(#{self.class.name}:#{'0x%08x' % (object_id * 2)}#{inspect_details})"
+        sprintf('(%s:%s%s)', self.class.name, hexy_object_id, inspect_details)
```
PragTob: there is a rubocop offense here? That wants you to use `sprintf` ? I'd need to read the style guid to probably get the reasoning but I never used `sprintf` before in my ruby life so I'd be against adding it and eventually turning off the cop.

PragTob:
Hi there! Thanks for doing this :) It's no pity that it comes so soon, we like lots of small PRs this way they are focussed and we can get them in quickly :)

I only have one remark and that is more about rubocop than the changes - I'll see if I can check that out (or maybe you can help me with what rubocop said there) and we'll see if there are good reasons for `sprintf` or we'll deactivate the cop and use the old interpolated version :)

shoes-core/lib/shoes/common/inspect.rb
PragTob: aha, I see it's about the `%` in the middle. hmm. See [here](https://github.com/bbatsov/ruby-style-guide#sprintf)

PragTob:
Adjusted this to only use `sprintf` where needed and then merging after a little spec run

PragTob:
so thanks a lot, one thing about pull request size, what you did so far seems about right to me - not too large to review and still big enough so that it makes sense :)

seethemhigh:
Hey, rubocop was telling 
```
Favor format over String#%.
        "(#{self.class.name}:#{'0x%08x' % (object_id * 2)}#{inspect_details})"

```
At first I didn't get it, until I found this [article](http://batsov.com/articles/2013/06/27/the-elements-of-style-in-ruby-number-2-favor-sprintf-format-over-string-number-percent). I think it's better to use the interpolated version with  "'0x%08x' % (object_id * 2)" replaced with sprintf. Don't know why I replaced the whole thing with sprintf :) . So rubocop just doesn't like interpolations with %.

Thank you for your feedback! I'm looking forward to contribute more.

Issue #1055, opened 2015-02-11
rake samples causes invalid thread access error on mac


richardbatty:
I'm guessing you already know about this given that the readme says that bin/shoes doesn't work on mac, but rake samples uses bin/shoes and therefore also doesn't work. I get

```
LoadError: load error: shoes/swt -- org.eclipse.swt.SWTException: Invalid thread access
```

This is on version 4.0.0.pre3

KCErb:
Hi @richardbatty, thanks for trying Shoes!

I'm on OS X and `rake samples` works fine for me right now. Can you give me some more details about how to reproduce the error?

* Are you running from a fork of `shoes4` or are you in the gem (i.e. which Rakefile is `rake` talking to)?
* What version of OSX are you on?
* Have you run `bundle install`?
* Which files are in your `shoes4/bin` directory? You should have a `shoes-backend` file. If so what's inside?

richardbatty:
Hi,

I cloned the ```shoes4``` repo and then ran rake samples from inside that. So ```rake``` is talking to https://github.com/shoes/shoes4/blob/master/Rakefile

I'm on OSX 10.10.1

I've run bundle install

in ```shoes4/bin``` I have ```shoes```, ```shoes-swt``` and ```shoes.bat```. No ```shoes-backend``` file. As far as I can tell my ```shoes4/bin``` directory is the same as https://github.com/shoes/shoes4/tree/master/bin

KCErb:
Hmm,  I believe the `samples` task should call the samples with `bin/shoes` like you said. But `bin/shoes` is supposed to create a `shoes-backend` file the first time it runs. It's this file that makes `bin/shoes` work on OS X. (If you're curious to know why, you can read [these notes](http://kcerb.github.io/cordwainer/start/) I wrote up to remind myself how this stuff works).

I'm not sure why yours hasn't created it yet though . . .

Someone with more experience than me may chime in here and figure this out (:+1:) but until then can you try creating a file called `test.rb` and running it with `$ bin/shoes path/to/test.rb`. The contents should just be

```ruby
puts "Hello"
```

Copy and paste your terminal output for me. And then run `rake samples --trace` and let me see the full output from that too.

Thanks!

PragTob:
Hi there,

yeah could you please do what @KCErb suggested? The output of running `bin/shoes` is highly interesting to us. It should generate a script that is mac compatible on the first run.

Thanks for the issue report! :+1:

richardbatty:
[...]

PR #1060, opened 2015-02-13
Replace Bundler.require with Bundler.require(:shoes) in ui picker

richardbatty:
This is a fix for https://github.com/shoes/shoes4/issues/1055

I'm not sure if it's the right fix though, so critique away.

In ```shoes-core/lib/shoes/ui/picker.rb``` ```Bundler.require``` doesn't specify a gemset and so the default gemset is used. But because shoes has a .ruby-gemset file the gems get installed to the shoes gemset and so picker.rb should use ```Bundler.require(:shoes)```

### DIFF
--- a/shoes-core/lib/shoes/ui/picker.rb
+++ b/shoes-core/lib/shoes/ui/picker.rb
@@ -12,9 +12,10 @@ class Shoes
       # Only bundle if we find a local Gemfile.  This allows us to work properly
       # running from source without finding gem-nstalled backends.
       def bundle
-        return unless File.exists?("Gemfile")
-        require 'bundler/setup'
-        Bundler.require
+        if File.exist?("Gemfile")
+          require 'bundler/setup'
+          Bundler.require(:shoes)
+        end
       end
 
       def select_generator


PragTob:
Hi there,

thank you for that fix. I|m not quite sure it|s the right one, though.

Not everyone uses `rvm` though so not everyone has gemsets. It just picks the one from the gems currently used :)

So this shouldn't be necessary imo. I'll comment on the other issue.

PragTob:
Hmmm... `.require` gives the bundler group to load, there should be no such group that is called shoes, whcih makes this weird :/

jasonrclark:
So I'm pretty sure that this isn't right, since according to [Bundler's documentation](http://bundler.io/v1.3/groups.html) the parameters to `require` are related to the groups within the `Gemfile`. They don't have any direct relationship to the gemsets in RVM.

That said, it's an interesting result that passing in a value there ended up making things work for you. I don't know why that would make any difference. On the master branch, do things work if you remove the `.ruby-gemset` file entirely from the directory?

PragTob:
This should be fixed in #1069 - I'm guessing that this pull request worked because it essentially did not load gems (as there is no `shoes` group), circumventing the problem.

Could you please check out if this is solved for you on master? :)

Thanks to you for the report and thanks to @jasonrclark for the fix :)

Issue #778, opened 2014-07-03
Element click regions not right (need custom in_bounds? implementation)


jasonrclark:
Noticed in the course of testing out #768 that some of the drawing elements don't have an appropriate click region--i.e. they allow you to click outside of the actual drawn part of the screen. They all seem to just treat the enclosing rectangle as the click region.
- [ ] Arc
- [x] Line
- [x] Oval
- [ ] Star
- [ ] _Advanced_ General fix to make this work right with `rotate`, `displace`,  and `translate` (none of the `in_bounds?` checks behave quite right there.

seethemhigh:
Add Rect to it :( It doesn't want to behave properly when there's a "rotate" applied. Here's an app I tested on:
```ruby
Shoes.app do¬                                                                                                                                                                        fill '#333'¬                                                                                                                                                                 
    rotate 45¬                                                                                                                                                                   
    rect(30, 30, 40, 40).click { }¬                                                                                                                                              end¬                                                                                                                                                                           
```
The misbehavior is noticeable when I hover the pointer over rect corners.

wasnotrice:
Seems like we would have to apply the transforms we use when painting rects also when bounds-checking.

seethemhigh:
Yeah, we need to consider the transform. Looks like transformations are entrusted on backend and they make core-level 'in_bounds?' useless. Shall registering click regions be in the backend layer? I don't like that idea, it apears to me that backend should only be drawing things and getting input.

jasonrclark:
Ugh, those options aren't going to be nice. Either we rely on the backend for the transformations (which can be fairly clean, but potentially loads up the work for later backends), or re-implement transformations in core (which could diverge from the backend's actual drawing algorithms, and is just plain a bunch of work).

I think it's actually reasonable to split that work into a separate item from this one, and just leave this as the basic `in_bounds?` for the elements in question. Hopefully whichever way we do, we can apply the transform -> bounds thing systematically across all the elements rather than one-by-one.

PragTob:
Thanks for noticing this @seethemhigh ! :)

I'm also pro on a separate issue... I really don't want to implement transform/rotate on the front end. Which makes bounds checking even harder... I don't know if we should do it all in the backend, if there is a rotate set or take rotate into account (which might be hard).

[...]

PR #1061, opened 2015-02-14
Click regions for Shoes::Line

seethemhigh:
Hello. Here's a fix addressing #778 . Although clicking regions seem to be right, it's hard to hit a line since it requires pixel-level aiming. Shall we add some sloppiness to it?

### DIFF
--- a/shoes-core/lib/shoes/line.rb
+++ b/shoes-core/lib/shoes/line.rb
@@ -1,3 +1,5 @@
+require 'matrix'
+
 class Shoes
   class Line
     include Common::UIElement
@@ -17,6 +19,26 @@ class Shoes
 
       style[:x2] = point_b.x
       style[:y2] = point_b.y
+    end
+
+    # Check out http://math.stackexchange.com/questions/60070/checking-whether-a-point-lies-on-a-wide-line-segment
+    # for explanations how the algorithm works
+    def in_bounds?(x, y)
+      # c is (x, y)
+      left_most, right_most = point_a.x < point_b.x ? [point_a, point_b] : [point_b, point_a]
+      left_c = Vector.elements (left_most - [x, y]).to_a, false
+      left_right = Vector.elements (left_most - right_most).to_a, false
+
+      boldness = style[:strokewidth].to_i / 2
+      left_c_dot_left_right = left_c.inner_product(left_right)
+      left_right_dot_left_right = left_right.inner_product(left_right)
+
+      if left_c_dot_left_right.between?(0, left_right_dot_left_right)
+        left_c_dot_left_c = left_c.inner_product(left_c)
+        left_right_dot_left_right * left_c_dot_left_c <= boldness ** 2 * left_right_dot_left_right + left_c_dot_left_right ** 2
+      else
+        false
+      end
     end
 
     def update_style(new_styles)
--- a/shoes-core/spec/shoes/line_spec.rb
+++ b/shoes-core/spec/shoes/line_spec.rb
@@ -84,4 +84,36 @@ describe Shoes::Line do
       expect(subject.point_b).to eq(Shoes::Point.new(20, 20))
     end
   end
+
+  describe "#in_bounds?" do
+    subject(:line) { Shoes::Line.new(app, app, Shoes::Point.new(100, 100), Shoes::Point.new(50, 50), input_opts) }
+
+    it "returns true if a point is in the end of the line" do
+      expect(subject.in_bounds?(100, 100)).to be true
+    end
+
+    it "returns true if a point is in the end of the line" do
+      expect(subject.in_bounds?(50, 50)).to be true
+    end
+
+    it "returns true if a point is in the middle of the line" do
+      expect(subject.in_bounds?(75, 75)).to be true
+    end
+
+    it "returns false if a point is not in the line" do
+      expect(subject.in_bounds?(201, 200)).to be false
+    end
+
+    it "takes into account :strokewidth style" do
+      line = Shoes::Line.new(app, app, Shoes::Point.new(50, 50), Shoes::Point.new(70, 50), input_opts) 
+      line.style(strokewidth: 20)
+      expect(line.in_bounds?(50, 52)).to be true
+      expect(line.in_bounds?(50, 48)).to be true
+      expect(line.in_bounds?(50, 60)).to be true
+      expect(line.in_bounds?(50, 40)).to be true
+      expect(line.in_bounds?(49, 50)).to be false
+      expect(line.in_bounds?(50, 61)).to be false
+      expect(line.in_bounds?(70, 50)).to be true
+    end
+  end
 end
--- a/shoes-core/lib/shoes/point.rb
+++ b/shoes-core/lib/shoes/point.rb
@@ -4,6 +4,7 @@ class Shoes
 
     def initialize(x, y)
       @x, @y = x, y
+      @dimensions = 2
     end
 
     attr_accessor :x, :y
@@ -30,6 +31,18 @@ class Shoes
       Shoes::Point.new(@x + x, @y + y)
     end
 
+    def +(other)
+      other = other.to_a
+      assert_dimensions_match(other.length)
+      Shoes::Point.new(x + other[0], y + other[1])
+    end
+
+    def -(other)
+      other = other.to_a
+      assert_dimensions_match(other.length)
+      Shoes::Point.new(x - other[0], y - other[1])
+    end
+
     def width(other = self)
       (@x - other.x).abs
     end
@@ -47,10 +60,18 @@ class Shoes
       "(#{@x || nothing},#{@y || nothing})"
     end
 
+    def to_a
+      [x, y]
+    end
+
     private
 
     def inspect_details
       " #{self}"
+    end
+
+    def assert_dimensions_match(other_dimension)
+      raise ArgumentError, "Dimensions mismatch: expected #{@dimensions}D point, given #{other_dimension}D point" if @dimensions != other_dimension
     end
   end
 end
--- a/shoes-core/spec/shoes/point_spec.rb
+++ b/shoes-core/spec/shoes/point_spec.rb
@@ -29,6 +29,34 @@ describe Shoes::Point do
     end
   end
 
+  describe "#+" do
+    it "performs element-wise addition with other point" do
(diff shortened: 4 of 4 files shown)


shoes-core/lib/shoes/line.rb
```diff
@@ -19,6 +19,16 @@ def create_dimensions(point_a, point_b)
       style[:y2] = point_b.y
     end
 
+    def in_bounds?(x, y)
+      left_p, right_p = [point_a, point_b].sort_by { |a, b| [a.x, a.y] }
```
PragTob: I'd favor writing the whole name, e.g. `left_point` :)

shoes-core/lib/shoes/line.rb
```diff
@@ -19,6 +19,16 @@ def create_dimensions(point_a, point_b)
...
+      dx = x - left_p.x
+      dy = y - left_p.y
+      dyb = right_p.y - left_p.y
+      dxb = right_p.x - left_p.x
```
PragTob: I kind of get dx/dy but dxb/dyb I dn't get, would love more explanatory names here :-)

PragTob:
Looks good, I'd appreciate a bit more communicative naming.

Regarding adding "sloppiness", my first thought was to be pro on it but then my second thought was that if someone REALLY wants a line to be clickable they can just make the line thicker, which lead me to notice that the thickness of the line is unaccounted for as of now.

I'm unsure which option sets the line thickness... `weight` or soemthing it was I believe

shoes-core/lib/shoes/line.rb
seethemhigh: Whoops, these names are completely confusing. They remnants from when I was experimenting with in_bounds :) Thanks, a PR will follow soon.

PragTob: I don't get those either :) dx/dy is somewhat fine imo as it is common enough but I've no idea what rl stands for

jasonrclark:
Went through the same thought process as @PragTob... the sloppiness seemed good initially, but then would block someone from actually getting the precise result.

It's a good point about the line's width, though! It's actually `strokewidth` in the element's styles that takes care of setting that. ([Relevant code where the graphics context gets it](https://github.com/shoes/shoes4/blob/master/shoes-swt/lib/shoes/swt/common/stroke.rb#L34)

An example app in action with a thick line:

```
Shoes.app  height: 250 do
  line 0, 0, 100, 100, strokewidth: 8
end
```

image

seethemhigh:
@jasonrclark Ahh, I see now, I'll try to figure out some algorithm that would take that into account.

shoes-core/lib/shoes/line.rb
seethemhigh: Too bad I'm very anxious about the '80 chars limit' :) How about ydistance_right_left_points/xdistance_right_left_points or hight_right_left_points/width_right_left_points?

PragTob:
@seethemhigh what's the state of this? :) Do you fancy any help or something? :) Thanks for all the work!

seethemhigh:
@PragTob Oops, sorry, haven't looked at it since discussing that rotation bug mentioned in #778 . But! Here are some thoughts: when there's a strokewidth applied, a line turns into... a rectangular! Which is obvious :) So basically the current `in_bounds?` needs to check whether `x, y` is inside the rect. 

Another idea. Turning `x, y` params into a square with sides equal to strokewidth. Then we basically need to check if a line crosses the cursor rect.

I'll try those ideas in the coming days, hoping to finish this soon :)
Thanks for reminding!

seethemhigh:
A total rewrite of the method. The algorithm was shamelessly stolen away from [here](http://math.stackexchange.com/questions/60070/checking-whether-a-point-lies-on-a-wide-line-segment). Specs are kinda messy, aren't they? :)

shoes-core/lib/shoes/line.rb
```diff
@@ -19,6 +21,25 @@ def create_dimensions(point_a, point_b)
       style[:y2] = point_b.y
     end
 
+    def in_bounds?(x, y)
+      # c is (x, y)
+      # a is the left-most point
+      # b is the right-most point
+      a, b = [point_a, point_b].sort_by { |a, _| [a.x, a.y] }
```
PragTob: couldn't this be a comparison? Like point_a.x < point_.x ? And then assign values based on an if or a ternary?

shoes-core/lib/shoes/line.rb
```diff
@@ -19,6 +21,25 @@ def create_dimensions(point_a, point_b)
       style[:y2] = point_b.y
     end
 
+    def in_bounds?(x, y)
+      # c is (x, y)
+      # a is the left-most point
```
PragTob: I'd prefer a to be called `left_most` - makes the code clearer to read :)

shoes-core/lib/shoes/line.rb
```diff
@@ -19,6 +21,25 @@ def create_dimensions(point_a, point_b)
       style[:y2] = point_b.y
     end
 
+    def in_bounds?(x, y)
```
PragTob: A link to the page where you "stole" the algorithm from would be nice as a comment above the method :)

PragTob:
Great, thanks! :)

I left a couple of comments mostly regarding naming :)

I actually really like the specs, they might benefit from one or two more specs checking the correct working with boldness but that's no must.

Cheers and thanks a lot,
Tobi

shoes-core/lib/shoes/line.rb
```diff
@@ -19,6 +21,26 @@ def create_dimensions(point_a, point_b)
...
+      left_most, right_most = point_a.x < point_b.x ? [point_a, point_b] : [point_b, point_a]
+      vec_left_c = Vector[left_most.x - x, left_most.y - y]
+      vec_left_right = Vector[left_most.x - right_most.x, left_most.y - right_most.y]
+
+      boldness = style[:strokewidth].to_i / 2
+      vec_left_c_dot_vec_left_right = vec_left_c.inner_product(vec_left_right)
+      vec_left_right_dot_vec_left_right = vec_left_right.inner_product(vec_left_right)
+
+      if vec_left_c_dot_vec_left_right.between?(0, vec_left_right_dot_vec_left_right)
+        vec_left_c_dot_vec_left_c = vec_left_c.inner_product(vec_left_c)
```
seethemhigh: Variables names read funny a bit?

shoes-core/spec/shoes/line_spec.rb
```diff
@@ -84,4 +84,33 @@
...
+    it "returns false if a point is not in the line" do
+      expect(subject.in_bounds?(201, 200)).to be false
+    end
+
+    it "takes into account :strokewidth style" do
+      line = Shoes::Line.new(app, app, Shoes::Point.new(50, 50), Shoes::Point.new(70, 50), input_opts) 
+      line.style(strokewidth: 20)
+      expect(line.in_bounds?(50, 52)).to be true
+      expect(line.in_bounds?(50, 48)).to be true
+      expect(line.in_bounds?(50, 60)).to be true
```
seethemhigh: More examples

shoes-core/lib/shoes/line.rb
```diff
@@ -19,6 +21,26 @@ def create_dimensions(point_a, point_b)
       style[:y2] = point_b.y
     end
 
+    # Check out http://math.stackexchange.com/questions/60070/checking-whether-a-point-lies-on-a-wide-line-segment
+    # for explanations how the algorithm works
+    def in_bounds?(x, y)
+      # c is (x, y)
+      left_most, right_most = point_a.x < point_b.x ? [point_a, point_b] : [point_b, point_a]
+      vec_left_c = Vector[left_most.x - x, left_most.y - y]
```
seethemhigh: `vec_left_c = Vector[left_most.x - x, left_most.y - y]`
Maybe it'd be better to add `+/-` methods to `Shoes::Point`?

PragTob: Sounds like a good idea, please go ahead and do so!

shoes-core/lib/shoes/line.rb
PragTob: Hm, yes, they read funny... I don't know. Math is too long ago for me to know if there is a good name for the product (like what it essentially means). Maybe leaving out the vec? `left_right` makes it a bit clear that it is a vector. You are right, right now the names are a bit too long.

Sorry for the name nit picking, it's a pet peeve of mine :)

shoes-core/spec/shoes/line_spec.rb
PragTob: Looks good :) I'd also like to include some negatives: (49, 50), (50, 61)

And also test if (70, 50) is actually inside :)

I can add those as well :)

Thanks!

PragTob:
Thanks for the adjustments :+1:

shoes-core/lib/shoes/line.rb
seethemhigh: I mean, maybe something like
```
class Point
  ...
  def -(other)
    other = other.to_a
    Point.new(x - other[0], y - other[1])
  end
  
  def to_a
    [x, y]
  end
end

```
And then it would allow to
```
p0 - p1
p0 - [1, 2]
```

PragTob: hm. I'd probably rather have:

```
def +(other)
  Point.new(x - other.x, y - other.y)
end
```

Of course that'd require a `Point.new` each time around, your solution with to_a is quite elegant, though :)

I'm not too sure. Choose what you like best

shoes-core/lib/shoes/line.rb
seethemhigh: Nah don't worry. It's not hard to change those names :) I'll remove `vec` from the names.

shoes-core/spec/shoes/line_spec.rb
seethemhigh: Ugh.. I forgot to add those :) Maybe tomorrow or you add them :)

PragTob: On my way to work right now and this evening I'm moderating the ruby user group berlin, so I'll get to it tomorrow at the earliest :)

seethemhigh: I'll add those missing ones and squash the PR :)

PragTob:
Looks great, thanks - moved this weekend hence the lateness :)

PR #1062, opened 2015-02-14
Refinements

seethemhigh:
There you go :) These are absolutely not critical, but make code more idiomatic (do they?). These changes were not required by Rubocop.

### DIFF
--- a/shoes-core/lib/shoes/color.rb
+++ b/shoes-core/lib/shoes/color.rb
@@ -61,11 +61,7 @@ EOS
 
     def <=>(other)
       raise_class_mismatch_error(other) unless other.is_a?(self.class)
-      if same_base_color?(other)
-        @alpha <=> other.alpha
-      else
-        less_or_greater_than other
-      end
+      [red, green, blue, alpha] <=> [other.red, other.green, other.blue, other.alpha]
     end
 
     # @return [String] a hex represenation of this color
@@ -95,20 +91,6 @@ EOS
     def raise_class_mismatch_error(other)
       fail ArgumentError,
            "can't compare #{self.class.name} with #{other.class.name}"
-    end
-
-    def same_base_color?(other)
-      @red == other.red && @green == other.green && @blue == other.blue
-    end
-
-    def less_or_greater_than(other)
-      own_sum   = @red + @green + @blue
-      other_sum = other.red + other.green + other.blue
-      if own_sum > other_sum
-        1
-      else
-        -1
-      end
     end
 
     class HexConverter
--- a/shoes-core/lib/shoes/color.rb
+++ b/shoes-core/lib/shoes/color.rb
@@ -60,7 +60,6 @@ EOS
     end
 
     def <=>(other)
-      raise_class_mismatch_error(other) unless other.is_a?(self.class)
       [red, green, blue, alpha] <=> [other.red, other.green, other.blue, other.alpha]
     end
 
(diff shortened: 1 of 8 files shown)


shoes-core/lib/shoes/color.rb
```diff
@@ -61,11 +61,7 @@ def black?
 
     def <=>(other)
       raise_class_mismatch_error(other) unless other.is_a?(self.class)
-      if same_base_color?(other)
-        @alpha <=> other.alpha
-      else
-        less_or_greater_than other
-      end
+      [red, green, blue, alpha] <=> [other.red, other.green, other.blue, other.alpha]
```
PragTob: This seems rather neat, it breaks behavior (that isn't fully caught in specs) though, or at least I believe so. 

With the current code, comparing [255, 0, 0] and [110, 110, 110] would result in the latter being better, with this the first one is bigger. 

Saying when a color is bigger than the other is hard business, the way we implement it right now is hoe shoes3 does it.

Your solution certainly is more elegant. 

I'm unsure here.

PragTob:
Thanks, looks better, I especially like the usage of Forwardable. I'm unsure about the color handling and would love someone else to review this as well.

shoes-core/lib/shoes/color.rb
seethemhigh: Yep, it's definitely my overlook. It's hard to reason about color comparison, but if that breaks compatibility with other shoes I would rather stay away from merging that in.

seethemhigh:
If that goes merging in, I could squash the whole thing into one (does Github supports it? Squashing commits which were sent in a PR)

shoes-core/lib/shoes/color.rb
```diff
@@ -60,12 +60,8 @@ def black?
...
 
     def <=>(other)
-      raise_class_mismatch_error(other) unless other.is_a?(self.class)
-      if same_base_color?(other)
-        @alpha <=> other.alpha
-      else
-        less_or_greater_than other
-      end
+      return unless other.is_a?(self.class)
+      [red, green, blue, alpha] <=> [other.red, other.green, other.blue, other.alpha]
```
jasonrclark: Really like the use of `[]` + `<=>` to accomplish these comparisons.

jasonrclark:
Commented over on the commit where you'd done @seethemhigh about the `raise`. FWIW, github displays the comments a little nicer if you make them from the pull request page itself (ala https://github.com/shoes/shoes4/pull/1062/files) rather than the specific commit view (https://github.com/seethemhigh/shoes4/commit/7c97c62375c7688eac3f8af47b2f162050cc08b3).

As far as squashing, I personally don't worry about it and we don't really try to enforce it on this project. How that works, though, is that you'd squash the commits on your branch, and then just force-push them to your fork. Github sees the changes and incorporates them into the existing PR just fine.

seethemhigh:
@jasonrclark thank you for the advices! Learnt a few things :)

jasonrclark:
@seethemhigh Always glad to help. Thanks for pitching in on Shoes!

shoes-core/lib/shoes/color.rb
PragTob: yeah we should probably adjust it and spec it out bettter :)

PragTob:
Sorry that this was stale for so long, feel free to ping us in those cases :)

Tobi

PR #1063, opened 2015-02-14
Complete renaming has_textcursor? to textcursor?

seethemhigh:
Ooops looks like I've forgotten to complete the renaming in #1059 . Tests were green, but now I've found out that samples/good-clock.rb shows an error. I apologize, guys. 
Btw, it's no good that the test doesn't react to the renaming. Maybe we should force the double to recieve ':textcursor?'

### DIFF
--- a/shoes-swt/lib/shoes/swt/text_block/cursor_painter.rb
+++ b/shoes-swt/lib/shoes/swt/text_block/cursor_painter.rb
@@ -42,7 +42,7 @@ class Shoes
         end
 
         def remove_textcursor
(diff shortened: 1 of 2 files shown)


PragTob:
Most definitely! Very bad that the tests don't catch this. We should move on to use [instance_double](https://www.relishapp.com/rspec/rspec-mocks/v/3-0/docs/verifying-doubles/using-an-instance-double) more. If you want to please go ahead, in the meantime I'll emrge this.

Thanks :)

PR #1064, opened 2015-02-18
Check disposal during startup

### LINKED ISSUE #1053: Shoes quit/close may fall?
Hello. Running a simple app that, from my perspective, should only render a window and be gone immediately,

example 1 
```ruby
Shoes.app { quit }
```
but instead it throws an exception "Java::OrgEclipseSwt::SWTException: Widget is disposed"

Running example 2 works fine though

example 2
```ruby
Shoes.app do
  keypress { |k| quit }
end
```

This is the intended behavior? I'm running on Ubuntu 14.04. Thanks

jasonrclark:
If `quit` was called during startup, our SWT resources might be disposed of
and the initial open will crash.

Fixes #1053. Just happened to glance at it to determine what was actually
crashing and accidentally fixed it :blush:

### DIFF
--- a/shoes-swt/lib/shoes/swt/app.rb
+++ b/shoes-swt/lib/shoes/swt/app.rb
@@ -35,6 +35,10 @@ class Shoes
       end
 
       def open
+        # If something called quit during the app block's initial evaluation
+        # we might already be disposed of, in which case get out of here!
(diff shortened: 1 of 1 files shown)


PragTob:
:+1: you're the best!

Issue #1066, opened 2015-02-24
ask_save_file does not allow user to type a filename


tlemburg:
First of all, love what everyone has going on here, and I'm looking forward to possibly contributing in the future. I'm using Shoes4, Mac OSX 10.10, and running samples/simple-dialogs.rb, and trying the save file dialog. But this dialog doesn't actually give me the option to type in a filename to save a new file. I can click an existing file to use that filename, but this is typically not what users want out of "Save File". I'm not sure if this works correctly in other OS's but it seems that this is just a weird oversight with which SWT dialog to query? I'm kind of a noob in this kind of thing, so any help would be appreciated. Thanks!

1. Run samples/simple-dialogs.rb
2. Click Save File...
3. Following dialog appears.
![screen shot 2015-02-23 at 8 31 56 pm](https://cloud.githubusercontent.com/assets/1199848/6342221/0e4ccc46-bb9b-11e4-9a0b-a20b63130d47.png)

tlemburg:
Upon further investigation, it seems the dialog_chooser method in shoes-swt/lib/shoes/swt/dialog.rb only uses the SWT::OPEN constant and not the SWT::SAVE constant. So I made an early pull request: https://github.com/shoes/shoes4/pull/1067

I will defer to the stylistic choices of the community if we want a little different separation of code here. I can also write tests, but I will have to look into it tomorrow. Thanks! :)

seethemhigh:
Can confirm the same issue on Ubuntu 14.04. Here's how it looks like
![screenshot from 2015-02-25 20 55 26](https://cloud.githubusercontent.com/assets/10567153/6376822/5d51b62a-bd31-11e4-852c-eb5237a77c6d.png)

PragTob:
Hi there, thanks a lot for the report! :)

Sorry for the late reaction, lots of stuff happening in the lifes of lots of shoes contributors atm (I'm moving flats atm :) )

Thanks so much - I'll check out the PR and see that this will get fixed.

PS: @seethemhigh nice device names... RIP Leonard Nimoy

tlemburg:
Pull request was merged so I'll close this.

PragTob:
Yes, thank you very much! :)

PR #1067, opened 2015-02-24
Fix to allow save file/folder dialogs

tlemburg:
Adds use of the ::Swt::SWT::SAVE constant for save file and folder dialogs to dialog_chooser method. Thus ask_save_file and ask_save_folder will work as intended.

### DIFF
--- a/shoes-core/lib/shoes/builtin_methods.rb
+++ b/shoes-core/lib/shoes/builtin_methods.rb
@@ -44,7 +44,7 @@ class Shoes
     end
 
     def ask_save_file
-      Shoes::Dialog.new.dialog_chooser 'Save File...'
+      Shoes::Dialog.new.dialog_chooser 'Save File...', false, :save
     end
 
     def ask_open_folder
@@ -52,7 +52,7 @@ class Shoes
     end
 
     def ask_save_folder
-      Shoes::Dialog.new.dialog_chooser 'Save Folder...', :folder
+      Shoes::Dialog.new.dialog_chooser 'Save Folder...', :folder, :save
     end
 
     def ask(msg, args = {})
--- a/shoes-core/lib/shoes/dialog.rb
+++ b/shoes-core/lib/shoes/dialog.rb
@@ -12,8 +12,8 @@ class Shoes
       @gui.confirm msg
     end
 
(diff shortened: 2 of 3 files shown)


jasonrclark:
Hey @tiemburg, thanks for the contribution! That looks right to me on Mac 10.9 (which appears you're running 10.10 from your screenshot over in #1066).

Would be very interested to get any other OS testing on this before we merge the changes, although I would hope that SWT would do the "right" thing per OS when asked for the open vs save dialogs.

As far as testing, the dialogs aren't super well-tested, but the place I'd consider adding something is https://github.com/shoes/shoes4/blob/master/shoes-swt/spec/shoes/swt/dialog_spec.rb  Right now it's a bit thin around the save dialogs, but you can see from the others how we were stubbing things out to avoid actually spawning windows during the tests.

Thanks again!

seethemhigh:
Hey @tlemburg ! I tested your changes on Ubuntu 14.04. Your change makes the field to appear but clicking 'save' doesn't actually save a file. Steps I did when testing:
1. run samples/simple-dialogs.rb
2. open file with ```Open File```
2. save file with a different name
Have you got the same issue?

shoes-core/lib/shoes/builtin_methods.rb
```diff
@@ -52,7 +52,7 @@ def ask_open_folder
     end
 
     def ask_save_folder
-      Shoes::Dialog.new.dialog_chooser 'Save Folder...', :folder
+      Shoes::Dialog.new.dialog_chooser 'Save Folder...', :folder, :save
```
PragTob: I almost commented saying that instad of `:save` we could pass in the style directly. But this is in `shoes-core` and therefore rightfully SWT free - good catch and engineering! :)

PragTob:
dialogs are one of the oldest parts of the code base, hence not as well tested and in need of some refactoring love (fun fact, I believe dialogs were one of the first things I ever worked on with shoes4). I think the structure is different than everywhere else as of now.

Will test on Linux and then save :)

PragTob:
Thanks heaps!

PragTob:
Note to self, remember to tag it with changelog :-/

PR #1069, opened 2015-03-01
Don't Bundler.require during Picker

### LINKED ISSUE #1055: rake samples causes invalid thread access error on mac

jasonrclark:
Fixes #1055

The Shoes::UI::Picker will use Bundler and the local Gemfile when
selecting a backend in local development. Previously this would do a
`Bundler.require` during the Picker's startup, but on newer versions of
JRuby (1.7.19 at least), this will hit the dreaded Thread Access
exception from SWT when those requires touch some font related files.

I haven't gotten to the bottom of why those hit thread access problems
when we didn't on earlier JRuby's, but looking at the code we use there
it doesn't seem unreasonable that it could fail in that way.

Luckily, the Picker shouldn't actually need the gems required--we just
need the paths to search for generators, which will be loaded up
directly to run.

### DIFF
--- a/shoes-core/lib/shoes/ui/picker.rb
+++ b/shoes-core/lib/shoes/ui/picker.rb
@@ -13,8 +13,11 @@ class Shoes
       # running from source without finding gem-installed backends.
       def bundle
         return unless File.exists?("Gemfile")
+
+        # Only need bundler/setup to get our paths right--we don't need to
+        # actually require the gems, since we find the generate-backend.rb's
+        # and just require them directly.
         require 'bundler/setup'
-        Bundler.require
       end
 
       def select_generator


PragTob:
Amazing - thanks so much!

Do you fancy opening a jruby issue so maybe they can check up what is going on and if they intend for that to happen?

jasonrclark:
Great idea @PragTob. Will get a minimal case together and follow up on JRuby about it.

jasonrclark:
Dug in on my way back from Ruby on Ales (:beer:). I found that my original diagnosis of a JRuby difference was wrong--it was just that I had different Bundler versions on my JRuby installations (old JRuby had Bundler 1.7.4, new version 1.8.3)

Turns out because of https://github.com/bundler/bundler/pull/3195, the older version of Bundler was incorrectly swallowing the thread access exceptions when the picker tried to load `shoes/swt`. So the exception had been happening all along, but the newer Bundler (correctly!) started propagating that out to us.

So, no bugs for anyone else as a result of this. :tada:

PragTob:
:+1:  thanks for digging into this!

PR #1074, opened 2015-03-17
Shoes layout specs

### LINKED ISSUE #994: Shoes Layout specs
Right now [shoes_layout](https://github.com/shoes/shoes4/blob/master/shoes-swt/lib/shoes/swt/shoes_layout.rb) has no specs. It is a crucial part of our swt backend implementation and the source of some bugs over time.

Having specs cover those would be VERY nice.

edit: specs shouldn't be that hard I believe, since the only external dependency that gets set is `gui_app`, everything else derives from there :)

jasonrclark:
Fixes #994

This is split between two commits. First one I wrapped specs around the whole of `ShoesLayout`.

Then with that safety-net I did a major refactoring. Pulled everything up into one method, then broke it back down as actually made sense to me. Hopefully the end result is easier to follow. If we don't think so, I can easily back that last commit off and leave just the shiny new specs.

Should be no functional differences, ran about half the samples (and most of the big ones) without any trouble.

### DIFF
--- a/shoes-swt/lib/shoes/swt/shoes_layout.rb
+++ b/shoes-swt/lib/shoes/swt/shoes_layout.rb
@@ -14,6 +14,8 @@ class Shoes
         end
       end
 
+      private
+
       def gather_height_info(dsl_app)
         height = dsl_app.height
         scroll_height = dsl_app.top_slot.contents_alignment
--- a/shoes-swt/lib/shoes/swt/shoes_layout.rb
+++ b/shoes-swt/lib/shoes/swt/shoes_layout.rb
@@ -4,42 +4,54 @@ class Shoes
       attr_accessor :gui_app
 
       def layout(*_dontcare)
-        dsl_app = @gui_app.dsl
-        height = dsl_app.height
-        scrollable_height = dsl_app.top_slot.contents_alignment
-        set_gui_size(height, scrollable_height)
+        height, scroll_height, is_scrolling = gather_height_info(@gui_app.dsl)
+        set_heights(@gui_app.dsl, @gui_app.real, height, scroll_height)
 
-        dsl_app.top_slot.width  = dsl_app.width
-        dsl_app.top_slot.height = [scrollable_height, height].max
-
-        vertical_bar = @gui_app.shell.getVerticalBar
-        vertical_bar.setVisible(scrollable_height > height)
-        if scrollable_height > height
-          handle_scroll_bar(vertical_bar, height, scrollable_height)
+        if is_scrolling
+          show_scrollbar(vertical_scrollbar, height, scroll_height)
         else
-          set_gui_location
+          hide_scrollbar(vertical_scrollbar, @gui_app.real)
         end
       end
 
-      private
-
-      def set_gui_size(height, scrollable_height)
-        width          = @gui_app.dsl.width
-        maximum_height = [scrollable_height, height].max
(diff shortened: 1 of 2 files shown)


shoes-swt/lib/shoes/swt/shoes_layout.rb
```diff
@@ -4,42 +4,54 @@ class ShoesLayout < ::Swt::Widgets::Layout
...
       end
 
-      def handle_scroll_bar(vertical_bar, height, scrollable_height)
-        vertical_bar.setThumb height * height / scrollable_height
-        vertical_bar.setMaximum scrollable_height - height + vertical_bar.getThumb
-        vertical_bar.increment = 10
+      def show_scrollbar(scrollbar, height, scroll_height)
+        scrollbar.setVisible true
+        scrollbar.setThumb height * height / scroll_height
+        scrollbar.setMaximum scroll_height - height + scrollbar.getThumb
```
PragTob: this doesn't use ruby style method calls, and we wanted to do ruby style method calls - maybe we should put that in code conventions somewhere :) 

Or put a comment here, why we can't use ruby style accessors here

shoes-swt/lib/shoes/swt/shoes_layout.rb
```diff
@@ -4,42 +4,54 @@ class ShoesLayout < ::Swt::Widgets::Layout
...
-          handle_scroll_bar(vertical_bar, height, scrollable_height)
+        if is_scrolling
+          show_scrollbar(vertical_scrollbar, height, scroll_height)
         else
-          set_gui_location
+          hide_scrollbar(vertical_scrollbar, @gui_app.real)
         end
       end
 
-      private
```
PragTob: I'm actually in favor of keeping private around. It helps me distinguish which methods are the public interface and which are just an implementation detail and should not be called from the outside. Happy to hear other opinions as always :D

PragTob:
Looks good, left two minor comments otherwise good. Thanks for tackling this important topic!

shoes-swt/lib/shoes/swt/shoes_layout.rb
jasonrclark: Yeah, had noticed those were old-school camel-cased. Will get them updated since as far as I know there's no reason we need that style here.

shoes-swt/lib/shoes/swt/shoes_layout.rb
jasonrclark: Happy enough to put it back. I don't tend to use them by default--often times it's a smell if I need tons of private methods anyway--but it seems useful here. Thanks!

PragTob: That private is a smell that there's probably another object or so hiding in the class was one of the first and major things Steve taught me about programming back in 2011 - good times :)

Tend to use `private` quite a bit myself as I heavily extract even one line methods to highlight a concept or isolate change to that one method :) Still, all approaches valid :)

jasonrclark:
Made the updates you pointed out, and definitely happier with the results. :ship:ing

jasonrclark:
Oh, and looks like we already comment on the snake_case vs camelCase issue on the wiki: https://github.com/shoes/shoes4/wiki/Style-Guide#use-ruby-notation-for-methods-coming-from-java

Issue #1075, opened 2015-03-18
Finishing the download of an image does not trigger a redraw


PragTob:
```
Shoes.app do
  image 'http://shoesrb.com/img/shoes-icon.png'
end
```

running this code you will see the "Downloading..." forever, unless you trigger a redraw (for instance by resizing the window). I believe this used to work, something does an asyncExec on the main thread to redraw, need to check what went wrong here.

PragTob:
Fixed by @seethemhigh :+1:

PR #1076, opened 2015-03-20
Fix an image redraw bug

seethemhigh:
Hello :) This PR addresses #1075 . Looks likes the reason why it bugged is that `update_image` was never called.


PragTob:
Yes, but most of it seems to be adding the redraw trigger for `create_image` - thanks a lot!

(Actually great that this got fixed, giving a shoes talk tomorrow and this kind of prevented a demo from fully working but ssshhh) :)

THANKS! :+1:

seethemhigh:
I'm glad it's helpful :)

Issue #1078, opened 2015-03-22
Performance - simple-manual.rb


PragTob:
It might just be me but `simple-manual.rb` seems A LOT slower on master. A lot as in clicking on `Colors` takes a felt `~ 10 seconds` to open and then scrolling is awfully slow.. even laggy where scrolling takes multiple seconds some times.

We need to investigate to see what introduced this (if reproducible) and fix it or just a good old general performance analysis. 

If it is not a degradation, then well. still FIX it :)

PragTob:
Went as far back as to pre2. On pre 2 it is still slow, but load feels at least `~3x` faster between pre2 and pre3. So that's a start - dunno if the best approach is to do a general performance analysis or look what happened inbetween.

Maybe it's just my subjectiveness. Any confirmations? :)

jasonrclark:
Confirmed very briefly that things seemed worse between pre2 -> pre3, and then running off master seemed worse still than pre3.

Don't have time before :sleeping: tonight, but I'm keen to bisect it at least given how many of the changes in that timeframe are mine :) I've also been doing a lot of perf work in my day job, so I'm keen to put some of what I've been using to work there as well.

Given the confirmation and my interest, tagging this for pre4 to at least assess what can be improved.

PragTob:
Through great work by @jasonrclark I believe we are at an even better stage than pre2 now - thanks you :metal: :rocket:

PragTob:
As written before, we went to even better perf than before - closing this one!

PR #1079, opened 2015-03-24
Improve perf on Common::Visibility

jasonrclark:
First of what I expect to be a number of PR's spawned from #1078. I'm having fun doing perf tuning now that I've started digging into it :rocket: 

The bottleneck here proved to be accessing `#style` when doing visibility updates. This did heavy lifting every time to update the dimensions in the styles hash, despite the fact they were almost never changing. This PR updates to just touch the styles directly when we write a dimension, instead of trying to frequently reconstruct it on reading.

I located this with a basic bisect between `v4.0.0.pre3` and `master` where the perf issue @PragTob had highlighted was even worse than he'd reported between `pre2` -> `pre3`. It resolved to 5d2e6196e3ce84f2d9c0771a376fa38b68aba9d8, and then it was fairly simple to isolate by commenting things out to determine the primary culprit here.

While it's a little ugly to have the dimension updating touch the styles, I found it more confusing to try and separate those out to a different unrelated module.

### DIFF
--- a/shoes-core/lib/shoes/common/initialization.rb
+++ b/shoes-core/lib/shoes/common/initialization.rb
@@ -14,6 +14,7 @@ class Shoes
 
         style_init(styles)
         create_dimensions(*args)
+        update_dimensions if styles_with_dimensions?
 
         @parent.add_child self
         @gui = Shoes.backend_for self
--- a/shoes-core/lib/shoes/common/style.rb
+++ b/shoes-core/lib/shoes/common/style.rb
@@ -29,7 +29,6 @@ class Shoes
       # Adds styles, or just returns current style if no argument
       def style(new_styles = nil)
         update_style(new_styles) if need_to_update_style?(new_styles)
-        update_dimensions if styles_with_dimensions?
         @style
       end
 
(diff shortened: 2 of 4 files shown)


PragTob:
Great, thank you! Short comment: "Can I haz benchmark pleasse?" (or check if it is visible in the benchmarks we already have)

wasnotrice:
My 2-year-old says all the time, "ROCKET SHOES!"

This is clearly what he means :rocket: :high_heel: :high_heel:

jasonrclark:
Great point @PragTob. I'll see about adding clear benchmark(s) if things don't show up in the existing one. Will probably add it on master and rebase so we can get an easy before and after on the changes.

PragTob:
Code looks good by the way, thanks a lot :+1:

jasonrclark:
Rebased off master with [`benchmark/visibility.rb`](https://github.com/shoes/shoes4/blob/master/benchmark/visibility.rb) added to expose this case.

**Before**
```
✔ ♥♥♥♥ [jruby] (master)*
~/source/shoes4: bin/shoes benchmark/visibility.rb
       user     system      total        real
100  0.050000   0.000000   0.050000 (  0.042000)
1000  0.460000   0.010000   0.470000 (  0.396000)
10000  4.120000   0.050000   4.170000 (  3.856000)
```

**After**
```
✔ ♥♥♥♥ [jruby] (visibility-styling-perf)*
~/source/shoes4:bin/shoes benchmark/visibility.rb
       user     system      total        real
100  0.020000   0.000000   0.020000 (  0.011000)
1000  0.080000   0.000000   0.080000 (  0.074000)
10000  0.760000   0.010000   0.770000 (  0.701000)
```

PragTob:
Really great, thanks a lot! :D :rocket:

PR #1080, opened 2015-03-24
Performance tuning in Dimension

jasonrclark:
Another perf PR inspired by #1078.

Running with [JRuby's --profile option](https://github.com/jruby/jruby/wiki/Profiling-jruby) revealed that we were spending a really huge amount of time in the various `Dimension` methods which were being meta-programmed up. It was up to 5x faster by my benchmarking to just do things directly, so I've unrolled the meta in favor of speed.

### DIFF
--- a/shoes-core/lib/shoes/dimension.rb
+++ b/shoes-core/lib/shoes/dimension.rb
@@ -31,8 +31,8 @@ class Shoes
     def extent
       result = @extent
       if @parent
-        result = calculate_relative(result) if relative?(result)
-        result = calculate_negative(result) if negative?(result)
+        result = calculate_relative(result) if @extent_relative
+        result = calculate_negative(result) if @extent_negative
       end
       result
     end
@@ -40,6 +40,10 @@ class Shoes
     def extent=(value)
       @extent = value
       @extent = parse_from_string @extent if string? @extent
+
+      @extent_relative = relative?(@extent)
+      @extent_negative = negative?(@extent)
+
       @extent
     end
 
@@ -100,30 +104,53 @@ class Shoes
     # getter... reason being that for ParentDimensions we need to be able to
     # figure out if a value has been modified or if we should consult the
     # parent value - see ParentDimension implementation
-    [:margin_start, :margin_end, :displace_start].each do |method|
-      define_method method do
-        instance_variable_name = '@' + method.to_s
-        value = instance_variable_get(instance_variable_name) || 0
-        value = calculate_relative value if relative? value
-        value
-      end
+    def margin_start
+      value_factoring_in_relative(@margin_start, @margin_start_relative)
     end
 
-    def self.define_int_parsing_writer(name)
-      define_method "#{name}=" do |value|
-        instance_variable_set("@#{name}", parse_int_value(value))
-      end
+    def margin_end
+      value_factoring_in_relative(@margin_end, @margin_end_relative)
     end
 
-    %w(start end margin_start margin_end displace_start).each do |method|
-      define_int_parsing_writer method
+    def displace_start
+      value_factoring_in_relative(@displace_start, @displace_start_relative)
+    end
+
+    def value_factoring_in_relative(value, relative)
+      value ||= 0
+      value = calculate_relative value if relative
+      value
+    end
+
+    def start=(value)
+      @start_relative = relative?(value)
+      @start = parse_int_value(value)
+    end
+
+    def end=(value)
+      @end = parse_int_value(value)
+    end
+
+    def margin_start=(value)
+      @margin_start_relative = relative?(value)
+      @margin_start = parse_int_value(value)
+    end
+
+    def margin_end=(value)
+      @margin_end_relative = relative?(value)
+      @margin_end = parse_int_value(value)
+    end
+
+    def displace_start=(value)
+      @displace_start_relative = relative?(value)
+      @displace_start = parse_int_value(value)
     end
 
     private
 
     def basic_start_value
       value = @start
-      value = calculate_relative value if relative?(value)
+      value = calculate_relative value if @start_relative
       value = report_relative_to_parent_start if value.nil?
       value
     end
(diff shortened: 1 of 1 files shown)


jasonrclark:
FWIW, when looking at a merge with both this and #1079, perf on the colors page in the manual was back up to `pre2` levels for me (anecdotally).

I'm interested to continue peeling back a bit more on the perf onion, though, to see if there's any other easy, low-hanging fruit that can get large apps like the manual running smoothly rather than just acceptably.

PragTob:
Thanks a lot for looking at this and fixing this :)

I'm curios though why there is such a big difference. Sure those methods are called a lot and they do more lookups to dynamically get the instance variable... I wonder if we could use the `define_method` version with an interpolated string to get rid off this "extra work" we did, because then theoretically that defined method should be the same as the written out versions.

Nothing to change here, just something for my personal learning and interest

jasonrclark:
I was as surprised as you are the impact of it. When I do the benchmarking I'll do a bit of comparison, but I found that actually each bit of the meta-work had a significant burden by itself over just doing it directly. Interesting stats to come in the near future, and thanks for asking about it! :)

jasonrclark:
Also rebased off master to provide [`benchmark/dimension.rb`](https://github.com/shoes/shoes4/blob/master/benchmark/dimension.rb) before and after

**Before**
```
✔ ♥♥♥♥ [jruby] (master)*
~/source/shoes4:bin/shoes benchmark/dimension.rb
       user     system      total        real
1_000_000x margin_start  3.040000   0.050000   3.090000 (  2.727000)
1_000_000x element_extent  6.810000   0.030000   6.840000 (  6.509000)
```

**After**
```
✔ ♥♥♥♥ [jruby] (less-meta-dimension)*
~/source/shoes4:bin/shoes benchmark/dimension.rb
       user     system      total        real
1_000_000x margin_start  0.540000   0.010000   0.550000 (  0.473000)
1_000_000x element_extent  1.970000   0.020000   1.990000 (  1.885000)
```

So that's a pretty big drop. You might wonder whether it matters, but here's a snippet of profiling from running the manual with some call counts of interest:

```
 %total   %self       total        self    children                 calls  name
---------------------------------------------------------------------------------------------------------
                       1.51        0.02        1.50       3627811/3924939  Shoes::Dimension#calculate_relative
                      10.75        0.03       10.72         19184/3924939  BasicObject#__send__
                      44.89        0.21       44.68        277944/3924939  Shoes::Dimension#element_end
  7727%     34%       57.15        0.26       56.90               3924939  Shoes::Dimension#element_extent
                      46.71        0.59       46.12       7773908/7900855  Shoes::Dimension#extent
                      39.83        0.07       39.75           38728/42527  Shoes::ParentDimension#extent
                      16.49        7.14        9.35       3887697/4421200  Shoes::Dimension#margin_start
                      14.57        6.47        8.09       3887697/3969633  Shoes::Dimension#margin_end
                       0.24        0.24        0.00       3887697/4657600  Kernel.nil?
                       0.00        0.00        0.00          37242/184482  NilClass#nil?
```

And yes, you read that right--`Shoes::Dimension#extent` was called 8 million times! Obviously we should do some hard looking at our layout paths to see if we can reduce that algorithmically, but regardless what we do I think that `Dimension` is bound to always be in our hot path.

@PragTob, to your question about where the time is going, I've put up a [sample gist here](https://gist.github.com/jasonrclark/1f71e5162ba8fbc77af7) that takes one of the more meta methods (`margin_start) and step by step deconstructs it to direct access to see the impact of each bit of "meta" we're using.

* Fully meta-programmed (`full_meta`)
* Extracting the ivar name (`hoist_ivar_name`)
* Directly accessing the ivar instead of `instance_variable_get` (`direct_ivar`)
* Memoizing the relative flagging rather than checking repeatedly (`memoize_relative_flag`)
* `def` instead of `define_method`, with everything done directly (`full_direct`)

This uses the most excellent [benchmark-ips](https://github.com/evanphx/benchmark-ips) which I find much easier for these type of comparisons than plain `benchmark`. The results:

```
           full_meta    16.473k i/100ms
     hoist_ivar_name    17.984k i/100ms
         direct_ivar    28.890k i/100ms
memoize_relative_flag
                        33.238k i/100ms
         full_direct    36.262k i/100ms
-------------------------------------------------
            full_meta    325.594k (± 5.8%) i/s -      1.631M
      hoist_ivar_name    423.482k (± 3.3%) i/s -      2.122M
          direct_ivar    798.759k (± 3.2%) i/s -      3.987M
memoize_relative_flag      1.208M (± 3.0%) i/s -      6.049M
          full_direct      1.523M (± 3.9%) i/s -      7.615M

Comparison:
          full_direct:  1522752.7 i/s
memoize_relative_flag:  1207850.6 i/s - 1.26x slower
          direct_ivar:   798758.6 i/s - 1.91x slower
      hoist_ivar_name:   423482.2 i/s - 3.60x slower
            full_meta:   325593.6 i/s - 4.68x slower

```

So that's the basis for my proposed changes here. I'm hoping that we can cut things down in a big way algorithmically as well at some point, but this at least gives us the fastest basis (or close at least) in `Dimension` that I think we can hope for.

Shoes on!

PragTob:
Hi there! Awesome break down work! :D You should really get that into a blog post, if you won't, then I will! :)

That's some great performance work right there and I agree... I'm still looking forward very much to algorithmic changes on this e.g. don't do full contents_alignment all the time, that's one of my favorite tickets waiting right there.

Looking at this I remembered a project I was once one and was wondering why people used eval and def instead of the meta programming `define_method` - and stuff like this is probably why. I forked your gist [and added one variant](https://gist.github.com/PragTob/eb43aaa99e4b86b99d4e) where I do an eval and `def` on a string. it performs as well as `full_direct`. We could interpolate that string and with that keep the meta programming around. We don't have to, Metaprogramming is often evil and with that we lose code highlighting as well - just if the duplication bugs us again. Your call :)

Also please add `benachmark/ips` I love it myself - no idea why I didn't use it with the benchmarks before. Maybe I didn't know it back then.

Really, AMAZING work here :star: 

Shoes on!
Tobi

jasonrclark:
Thanks @PragTob, and interesting other stuff you turned up on JRuby there! Have been meaning to blog more (like everyone I think :stuck_out_tongue:), and this would be a good thing to write up.

Merging so I can run some things off master with both perf branches landed.

PR #1084, opened 2015-03-27
element_extent perf tuning

jasonrclark:
Another night, more perf tuning :notes: 

This time around, I did a `--profile` style run against the activity that happens when clicking the Colors link in the manual. I noted that `element_extent` showed up heavily in that list, with the majority of its time being spent in children. Example output:

```

total        self    children                 calls  name
------------------------------------------------------------------------------------
 1.51        0.03        1.48         833372/844616  Shoes::Dimension#element_extent
```

Looking at the method, most of it is delegating out to `margin_start` and `margin_end`. Since those margins change infrequently, it made sense to just memoize the combined value when those are set.

I also noted that we were calling `extent` an additional time unnecessarily. While typically pretty fast, a method call is still slower than no method call :+1: 

End results from benchmark update:
**Before**
```
      element_extent    19.775k i/100ms
-------------------------------------------------
      element_extent    405.070k (± 3.9%) i/s -      2.037M
```

**After**
```
      element_extent    35.642k i/100ms
-------------------------------------------------
      element_extent      1.021M (± 4.8%) i/s -      5.097M
```

More than 2x improvement in the microbenchmark!

The real exciting part of this, though, was when I tried it out in the manual. I updated the code there to spit out a simple timer around the [`color_page` method](https://github.com/shoes/shoes-manual/blob/master/lib/shoes/manual/app.rb#L188-L201).

**Before**
```
1.569000005722046
1.5139999389648438
1.553999900817871
```

**After**
```
0.6310000419616699
0.5889999866485596
0.5909998416900635
```

From a rough wall-clock timing of my own, it also sped up the painting. Overall a click on Colors went from ~4s to ~2s. Anecdotally, when I try out the manual on `pre3` it takes more than 10s for the page to load.

Not a bad start! :rocket:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,4 +27,5 @@ group :development do
   gem 'webmock'
   gem 'hometown'
   gem 'rubocop'
+  gem 'benchmark-ips'
 end
--- a/benchmark/dimension.rb
+++ b/benchmark/dimension.rb
@@ -1,21 +1,17 @@
-require 'benchmark'
+require 'benchmark/ips'
 
-Benchmark.bm do |benchmark|
+Benchmark.ips do |benchmark|
   dimension = Shoes::Dimension.new
   dimension.start  = 0
   dimension.extent = 10
   dimension.margin_start = 2
 
-  benchmark.report '1_000_000x margin_start' do
-    1_000_000.times do
-      dimension.margin_start
-    end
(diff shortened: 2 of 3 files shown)


PragTob:
:heart: :heart: :heart: :heart: :heart: :rocket:

PR #1081, opened 2015-03-25
Default class styles do not override app styles

seethemhigh:
A work in progress for #1077 . Here's how it looks now:
image

and the code for the above example:
```ruby
Shoes.app do
  stack do
    font = 'Impact'
    style(font: font)
    txt = 'The quick fox runs fast'
    title txt
    style(Shoes::Title, font: font)
    title txt
    para txt
    rect(200, 200, 100, 100)
  end
end
```
Tests are all red due to the removal of `:fill` from `Shoes::Common::Style::DEFAULT_STYLES`. More explanation in the commit notes.

PR #1085, opened 2015-03-28
Shoes::Common::Styles specs take into account the style overrides

seethemhigh:
Specs for PR #1081 . 
In order to write these, I reverted back to the older version of Shoes::Common::Styles to the point where there was no bugfix #1081, wrote the specs and cherry-picked them. 
I'm not sure how thorough they are, but I saw them failing and then passing when fixed, which is nice :)
Thanks!


PragTob:
Thanks a lot!

I added 3631f081f60341b47b1a to the specs, as in that case `let` statements are a bit more concise and generally more flexible and we use them all over the place :)

PR #1090, opened 2015-04-12
Perf Tuning: Only call Shoes::Swt::Common::Painter#set_rotate if we need it

jasonrclark:
There's a significant cost to generating the transforms, so skip if rotate is zero.

How did I come to this conclusion? Well, I'd noticed `samples/expert-game-of-life-adjusted.rb` ran slowly, and it seemed related to painting. It was hard to extract the timings I wanted from JRuby's `--profile`, so I built something to make it easier.

Introducing `newrelic-shoes`! (source: https://github.com/jasonrclark/newrelic-shoes, rubygems: https://rubygems.org/gems/newrelic-shoes)

Installing this gem, you can press Ctrl+Alt+N to start tracing an app (it outputs those instructions to STDOUT when required). On shutdown, metrics are written out for the specific methods being traced. The 0.1.0 release of the gem just watches painting methods, which I found hard to isolate otherwise.

When I was measuring, I followed these steps
* Start up `samples/expert-game-of-life-adjusted.rb`
* Press Ctrl+Alt+N once it was running
* Click "Glinder" (typo?)
* Once the screen painted the glider, quit

The results:

**master**
```
Call                                      Count  Total Time (s)  Excl. Time (s)  IPS
Shoes::Swt::RectPainter/paint_control     4800   4.871           0.415998        985
```

**set-rotate-optimize**
```
Call                                      Count  Total Time (s)  Excl. Time (s)  IPS
Shoes::Swt::RectPainter/paint_control     4800   4.240998        0.407998        1132
```

So this is an almost 13% improvement over the painting speed when `set_rotate` is called blindly for every rectangle (IPS aka "iterations per second" shows this especially nicely). Seems like a nice, small win.

Foreshadowing... At least one more PR of this variety around rectangle painting and another around text painting on their way when I can get them polished up and tested. By the time that's all there, the Game of Life and the manual will be running much more smoothly!


shoes-swt/lib/shoes/swt/common/painter.rb
```diff
@@ -66,6 +66,15 @@ def draw_setup(graphics_context)
...
         end
 
+        def needs_rotate?
+          dsl = @obj.dsl
+          supports_rotate?(dsl) && dsl.rotate != 0
+        end
+
+        def supports_rotate?(dsl)
+          dsl.is_a?(::Shoes::Oval) || dsl.is_a?(::Shoes::Rect)
+        end
```
PragTob: This should eventually move up to be an instance method or something else so we don't do class checks :)

shoes-swt/lib/shoes/swt/common/painter.rb
```diff
@@ -33,18 +33,18 @@ def paint_object(graphics_context)
...
+          if needs_rotate?
+            set_rotate graphics_context, obj.rotate,
+                       obj.element_left + obj.element_width / 2.0,
+                       obj.element_top + obj.element_height / 2.0 do
               fill graphics_context if fill_setup(graphics_context)
               draw graphics_context if draw_setup(graphics_context)
+            end
+          else
+            fill graphics_context if fill_setup(graphics_context)
+            draw graphics_context if draw_setup(graphics_context)
```
PragTob: Couldn't we just do the rotation in an if and always do the fill and draw? Like

```
rotate if needs_rotate?
draw_it # the fill and draw calls
```

?

PragTob:
Excellent work as usual, minor code style comments :+1: 

Also <3 <3 <3 what you did with newrelic_shoes

shoes-swt/lib/shoes/swt/common/painter.rb
jasonrclark: Yeah, let me look at flattening that out a bit more while I'm here. It takes a little state since we have to undo the transform afterward, but might be cleaner.

jasonrclark:
Updated for both your comments, @PragTob. Loved moving `needs_rotate?` to an instance methods.

A little less certain whether flattening out `set_rotate` was a win--did that entirely in 6afc8a8, so if you want to weigh in, I'm happy to either take that or remove it depending how we feel. While it avoids duplication, the need to undo the rotation work after the fill/draw makes it a bit more complicated than just calling to rotate when we need.

jasonrclark:
Build break here, removed method no longer used from production code, missed that a test was calling it. Fixed up and re-pushed for :green_heart:

shoes-swt/lib/shoes/swt/common/painter.rb
```diff
@@ -30,22 +30,17 @@ def paint_control(event)
...
-              fill graphics_context if fill_setup(graphics_context)
-              draw graphics_context if draw_setup(graphics_context)
-          end
+
+          rotation = rotate(graphics_context, dsl) if dsl.needs_rotate?
+
+          fill graphics_context if fill_setup(graphics_context)
+          draw graphics_context if draw_setup(graphics_context)
+        ensure
+          undo_rotate(graphics_context, dsl, rotation) if rotation
```
PragTob: not sure why we need this `undo_rotate` ? Why is it needed now and not before? I'm sure there is a reason, I just don't get it :)

But yeah in this case it's probably better to get with the solution you had before (maybe + pulling out the `fill`/`draw`)

PragTob:
:) Thanks, added a comment as I don't know why we need the `undo_rotate` - otherwise, looks good :)

shoes-swt/lib/shoes/swt/common/painter.rb
jasonrclark: So before we were calling `set_rotate`, which took a block and did work before and after it. `undo_rotate` in the new code is just what was after the `yield` in `set_rotate` before.

Reading over it again, I'm with you that it's probably be better to just extract the `fill` + `draw`, so unless you object will roll with that.

PragTob: :+1:

PR #1091, opened 2015-04-12
Benchmark ips usage

### LINKED ISSUE #1083: Use benchmark-ips gem at least for micro benchmarks
[benchmark-ips](https://github.com/evanphx/benchmark-ips) automatically runs the benchmark a certain number of times (with warmup) which is nice at least for micro benchmarks.

Find existing micro benchmarks in benchmarks and use benchmark-ips there instead please :)

PragTob:
This should fix #1083 

I didn't add it to all benchmarks as it didn't feel appropriate to me when, for instance, the benchmark should show the effect of a lot of elements in the app. But that might be me, or I'm holding it wrong :)

Oh, also commented old outcommented code back in :)

### DIFF
--- a/benchmark/app_clear.rb
+++ b/benchmark/app_clear.rb
@@ -1,19 +1,17 @@
-require 'benchmark'
+require 'benchmark/ips'
 
 Shoes.app do
   background green
 
   button 'start repeated app clear benchmark' do
-    Benchmark.bm do |benchmark|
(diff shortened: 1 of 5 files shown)


jasonrclark:
I agree with your choices of which to use IPS and which not here--there are certain types of tests that just don't quite make sense for it.

Thanks for getting those updated!

PR #1094, opened 2015-04-15
Call change blocks on programmatic list updates

### LINKED ISSUE #1089: ListBox#change block doesn't trigger when ListBox#choose is called
Example:

```
Shoes.app do
  button "change" do
    @list.choose("x")
  end
  @list = list_box items: ["x", "y", "z"] do |x|
    p x.text
  end
end
```

When you manually change the ListBox, the change block triggers, but when ListBox#choose is called by the button's click block, it doesn't. In Shoes 3, ListBox#choose does trigger the change block.

jasonrclark:
Fixes #1089

Calling `ListBox#choose` wasn't triggering the change blocks in the DSL
layer because SWT apparently only sends selection events for
user-initiated changes.

Fixed by simply calling the updates when we explicitly alter the list
choice.

Chose to do this at the SWT layer so that if future backends do
appropriately trigger changes on choose already, they don't have to work
around the DSL enforcing it.

### DIFF
--- a/shoes-swt/lib/shoes/swt/list_box.rb
+++ b/shoes-swt/lib/shoes/swt/list_box.rb
@@ -38,6 +38,7 @@ class Shoes
 
       def choose(item)
         @real.text = item
+        @dsl.call_change_listeners
       end
 
       def enabled(value)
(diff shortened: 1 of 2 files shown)


PragTob:
:+1:

PragTob:
btw. really good call on doing this on the SWT/backend layer! :)

PR #1096, opened 2015-04-19
Fix for star default fill

### LINKED ISSUE #1095: Star should have a default black fill
(this regression was likely introduced in #1081 at least it seems so 99159983cefad65689 is broken while 15121ad919373cb7cd is fine)

On current master simple-face.rb looks like this:

screenshot from 2015-04-18 15 43 11

while it is supposed to look like this:

screenshot from 2015-04-18 15 45 26

It seems like a simple mixup of the default fill colors - probably just one setting missing :)

Tobi

jasonrclark:
Just what it says on the tin.

Fixes #1095


PragTob:
Thanks, won't do a changelog entry since this is an "intran this version" fix :)

Issue #1097, opened 2015-04-19
Specs for `TextStyleFactory` and `TextSegment`


jasonrclark:
Interactions with styling aren't well spec'd, and in moving things around some coverage was removed in 406e925.

I don't necessarily want precisely those specs back, but would rather we test the classes appropriately for how we think they should be tested today.

Newcomer friendly since those classes are pretty isolated, so just backfilling a spec for it should be relatively accessible.

piisalie:
Is this still wanted? I did a little git digging on the TextSegment and TextSegmentCollection specs which look like they haven't been updated much since this issue. I didn't see any TextStyleFactory specs. (I could be looking in the wrong place / misunderstanding)

I'd be willing to give this a try if it still needs doing.

jasonrclark:
Yeah, this is definitely still desired. That area isn't well tested, so any help there could be appreciated.

`TextStyleFactory` is also a good candidate from what I remember for some more testing. The struggle there I think will be getting specs that aren't just transliterating the production code, but worth see what we can do.

Thanks!

PR #1098, opened 2015-04-19
Perf Tuning: Only apply styling to TextSegmentCollection once

jasonrclark:
Found another thing in my perf wanderings through Shoes... noticed that the colors page in the manual was super slow when I was scrolling, which seemed odd. Dug in and I discovered that we were creating a new `TextSegmentCollection` every time we painted (which was multiple times per `contents_alignment` potentially) and that was costing us a lot of time.

So now, the `Shoes::Swt::TextBlock` creates a collection directly itself (instead of making a new one each time), and the `Shoes::Swt::TextBlock::TextSegmentCollection` will only do it's styling on the internal SWT objects once.

Duplicating steps:
* Load manual with `newrelic-shoes` available
* Click on `Colors`. Wait for it to load.
* Start New Relic (Ctrl+Alt+N)
* Scroll to bottom of the page.

Results
**master**
```
Call                                                 Count  Total Time (s)  Excl. Time (s)  IPS
Shoes/Shoes::Swt::TextBlock::Painter/paintControl    1505   1.266002        0.153001        1189
```

**keep-text-segment-collection**
```
Call                                                 Count  Total Time (s)  Excl. Time (s)  IPS
Shoes/Shoes::Swt::TextBlock::Painter/paintControl    1505   0.345998        0.345998        4350
```

Note that this does remove a bunch of specs we had against the `TextBlock::Painter`. These all turned out to be testing various facets of the styling from several of the layers beneath, and had little to do with the actual painting. I've entered #1097 for getting that put back. I'm not worried about those bits breaking immediately, and the specs were  a bit messy anyway.

### DIFF
--- a/shoes-swt/lib/shoes/swt/text_block.rb
+++ b/shoes-swt/lib/shoes/swt/text_block.rb
@@ -13,10 +13,10 @@ class Shoes
       attr_accessor :segments
 
       def initialize(dsl, app)
-        @dsl            = dsl
-        @app            = app
-        @segments       = []
-        @painter        = Painter.new @dsl
+        @dsl      = dsl
+        @app      = app
+        @segments = TextSegmentCollection.new(@dsl, [], default_text_styles)
+        @painter  = Painter.new @dsl
         @app.add_paint_listener @painter
       end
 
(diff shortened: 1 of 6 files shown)


PragTob:
Seems good to me and a HUGE performance boost (also always noticed the slowness of scrolling there) - thanks a lot! :)

PR #1099, opened 2015-04-19
Gradient application fix

PragTob:
For me this fixes the display of stars with a gradient.

screenshot from 2015-04-19 21 38 38

Also fixes the inspect on internal app, on which I stumbled while investigating this.

Sadly, I can't figure out what needs to change around shape to make the gradient work :-/

### DIFF
--- a/shoes-swt/lib/shoes/swt/gradient.rb
+++ b/shoes-swt/lib/shoes/swt/gradient.rb
@@ -30,33 +30,26 @@ class Shoes
       end
 
       def apply_as_fill(gc, dsl)
-        left, top, width, height, angle = extract_dimensions(dsl)
-        pattern = create_pattern(left, top, width, height, angle)
+        pattern = create_pattern(dsl)
         gc.set_background_pattern pattern
       end
 
       def apply_as_stroke(gc, dsl)
-        left, top, width, height, angle = extract_dimensions(dsl)
-        pattern = create_pattern(left, top, width, height, angle)
+        pattern = create_pattern(dsl)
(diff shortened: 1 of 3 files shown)


shoes-swt/lib/shoes/swt/gradient.rb
```diff
@@ -44,13 +44,8 @@ def apply_as_stroke(gc, dsl)
...
       def extract_dimensions(dsl)
-        if dsl.is_a?(Star)
-          left = dsl.element_left - dsl.element_width / 2.0
-          top  = dsl.element_top - dsl.element_height / 2.0
-        else
-          left = dsl.element_left
-          top  = dsl.element_top
-        end
+        left = dsl.element_left
+        top  = dsl.element_top
```
jasonrclark: Interesting that this had that effect... don't feel like I understand why, but I haven't gotten familiar with the gradient code.

Noticed too that I think we could just unroll the `extract_dimensions` method now that it's basically just using the `dsl.element_*` accessors with no real additional computation. Saves an array allocation :wink:

PragTob: yeah I guess it was from back in the days when each thing had its own dimensions and the one of star were slightly weird... at least that's what I tell myself when I hope I didn't break any important behavior!

Also good catch about unrolling, will do!

PragTob:
Pushed update with unrolled version! :)

jasonrclark:
Nice! Pretty sure the Travis build was unrelated to us--failed bundling. Rerunning, and will happily merge once it's there.

Issue #929, opened 2014-10-27
Allow interactively selecting a Shoes backend


jasonrclark:
Implementation of #925 made the spot where we could select a specific backend, but only implements it for a single backend (which doesn't require user interaction). Let's handle the interactive selection (and potentially other means, like an ENV?) as a separate issue.

I'm imagining something command-line, but like you get for uninstalling gems with a numbered list that you can select from.

PragTob:
And this is... IN! :clap:

PR #1102, opened 2015-04-21
Allow for selecting between multiple backends

jasonrclark:
First step (though potentially not only?) toward #929.

So to help test things out, I made a fake Shoes backend at https://github.com/jasonrclark/shoes-faux which I then did testing against by including in my Gemfile locally for Shoes, removing my `shoes-backend` file, and seeing what happened.

This resulted in a few specific changes:
* First off, the straight-forward selection logic in `Shoes::UI::Picker` to ask for which backend to generate.
* Needed to not hard-code loading `shoes/swt` in `Shoes::UI::CLI`, so got away from that and enforced that we pass the backend explicitly when constructing the CLI.
* Then I noticed that things were very unhappy having two backend gems present in the same Gemfile. This turned out to be because just requiring `shoes/swt` was forcing it all to load. So I moved that into an explicit method, and new backends should follow the same practice--don't assume you're the backend just because you're required, wait for the `Shoes::Backend.initialize_backend` call.
* That loading broke the SWT tests which relied on loading the backend automatically. Now it loads it explicitly like a good citizen (from a spec_helper)
* Also sneaked in some changes so the `Shoes::Packager` is actually fine if we don't have a backend. `run` will fail on it, but it will not stop you from running an app against a backend just because it can't package yet. This eases a little of the on-ramp for stubbing out a new backend.

This *might* qualify as something fully-featured enough to land, but it's late and I want to consider it further before saying it's ready :sleeping:

### DIFF
--- a/shoes-core/lib/shoes/ui/picker.rb
+++ b/shoes-core/lib/shoes/ui/picker.rb
@@ -3,6 +3,11 @@ class Shoes
     # This class is used for interactively (if necessary) picking the Shoes
     # backend that the user will run their Shoes app with.
     class Picker
+      def initialize(input=STDIN, output=STDOUT)
+        @input  = input
+        @output = output
+      end
+
       def run
         bundle
         generator_file = select_generator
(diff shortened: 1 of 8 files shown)


shoes-core/lib/shoes/ui/picker.rb
```diff
@@ -21,15 +26,47 @@ def bundle
...
         else
-          fail NotImplementedError("Currently can't interactively select a backend. See #929")
+          output_candidates(candidates)
+          prompt_for_candidate(candidates)
         end
+      end
+
+      def output_candidates(candidates)
+        @output.puts
+        @output.puts "Select a Shoes backend to use (This is a one-time operation):"
```
PragTob: should probably mention that one is expected to enter the number :)

shoes-core/lib/shoes/ui/picker.rb
```diff
@@ -21,15 +26,47 @@ def bundle
...
+        candidate = nil
+
+        until candidate
+          entered_index = @input.readline.to_i
+          if entered_index > 0
+            if candidate = candidates[entered_index - 1]
+              break
+            end
+          end
+          @output.puts "Invalid selection. Try again with a number from 1 to #{candidates.size}."
```
PragTob: oh ok it has the hint here, that's probably fair enough :)

PragTob:
Looks great to me! :+1: 

Waiting for you to merge it, though. Especially nice IO testing through DI :+1: 

have a good night sleep!

shoes-core/lib/shoes/ui/picker.rb
jasonrclark: Might not be a bad idea, even with the message after a failure. I did pattern the styling after the `gem uninstall` commands, so it might be familiar to some, but no sense making it more cryptic than necessary.

jasonrclark:
On further reflection, updated the message and I do think this is ripe for a merge. Another PR will follow shortly with some additional pieces, but best to keep things small, so will merge on :green_apple: here.

PR #1103, opened 2015-04-22
--backend flag for shoes executable

jasonrclark:
This fills out the remainder of my thoughts for #929. It allows a flag on `shoes` to control backend selection, which will be especially useful for future development (though unnecessary for most users typical interactions with Shoes).

Two usages for the `--backend` flag.

To re-run the same selection logic as when we're missing `shoes-backend`:
```
bin/shoes --backend
```

To restrict (and try to select a single backend), use this:
```
bin/shoes --backend swt
```

Note that the second form, if it somehow matches multiple backends, will then prompt. However the search is set up to avoid that (you'd need a gem like `shoes-swt-fakeout` to interfere with `--backend swt` finding `shoes-swt`).

My env setting edits on Windows `shoes.bat` are untested as yet, but they do follow how we set env variables in other places in the batch files.

Lastly, slipped in a sorting on the output when prompting, as I noticed that the order was flip-flopping between runs which annoyed me.


PragTob:
Looks great, thanks :+1:

PR #1104, opened 2015-04-26
Rename Shape#arc to Shape#arc_to

### LINKED ISSUE #795: Support `Shoes::Shape#arc_to`
From my brief testing `arc_to` seems to be slightly different from `arc`--at least when I passed the same values on Shoes 3 to `arc` and `arc_to` I got different results... :confused: 

In any case, you can see reference to it in the online Shoes manual at http://shoesrb.com/manual/Art.html#shape.

Appears to be some conversation about it in #424 as well that might be worth reviewing.

jasonrclark:
Fixes #795.

So a funny story... turns out we already had a method that behaved the same as the `Shape#arc_to` back on Shoes 3. It just ended up being called `Shape#arc` instead.

For compatibility, I'm renaming it back, along with adding a sample and some tests (I'm betting the lack of samples and only having one reference to it in the manual is why it got mixed up to begin with). The method `arc` remains accessible, as you can use base Shoes methods from within the shape anyway--they just aren't related to the shape drawing.

Results:

`samples/shape-arc_to.rb` on Shoes 3:
image

`samples/shape-arc_to.rb` on Shoes 4:
image

### DIFF
--- a/samples/README
+++ b/samples/README
@@ -39,6 +39,8 @@ simple-timer.rb
 simple-tictactoe.rb
 simple-visibility.rb
 
+shape-arc_to.rb
+
 good-arc.rb
 good-clock.rb
 good-follow.rb
--- a/samples/shape-arc_to.rb
+++ b/samples/shape-arc_to.rb
@@ -1,0 +1,9 @@
+Shoes.app do
+  shape do
+    move_to(90, 55)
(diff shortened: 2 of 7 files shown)


PragTob:
Love it when it turns out to be so easy, thanks @jasonrclark ! :)

Issue #526, opened 2013-12-31
samples/simple-form.rb doesn't work properly (background image placement + Shoes.p)


wasnotrice:
- background doesn't position images properly
- `Shoes.p` doesn't do anything

![image](https://f.cloud.github.com/assets/908/1825768/95222da0-71d3-11e3-831a-8567dc1bebb2.png)

jasonrclark:
So I'm also seeing the `Shoes.p` on the console, so that's good.

On the offset, I started digging into this a little, and decided to monkey with the image that we're actually using on the right background to put a border around it in the image itself. I hoped this would give me some insight into what was happening, and it has.

So the new right image I was using looks like this:

![image](https://cloud.githubusercontent.com/assets/130504/5890409/b2260c38-a40d-11e4-8916-2fec6abaebb4.png)

But when I run the app, surprisingly I see this:

![image](https://cloud.githubusercontent.com/assets/130504/5890410/c67f93fc-a40d-11e4-9dc2-d7372d570739.png)

When I resize, it appears almost as if the background image is being tiled from the left edge of the screen, and the background is just showing us the slice of it where it's positioned. Don't really understand how that's getting painted yet, and probably won't tonight, but thought I'd jot this down in case someone who knows more about that understands why I'd be seeing that effect.

PragTob:
Yeah I also had the impression that this was happening... I guess the solution should/could be to just paint that background once if it gets positioned like this (it would be like a normal image in the background then)

PragTob:
On this again and I use my new patent method: "Change values until you find the value that changes what you like"tm

No, not really but it's the rect painter and it might simply be that we skip background positioning in the positioning code... investigating!

jasonrclark:
Made a PR against this @PragTob... hope you didn't have too much work in flight, and happy to recant on mine if you have better thoughts. What I put up isn't quite ready for merging, but I think it might be the best way to handle how SWT's pattern falls short of what we want to achieve here.

PragTob:
Was in the research phase, no code written yet :)

PR #1105, opened 2015-04-27
Replace Swt::Pattern with custom image drawing

jasonrclark:
All right, here's a beginning to addressing #526

Turns out that we get minimal control over the positioning when `Swt::Pattern` is used to draw images, so we have to take control ourselves to get the results we want.

Specifically, when an element applies the image pattern as a background, I expect it will start over with the images from the corner of that new element, not just show the images cut off wherever they end up positioning on a repeat from the window's edge.

Modifying the sample file `menu-gray.png` to give it a border, you can see the issue when you run it against master with this app:

```
Shoes.app do
  file = File.expand_path(File.join(__FILE__, "..", "..", "samples/menu-gray.png"))

  background file
  rect 120, 120, 200, 200, strokewidth: 5, fill: pattern(file)
end
```

**master**
image

**This PR**
image

Bonus shot of `samples/simple-form.rb` looking nice:
image

One change is that this pushes the return value for deciding whether to do further drawing down to the `Color`, `Gradient` and `ImagePattern` classes. This was necessary because our drawing in `ImagePattern` gets overwritten by `RectPainter#fill` if we don't short-circuit it. I left comments as the return value there was a bit subtle to rely on.

TODO:
* [x] Legit tests around drawing the images, especially all the edge cases and cut-offs
* [x] Break down the drawing method further
* [ ] Test against all the samples to catch any issues with tweaking contracts on fill/stroke for rect drawing

### DIFF
--- a/shoes-swt/lib/shoes/swt/color.rb
+++ b/shoes-swt/lib/shoes/swt/color.rb
@@ -29,6 +29,7 @@ class Shoes
       def apply_as_fill(gc, _dsl = nil)
         gc.set_background real
         gc.set_alpha alpha
+        true
       end
 
       # @param [Swt::Graphics::GC] gc the graphics context on which to apply stroke
@@ -36,6 +37,7 @@ class Shoes
       def apply_as_stroke(gc, _dsl = nil)
         gc.set_foreground real
         gc.set_alpha alpha
+        true
       end
     end
 
--- a/shoes-swt/lib/shoes/swt/common/fill.rb
+++ b/shoes-swt/lib/shoes/swt/common/fill.rb
@@ -34,8 +34,8 @@ class Shoes
 
         def apply_fill(context)
           if fill
+            # Return value determines whether to continue upstream.
             fill.apply_as_fill(context, self)
-            true
           end
         end
       end
--- a/shoes-swt/lib/shoes/swt/common/stroke.rb
+++ b/shoes-swt/lib/shoes/swt/common/stroke.rb
@@ -36,9 +36,10 @@ class Shoes
 
         def apply_stroke(context)
           if stroke
-            stroke.apply_as_stroke(context, self)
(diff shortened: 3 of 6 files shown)


jasonrclark:
So, made some tests and broke things down, then started testing samples and realized two things I'd missed in my consideration.

Image patterns can be used as strokes (ala `samples/sample24.rb`), and even fill background isn't always rectangular... you know, like, ovals. D'oh.

Left to wonder... is it worth the specialized path here to get the behavior we want for `background` and maybe `rect`, but leave pattern as it was for other drawing and strokes. Or do we just declare defeat and update `samples/simple-form.rb`? Where do we stand on that?

PragTob:
Hi @jasonrclark - thanks for tackling this.

This is a rather difficult issue :-/ Personally I'd hoped (and searched) that there would be some way to tell SWT to create a pattern or image with an offset. Or even the GC when we set the pattern. Haven't found one so far, at least it doesn't seem to be a constructor.

I'm not sure where I stand overall - doing custom things means more maintenance for us. For simple-form one could argue it's a better use case for an image either way but well... just using the custom code in some scenarios seems as fine as declaring defeat (for now). Defeat seems like less work, I might skim through the docs again maybe there is a magic setter. This also seems to be hard to google.

Tobi

jasonrclark:
Yeah, I had the same hope of an offset, but there seems to be little available on the SWT Pattern (docs or runtime inspection of the objects) of use.

Not attached at all to this code, so if we don't find the easy setter, you want to just dump this, change the sample and move on?

jasonrclark:
More that I've thought about this, the more I don't think it's a good idea. Closing this and pitching #1108 instead.

PragTob:
:) yeah was leaning a bit more towards that solution myself but wanted you to have the call on that :)

PR #1107, opened 2015-04-28
Modify #replace to use options hash and parse fonts.

### LINKED ISSUE #523: simple-anim-text.rb doesn't work properly
Problems:

- [x] the slot should be centered on the window, but it's at upper left
- [x] the motion block doesn't print anything (uses Shoes.p - does not print anything for @PragTob on shoes3 either)
- [ ] font is not parsed/displayed correctly

shoes3:

screenshot from 2014-02-20 20 59 20

shoes4:

screenshot from 2014-02-20 20 59 55

bookstein:
Fixes #523 

Only the font parsing (and style setting) was still broken in the sample `simple-anim-text`. This modifies `TextBlock#replace` to look for an options hash in params and parse the font.

### DIFF
--- a/shoes-core/lib/shoes/text_block.rb
+++ b/shoes-core/lib/shoes/text_block.rb
@@ -34,6 +34,10 @@ class Shoes
     end
 
     def replace(*texts)
+      opts = texts.pop if texts.last.is_a?(Hash)
+      style(opts)
+      handle_styles(opts || {}) # handle fonts, after any other merge
+
       # Order here matters as well--backend#replace shouldn't rely on DSL state
       # but the texts that it's passed if it needs information at this point.
       @gui.replace(*texts)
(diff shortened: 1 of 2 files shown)


jasonrclark:
Build break looks Travis-related. Restarted it so hopefully that'll get to :green_apple:

PragTob:
Thank you very much for your contribution, yay! :+1: :dancer: :star: :star2: 

I made some small follow up adjustments, e.g. wrapping the styling in an `if opts` and adding the sample to samples/README (there is a list of all samples we consider working that are run automatically by a rake task - and thanks to you it now has one more sample :) :) :) )

Cheers and thanks again!
Tobi

PS: the email you used seems not to be connected to your github account, if you want the commit linked to your account you might want to add it to your account :)

PR #1108, opened 2015-04-29
Use images for samples/simple-form.rb

jasonrclark:
Alternate way for us to handle #526--just realize backgrounds are going to repeat differently and change the sample to work as expected with that.

This also adds samples demonstrating how the patterns repeat for fills and strokes, since that wasn't obvious from `samples/simple-form.rb`, and actually took me a while to realize when digging into this.

Supercedes #1105.


PragTob:
Seems great to me! :)

PR #1109, opened 2015-04-29
Fix for zombie apps remaining in Shoes.apps list

### LINKED ISSUE #1065: Closing leaves a zombie app (Shoes.apps)?
I believe I've spotted something suspicious when closing an app. Here's the output from IRB:
```
irb(main):001:0> require 'shoes'
=> true
irb(main):002:0> Shoes.apps
=> []
irb(main):003:0> Shoes.app {}
=> (Shoes::App:0x00003b24 "Shoes 4")
irb(main):004:0> Shoes.apps
=> [(Shoes::App:0x00003b24 "Shoes 4")]
irb(main):005:0> Shoes.apps.first.close
=> nil
irb(main):006:0> Shoes.apps
=> []
irb(main):007:0> 

```
Line 4 shows that there's a registered app, though I've already closed it using the mouse. But only after the line 5 the app goes away completely. Btw ```Shoes.app { quit }``` also leaves the app in ```Shoes.apps```.

It's a bug, right?

jasonrclark:
Fixes #1065

When a window was closed, we were properly unregistering it from the `Shoes::Swt.apps` list, but Shoes keeps a list of the DSL-side apps as well over in `Shoes.apps` that needed to get notified.

IRB works relatively well as in the bug report, although you need some extra business to get Shoes started up these days (a call to `Shoes::Swt.initialize_backend`). To that end, I found this test app useful:

```
Shoes.app do
  button "Apps?" do
    alert Shoes.apps
  end

  click do
    Shoes.app do
      para "Yo"
    end
  end
end
```

Click anywhere to make a new window, click the Apps button to see a list. The list should always match the open window count.

### DIFF
--- a/shoes-swt/lib/shoes/swt/app.rb
+++ b/shoes-swt/lib/shoes/swt/app.rb
@@ -235,6 +235,7 @@ class Shoes
       def unregister_app
         proc do |_event|
           ::Shoes::Swt.unregister(self)
+          ::Shoes.unregister(self.dsl.app)
         end
       end
 
--- a/shoes-swt/spec/shoes/swt/app_spec.rb
+++ b/shoes-swt/spec/shoes/swt/app_spec.rb
@@ -2,11 +2,12 @@ require "shoes/swt/spec_helper"
 
 describe Shoes::Swt::App do
   let(:opts) { {:background => Shoes::COLORS[:salmon], :resizable => true} }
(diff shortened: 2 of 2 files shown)


PragTob:
And thanks a lot, like so often these days, goes to @jasonrclark keep up the great work! :+1: :shipit:

Issue #945, opened 2014-11-14
Getting rid of rubocop offenses


PragTob:
[Rubocop](https://github.com/bbatsov/rubocop) is a tool that automatically checks for ruby style violations. We ran an autofix but still got a couple of them in here. 

If you want, you can help by getting rid of a couple of them, which would be appreciated :+1: 

Here is a nice website that shows an overview: http://www.refactorcop.com/shoes/shoes4

In case we don't like some of them, we might also deactivate them so rubocop doesn't scan for them any more.

Here is a current list of the offenses we have, generated through `rubocop lib/ --format offenses -D --out rubocop_offenses.txt`:

```
177  Style/Documentation
172  Metrics/LineLength
51   Metrics/AbcSize
46   Metrics/MethodLength
23   Lint/AmbiguousOperator
20   Metrics/ParameterLists
20   Style/CaseIndentation
17   Style/GuardClause
13   Style/TrivialAccessors
12   Lint/UselessAssignment
10   Metrics/ClassLength
9    Style/AccessorMethodName
9    Style/Semicolon
7    Metrics/CyclomaticComplexity
6    Style/PredicateName
5    Style/FormatString
4    Style/GlobalVars
4    Style/IfUnlessModifier
3    Metrics/PerceivedComplexity
3    Style/EvenOdd
3    Style/RegexpLiteral
2    Lint/Eval
2    Lint/RequireParentheses
2    Lint/ShadowingOuterLocalVariable
2    Lint/UselessAccessModifier
2    Style/NestedTernaryOperator
2    Style/SelfAssignment
1    Lint/RescueException
1    Lint/UnderscorePrefixedVariableName
1    Style/AsciiComments
1    Style/Blocks
1    Style/CommentAnnotation
1    Style/ConstantName
1    Style/EndOfLine
1    Style/FileName
1    Style/MultilineBlockChain
1    Style/OpMethod
1    Style/SingleLineBlockParams
1    Style/UnlessElse
--
638  Total
```

And here is a gist with a detailed analysis of what went wrong generated through `rubocop lib/ -D --out rubocop.txt` [rubocop gist](https://gist.github.com/PragTob/756ad5684c8dc47e5f2c)

Some time I'll probably go and check out the linting ones and others.

ybur-yug:
@Zanadar Have you got a PR youre still going to submit or have you stopped working on it? I was planning on hopping onto it myself.

PragTob:
[...]

PR #1112, opened 2015-05-02
Some Rubocop fixes

jaredmoody:
For #945.

If someone could double check bb44521, I'm not sure about changing the symbol literal in `shoes-swt/lib/shoes/swt/tooling/leak_hunter.rb`

Thanks!


jasonrclark:
Looks great, and that symbol change is totally legit. Only reason for using quotes around a symbol is if it includes things like a `.` character, and this one didn't need it.

PR #1113, opened 2015-05-03
MOAR Rubocop

jaredmoody:
For #945 

Realized I wasn't running on only `lib/` directories, but it's probably good that the samples have good style too?


jasonrclark:
Agreed that we may as well be running it on as much code as possible. Looks good here :rocket:

PR #1116, opened 2015-05-04
Prep 4.0.0.pre4

jasonrclark:
Since we're down to just a few more things, thought I'd post a version bump and initial CHANGELOG draft.

If we land anything else, will obviously need to append to CHANGELOG.


CHANGELOG
```diff
@@ -1,3 +1,48 @@
+4.0.0.pre4 (153 commits)
+-------------------------------
+
```
PragTob: I'd add a short summary here to give a broad overview :) I think the biggest story this time was performance, mostly thanks to @jasonrclark ! :)

jasonrclark:
Pushed out with a little intro/overview.

Last question, do we want to bump the furoshiki dependency to 0.3.1 before we cut things? I think so but the `~>` might be sufficient since it's just picking up bug fixes, nothing structural depending on it.

In light of that, proposed steps for getting `pre4` out the door:

* [ ] Run samples as final check for show-stoppers
* [ ] Release furoshiki 0.3.1
* [ ] Merge this PR
* [ ] Release the shoes! :rocket: :athletic_shoe: :high_heel: :mans_shoe:

PragTob:
:) I'm in favor of updating the furoshiki dependency. As you say it should not be needed due to `~>`, imo it communicates that we expect people to use that newer version - otherwise packaging will break :)

jasonrclark:
Found two regressions, noted above.

Both look straight-forward. I have a question out on approach with #1121, but I will hopefully get to them in the next day or so. Maybe a weekend release if all that comes together nicely?

jasonrclark:
Build break is expected based on bumping furoshiki version to the next rev that isn't actually out as a gem yet. Will get the tests running green before final merge on this.

KCErb:
Yeah! :dancer: :tada:

PR #1118, opened 2015-05-06
I love the smell of Rubocop fixes in the morning.

jaredmoody:
Low hanging fruit - style autocorrect fixes for #945

### DIFF
--- a/samples/expert-irb-adjusted.rb
+++ b/samples/expert-irb-adjusted.rb
@@ -79,16 +79,16 @@ Shoes.app do
     when "\n"
       begin
         out, obj = IRBalike.run(@cmd + ';')
-        @str += ["#@cmd\n",
+        @str += ["#{@cmd}\n",
           fg("#{out}=> #{obj.inspect}\n", "#fda"),
           "#{CURSOR} "]
         @cmd = ""
       rescue MimickIRB::Empty
       rescue MimickIRB::Continue
-        @str += ["#@cmd\n.. "]
+        @str += ["#{@cmd}\n.. "]
         @cmd = ""
       rescue Object => e
-        @str += ["#@cmd\n", fg("#{e.class}: #{e.message}\n", "#fcf"),
+        @str += ["#{@cmd}\n", fg("#{e.class}: #{e.message}\n", "#fcf"),
           "#{CURSOR} "]
         @cmd = ""
       end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -434,12 +434,6 @@ Style/SpaceAfterComma:
 Style/SpaceAfterControlKeyword:
   Enabled: false
 
-# Offense count: 30
-# Cop supports --auto-correct.
-# Configuration parameters: EnforcedStyle, SupportedStyles.
-Style/SpaceAroundEqualsInParameterDefault:
-  Enabled: false
-
 # Offense count: 385
 # Cop supports --auto-correct.
 # Configuration parameters: MultiSpaceAllowedForOperators.
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -460,11 +460,6 @@ Style/SpaceInsideHashLiteralBraces:
 
 # Offense count: 4
 # Cop supports --auto-correct.
-Style/SpecialGlobalVars:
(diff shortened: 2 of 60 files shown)


shoes-core/lib/shoes/oval.rb
```diff
@@ -10,9 +10,9 @@ class Oval
     STYLES = { fill: Shoes::COLORS[:black] }
 
     def create_dimensions(left, top, width, height)
-      left   ||= @style[:left]
-      top    ||= @style[:top]
-      width  ||= @style[:diameter] || @style[:width] || (@style[:radius] || 0) * 2
+      left ||= @style[:left]
+      top ||= @style[:top]
+      width ||= @style[:diameter] || @style[:width] || (@style[:radius] || 0) * 2
```
PragTob: hm, I do disagree with this - might be my personal style (and that my IDE makes those for me) but I like lining op `=` operators just as I like lining up `=>` in hashes, tremendously improves readability imo :)

shoes-core/spec/shoes/animation_spec.rb
```diff
@@ -60,7 +60,7 @@
   it { is_expected.not_to be_stopped }
 
   describe "with framerate" do
-    let(:opts) { {:framerate => 36, :app => app} }
+    let(:opts) { { :framerate => 36, :app => app } }
```
PragTob: not sure about the spacious usage of whitespace in hash brackets either - I remember we talked about this, does someone know what we decided? @jasonrclark

shoes-package/spec/spec_helper.rb
```diff
@@ -12,7 +12,7 @@
 require 'pathname'
 require 'rspec/its'
 
-Dir["#{SHOES_PACKAGE_SPEC_ROOT}/support/**/*.rb"].each {|f| require f}
+Dir["#{SHOES_PACKAGE_SPEC_ROOT}/support/**/*.rb"].each { |f| require f }
```
PragTob: also not too certain about the spaces here... but maybe it is indeed better and I should adapt :)

PragTob:
Most of this looks good :+1: 

Personally I like aligning operators such as `=` for multiple assignmetns or something like that, much more readable.

The other things are about spaces around braces, where I'm not entirely sure.

Tobi

shoes-core/lib/shoes/oval.rb
jaredmoody: Thanks for pointing this out, I missed this and I agree with you. Looks like there's a configuration option called `MultiSpaceAllowedForOperators` that will allow lining these up. I'll drop this commit from the PR and make a new one for this cop.

shoes-package/spec/spec_helper.rb
jaredmoody: I also prefer no space before my block parameters. Looks like there's a config option for this too, `SpaceBeforeBlockParameters`. I'll drop this commit and do a new PR for this too.

jaredmoody:
Added some notes above, I'll move the questionable commits into their own PRs

shoes-core/lib/shoes/oval.rb
PragTob: :+1:  sounds good

jaredmoody:
Rebased against master and bumped some method length/complexity limits from new code in master :P

jaredmoody:
Rebased against master after `pre4` merge.

shoes-core/spec/shoes/animation_spec.rb
jasonrclark: Don't recall a specific decision, but I can go either way on it so long as it's consistent between begin and end, which this does.

samples/expert-irb-adjusted.rb
```diff
@@ -79,16 +79,16 @@ def run(str)
     when "\n"
       begin
         out, obj = IRBalike.run(@cmd + ';')
-        @str += ["#@cmd\n",
+        @str += ["#{@cmd}\n",
```
jasonrclark: Didn't even realize this was allowed with strings! Glad to see it going away :grin:

jasonrclark:
Reading through this, I'm happy with the hash changes as they stand so I'll take it.

PR #1119, opened 2015-05-06
Rubocop: Disable Style/WordArray

jaredmoody:
This one seems questionable to me - I think the array notation is sometimes clearer.

Merge away if you prefer `%w` though.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -29,3 +29,7 @@ Style/TrailingComma:
 # methods)
 Style/TrivialAccessors:
   Enabled: false
+
+# We don't care if you use %w or literal string arrays
+Style/WordArray:
+  Enabled: false
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -495,9 +495,3 @@ Style/SymbolProc:
 # Cop supports --auto-correct.
 Style/VariableInterpolation:
   Enabled: false
-
-# Offense count: 13
-# Cop supports --auto-correct.
-# Configuration parameters: WordRegex.
-Style/WordArray:
-  MinSize: 3


PragTob:
Good call on it being somewhat questionable :)

Personally I used to not use `%w` and `%i` but recently started to do so - I still find the normal Array a bit clearer but the `%`s are nicer to type :)

So I want to confer with @jasonrclark - what's your opinion? I'd say I'm in slight favor of `%w`.

jasonrclark:
Personal preference, I tend toward the literal strings.

I might be projecting, but the `%` style also seems more cryptic to me for beginners. The literal style looks plainly like typical use of strings and arrays, and every time I find myself explaining an alternate-slightly-shorter syntax to a newcomer, it leaves me :grimacing:.

KCErb:
I'm on the fence about `%w`. I for one used to avoid it, and like you Jason cringe at these kinds of things when teaching a newcomer.

On the flip-side I've been using `guard` with `rubocop` religiously on a recent project and it's gotten me into the habit of using `%` *and* I feel like it's helped me out slightly with maintainability when working with lots of arrays (2 quotes and a comma is 3 keystrokes per array item, that adds up pretty fast).

Can we pick and choose? For example can we ignore this rule for the samples (where newcomers will be spending time) but leave it in for the code-base? Or can we say "ignore if array length is < number"?

If we turn off this cop do we want to discourage the use of `%w` so that the code-base is consistently friendlier to newcomers?

PS school is almost out! Shoes here I come! :tada:

jaredmoody:
I agree that `%w` is more cryptic for new rubyists.

You can pick and choose by adding comments in the code to tell rubocop to ignore those cases, but I think that's messy. I don't think there's a simple way to do it by directory.

There is a `MinSize` configuration, so we could use that to say word arrays with size >= x should use `%w`

I'd probably either choose the latter with a `MinSize` of perhaps 5 or turn off this check and leave it to judgement so we can use `[]` notation in the samples directory and `%w` anywhere else that's appropriate.

PragTob:
Min size is a cool config and I'd also go for 5 there.

That said, agree with jason - let's turn this cop off and say we are ok with both for now :)

KCErb:
:+1:

jaredmoody:
Updated the PR to disable this cop.

jasonrclark:
Thanks for the update on that @jaredmoody!

I will get to merging your other rubocop PR, but since I'd tested through all the samples for the `pre4` release wanted to get that cut before making any further changes which would drive me to re-test. Really appreciate all your work on tidying things up! :sparkles:

jaredmoody:
:+1:

Issue #559, opened 2014-01-05
Weird issue with TextBlocks, slots and alerts (CRASH)


PragTob:
Look at this little script:

``` ruby
Shoes.app do
  every do # could be any event like click/button
    stack do para 'Foo' end
    para 'hahahahaha destroy everything'
    alert 'muhhh'
  end
end
```

it results in this: ![screenshot from 2014-01-05 13 19 10](https://f.cloud.github.com/assets/606517/1846556/b86682a0-7603-11e3-997a-46c67dbcb675.png)

Through bisecting I already got the most likely culprit which is: a08657244e6c6a0

It might be (well it is somehow likely) that this issue is somehow related to `samples/expert-minesweeper-adjusted.rb` not running on #503

PR #1124, opened 2015-05-11
Position element as they are added

### LINKED ISSUE #590: Text has no width/height set after creation
Right now after we create a paragraph/TextBlock the width is `nil`.

That is due to the fact that to tell exactly how wide it's gonna be we need other layout information, that is not available yet. Which gets us back on the road of assigning those immedeatly - which is a hard one.

Why would I want to have the width of a text straight at my hands? Well so that I could reposition it, e.g. probably to be centered on another element (or something).

@jasonrclark do you have an idea how we could get this going?

edit: same goes for height

### LINKED ISSUE #593: Assign elements their position right after adding them and update as needed

jasonrclark:
Fixes #593! Also fixes #590!!!

Decided to take a swipe at positioning elements as we go, and it turned out to not be so bad. Much of the positioning mechanism was already in place from `contents_alignment`--we just needed to invoke it at different times.

The basic approach is that a `Slot` will now track a persistent `@current_position`, so when an element is added, it can be positioned based on that. `contents_alignment` interacts with that same position, so any later element additions after alignment pick up where it left off.

The implementation is greatly simplified by giving up in a couple of (to my perception uncommon) cases.

First punt is prepending--then we'll just do the `contents_alignment` instead. If you did this a ton, you'd have a bad time perf-wise, but the form of the method isn't very shoes-y anyway, and I don't see lots of samples with prepending, so I'm good with it.

Secondly, if you change a slot in the middle of a list of other slots, we don't get smart, we just `contents_alignment` on the parent to sort it out. If this proves to happen a lot and cause performance problems we can try to be smarter, but for now was more comfortable with keeping that path simpler.

Only found one small perf issue raised by the changes. Game of life was slow to start because of the large number of `rect` elements, all of which claimed to need positioning. Turns out, they didn't, but the art elements were wrongly reporting that they took up space. A quick change in `AbsoluteDimensions` and we were back to good.

Did require quite a bit of spec futzing to get things right (although the actual changes in the end were pretty small). Specifically I saw issues when things weren't quite set up right that would cause us to raise swallowed exceptions in positioning, which while not crashing the test immediately, did often result in failures as expectations weren't met. If you see any of these type of swallowed errors on the terminal with an app, please let me know through an issue!

Shoes on! :panda_face: :fire: :money_with_wings:

### DIFF
--- a/samples/README
+++ b/samples/README
@@ -4,6 +4,7 @@
 # NOTE: Comment out anything that isn't a filename, so we can execute this list.
 #       See the `rake samples` task
 
+simple-alert.rb
 simple-anim-shapes.rb
 simple-animate.rb
 simple-anim-text.rb
@@ -34,6 +35,7 @@ simple-keypress.rb
 simple-manual.rb
 simple-mouse-follow.rb
 simple-oval.rb
+simple-position-as-we-go.rb
 simple-sample-executor.rb
 simple-shoes-intro.rb
 simple-slide.rb
--- a/samples/simple-alert.rb
+++ b/samples/simple-alert.rb
@@ -1,0 +1,9 @@
+Shoes.app do
+  para "Click somewhere"
+  click do
+    # Earlier versions had trouble because para would get drawn while the
+    # alert was up, but before it got positioned. Make sure we don't regress.
+    para 'hahahahaha destroy everything'
+    alert 'muhhh'
+  end
+end
--- a/samples/simple-position-as-we-go.rb
+++ b/samples/simple-position-as-we-go.rb
@@ -1,0 +1,9 @@
+Shoes.app do
+  stack do
+    @p = para "Things get placed"
+    @b = button "as we go"
+
+    para "#{@p.element_left}, #{@p.element_top} -> #{@p.width}, #{@p.height}"
+    para "#{@b.element_left}, #{@b.element_top} -> #{@b.width}, #{@b.height}"
+  end
+end
--- a/shoes-core/lib/shoes/common/initialization.rb
+++ b/shoes-core/lib/shoes/common/initialization.rb
@@ -16,13 +16,22 @@ class Shoes
         create_dimensions(*args)
         update_dimensions if styles_with_dimensions?
 
-        @parent.add_child self
-        @gui = Shoes.backend_for self
+        create_backend
+        add_to_parent(*args)
 
         handle_block(blk)
         update_visibility
 
         after_initialize(*args)
+      end
+
+
+      # This method will get called with the incoming styles hash and the
+      # other arguments passed to initialize.
+      #
+      # It is intended for performing any additions to the styles hash before
+      # that gets sent on to style_init.
+      def before_initialize(_styles, *_)
       end
 
       # Set the dimensions for the element. Defaults to using the Dimensions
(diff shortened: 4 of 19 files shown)


PragTob:
Hey, thanks for tackling this! :)

Looks good from what I can tell, the real test comes when we remove the top slot contents_alignment call from the redrawing aspect, though :)

```ruby
       before_every NEED_CONTENTS_ALIGNMENT do |*_args|
          app.dsl.top_slot.contents_alignment
        end
```

And all other contents_alignment calls (except for resize) because they shouldn't be necessary anymore and therefore give us the performance++ - might also show other edge cases that we haven't thought of yet.

Thanks for all your hard work, got to work and feel a bit better now so i could muster the time for a review :)

jasonrclark:
Great point @PragTob! Will definitely examine the code for where we're making other `contents_alignment` calls and see whether I can get rid of them.

That case you called out in `RedrawingAspect` is interesting, because it only actually triggers from bringing up dialogs https://github.com/shoes/shoes4/blob/54817deb10f8900d5c47270b2e515409f41b38b8/shoes-swt/lib/shoes/swt/redrawing_aspect.rb#L39-L43, and has this comment:

```
      # These methods trigger SWT painting listeners, so we need to be sure
      # that positioning has run properly before allowing them to continue.
```

Will dig deeper when I get a chance later on to see if I can find the case this was added to address and make sure it still works.

Past removing those other calls to `contents_alignment`, will do a big samples pass-through and make sure things are behaving. Once that's done, clear for a merge? :green_book: :arrow_right: :boom: (Glad you're feeling a bit better too, and hope that continues!)

PragTob:
Ah sorry I kind of left that in a hurry, I thought there were more places... I believe there is at least one top level place that triggers contents_alignment.

So I believe we should be able to remove contents_alignment from Slot#positioning then as well app#open

Also from the layout

Basically, everywhere but the Slot class itself I guess... and maybe resizing.

And yes, ready for merge then :)

jasonrclark:
Ah yes, I actually added that `RedrawingAspect` bit :grin: #559 was the PR, and the sample from that still passes without the redraw hooks for `contents_alignment` here (as I'd expect.) Woot!

May have some bits of time across the afternoon and evening today and tomorrow, so will pick away at the various locations and then run through samples once they're locked down. Shoes on!

jasonrclark:
Build break was because this was against an outdated master which fouled some late furoshiki stuff I did before `pre4`. Rebasing and pushing again. Think this covers the necessary changes, will test samples tonight (hopefully) and merge if all is successful.

jasonrclark:
Samples looked good, and added two that exercise the positioning at addition as well. Mergin'

KCErb:
Woohoo!! :tada:

PragTob:
OMG THIS IS AMAZING!!!!!!!!!!!!!

Shoes on! And big thanks and hat tip to @jasonrclark :tada: :star:

PR #1125, opened 2015-05-11
Rubocop: Fix all Style/SpaceInsideBlockBraces

jaredmoody:
With configuration for no space before block parameters

This one's questionable, so let's discuss.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,6 +18,10 @@ Style/CollectionMethods:
   PreferredMethods:
     inject: 'inject'
 
+# We don't have a strong opinion on how you use spaces inside braces
+Style/SpaceInsideHashLiteralBraces:
+  Enabled: false
+
 # We like to be flexible if we want to add in interpolation
 # see discussion #859
 Style/StringLiterals:
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -383,9 +383,3 @@ Style/SpaceAroundOperators:
 # Configuration parameters: EnforcedStyle, SupportedStyles, EnforcedStyleForEmptyBraces, SpaceBeforeBlockParameters.
 Style/SpaceInsideBlockBraces:
   Enabled: false
-
-# Offense count: 200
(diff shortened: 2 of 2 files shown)


jasonrclark:
Hmm, looks like this got a merge conflict somewhere along the way. Can you rebase or merge to see what's up with it?

Also FYI, I'm out a lot of this weekend, but next thing I'll do when I'm back online is review and get your clean-ups merged now that `pre4` is out. Thanks again for tackling so many of those!

jaredmoody:
Rebased against master. 

No problem :thumbsup: Still got a lot more to go :)

jasonrclark:
Hmm, on reading over this more carefully, I kind of question this one. To me it feels inconsistent that it's fine leaving the space at the end, but removes it at the beginning just when there's a list of parameters. Doesn't help that I always put the spaces in my coding :)

@PragTob @KCErb @wasnotrice anyone else have an opinion? I'd be in favor of just punting this rule personally.

jasonrclark:
A thought I had as well related to Rubocop--if we hit a point where we're green on master for the existing rules, how do you feel about integrating it into Travis?

I noticed that a number of my parallel changes conflicts with stuff that @jaredmoody already did or was doing at the same time. Running this sooner rather than later will stave off more churn, as I'll start running rubocop (at least on the enabled rules) with my changes. What do you think?

KCErb:
Hmm, I don't feel like I'm entrenched either way yet. I used to skip the space before the block parameters but I've been using Rubocop on another project with guard so I've gotten used to adding it.

So on that note I'm in favor of not only adding Rubocop to Travis but to guard as well.

PragTob:
Seems like we have no clear opinion so we'd probably deactivate this rule :)

jasonrclark:
Yeah, I'd agree with just switching this one off. That cool @jaredmoody?

jaredmoody:
Sure, I'll update the PR

PragTob:
:+1:

PR #1126, opened 2015-05-12
Correct cursor height for empty text layouts

### LINKED ISSUE #1114: Too small cursor in tweaked simple-editor.rb
In testing out something else, noticed that changing the font size in `samples/simple-editor.rb` yielded a cursor not sized to the current text.

Change the line:

```
    t = para "", font: "Monospace 12px", stroke: white
```

to:
```
    t = para "", font: "Monospace 42px", stroke: white
```

For me, the app then looks like this:
image

jasonrclark:
Fixes #1114.

If a text block with a cursor starts out life empty, SWT returns the wrong height for that block. To get around this, we spin up a new segment (which won't be displayed) to get an appropriate height.

Minimal sample app (although `samples/simple-editor.rb` is pretty small too):

```
Shoes.app do
  p = para "", font: "Monospace 48px"
  p.cursor = -1

  keypress do
    p.replace(p.text + ".")
  end
end
```

Didn't spec this because the specs in this neighborhood all stub out the layout to a point where tests would be pretty much testing that we mocked things right. :grimacing:

### DIFF
--- a/shoes-swt/lib/shoes/swt/text_block/text_segment_collection.rb
+++ b/shoes-swt/lib/shoes/swt/text_block/text_segment_collection.rb
@@ -164,9 +164,20 @@ class Shoes
         end
 
         # This could be smarter, basing height on the actual line the cursor's
-        # in. For now, just use the first line's height.
+        # in. For now, just use the first line's (purported) height.
         def cursor_height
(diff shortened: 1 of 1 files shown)


PragTob:
Ahaha... great bug :D

PR #1128, opened 2015-05-12
Rubocop: :hash: :rocket: :skull: :fire:

jaredmoody:
This one's probably controversial?

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -227,12 +227,6 @@ Style/GlobalVars:
 Style/GuardClause:
   Enabled: false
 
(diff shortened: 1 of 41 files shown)


jasonrclark:
So actually we'd agreed on new-skool hash syntax at some earlier point. Still paused until `pre4`, but I'm good with this once we're released.

jasonrclark:
Did a quick rebase-and-merge on this, thinking it would close the PR out... turns out it didn't. Ah well.

Thanks for getting this in. Maybe I'll finally start using new hash syntax consistently :)

PR #1129, opened 2015-05-14
Gif class + builtin exit method

ducki13:
Gif class


jasonrclark:
This is so cool! Having helped out writing a presentation app for Shoes (https://github.com/pragtob/wingtips), my life will be greatly improved by the presence of animated gifs. :grinning: 

Got a few high-level questions for now, and will have more detailed code comments next week (I'm out much of this weekend)

* From an end-user perspective, I'm not sure the distinction between `gif` for animated files and `image` (for non-animated .gif files) is as simple as it could be. Could we detect that we're looking at an animated gif and branch within the existing `image` call somewhere? Maybe we could substitute a subclass of `Shoes::Swt::Image` if we see we're animated?
* I noticed you'd added the file for tests against the DSL but hadn't gotten any of them written. Do you need any guidance on how/what to test there? Also, would be great to get tests against the `Shoes::Swt::Gif` part of things too.
* I checked things out with a couple of gifs, and noticed odd white flickering when viewing http://giphy.com/gifs/the-simpsons-bart-simpson-DUfyGJvh8p9kI and a few others. Have you seen that? No idea of that's something about the file, the timing of things on my computer, or some other factor entirely.

Really excited to help get this landed, so let me know how I can assist you! :sparkles: :tada:

ducki13:
I thought about flag or sth like that in Image instead of creating new separate Gif class, and I admit that I felt quite strange when copying & pasting code from Image ;) but I decided that separate class would be more flexible in the future, eg. if someone would like to implement new functionalities, but if you prefer first option, it is ok :) (in fact I prefer it too :) ) it seems not to be very difficult task to change it.
I try to learn sth about RSpec. If you could provide me some guidance it would be great :)
When it comes to strange white flickering, I think it is gif's fault. Please, take a look at 5 and 6 frame from The Simpson gif, eg. change call in more or less 107 line in shoes/swt/gif.rb file
::Swt.display.asyncExec do
                puts frame_nr
            >>create_gif(data[frame_nr]) # change to create_gif(data[5])
                app.redraw
              end
Even when I tried to speed up gif, it is still visible.

ducki13:
**  code is of course without "puts frame_nr"

ducki13:
In my opinion this site adds damaged frames into gif when downloading (or maybe it is not website's fault). When I added sth like: if frame_nr == 5 then frame_nr = 7 (so I just ignored 2 frames), logic of frames in gif was still ok, and it looked similar to gif on website, so I deduced that this 2 frames are artifically added

ducki13:
Or maybe I am wrong, but when printing data[frame_nr].disposalMethod , then for properly working gifs there is SWT.DM_FILL_BACKGROUND value ("GIF image disposal method constants indicating that the the previous images should be covered with the background color before displaying the next image (value is 2).") but for ones with flicking white frames is SWT.DM_FILL_NONE
("GIF image disposal method constants indicating that the disposal method is to do nothing; that is, to leave the previous image in place (value is 1).").

jasonrclark:
I haven't had a chance to read through the code in more detail yet, but I did test the .gif from another player (Chrome) based on the locally downloaded file, and it didn't seem to have the same artifacts. I wonder if there's something in SWT's reading/interpreting the file that's out of whack? Will have to dig in when I've got a longer slice of time to focus and understand the GIF format since it's totally new to me.

For the testing, the DSL level looks mostly like `Image`, so when we merge things back there won't be much separate to test.

In the SWT end, though, we definitely have some things we can test.

* Given a URL, does it try to download (although I would stub out the actual downloading via a call like `expect(@dsl.app).to receive(:download)`, and not necessary try to do anything more complicated there).
* `create_gif` gets expected dimensions (or defaults)
* I'd add an actual (small) gif in the specs, and confirm that `display_gif` is able to load it.
* That said, I'd also probably stub out `Thread.new` (`expect(Thread).to receive(:new)` or something along those lines) rather than let it create a real thread
* I'd extract the guts of your `Thread.new`'s block to a separate method, and then unit test that, passing it expected loader data, and confirming the number of times through the loop we get.

That's just an outline of the main tests I'd write. Rspec-wise, the docs are pretty good, or you can also browse around Shoes' own specs for inspiration. Since most of your tests are going to be in `shoes-swt`, I'd stick to that subdirectory when looking for good examples. An example that might be of particular interest to you is https://github.com/shoes/shoes4/blob/master/shoes-swt/spec/shoes/swt/image_spec.rb.

Hope that's helpful, and let me know if there's anything more specific I can get you to help you move forward.

ducki13:
Solved! :) I just translated code from tutorial with more accuracy, earlier I had wrongly assumed that not everything from above tutorial is necessary. I will send pull request today or tomorrow and then start to write some tests (thank you for tips! :) )

PragTob:
yay gifsssss!!!!!!!! :D thank you so much!

By sending a pr do you mean a new PR deprecating this one? You could just push this branch once more to add the commits - either way happy to see it land, thanks for your work! :)

(I'll also put off the detailed code review until then) :)

ducki13:
Yes, I will push it once more :)
I have one more problem:
Everything seems to work fine, except one thing: when gif is bigger than default window, it is of course cutted and when I try to resize window to see all gif, gif disappears. I had to create separate shell and graphic context in thread (is it ok or maybe it will cause problems?), because white flickering was caused by sth connected with graphic context (paint listener handles events and obtains graphic context, then redraw image<-sth was wrong with it). As I suppose, when window is resized old shell covers new one.  Maybe you know what should I do?

PragTob:
Hi @ducki13 ! :)

I don't know if you fixed it up by now - a separate graphics context seems a bit dangerous but ok - just a feeling, no intrinsic knowledge of SWT. Might be worth checking out what was causing it and maybe fixing it :) It should also ok with a separate context.

Is it done now or are there other questions/TODOs ? :)

ducki13:
Hello :)
This solution seems to work (I hope so), I tested: several gifs from different sources, gifs with different disposalMethod() also image method called with gif from link as a source. I have one comment: for every frame from gif is sth called disposalMethod(), which tells us whether before showing current frame there is need to show previous one or repaint background (e.g. if background of gif is "invisible", then between frames there is need to paint it e.g. white, to cover previous frame). Line 153 is commented, in original Java version of gif handling this line is used, but as I noticed, for some reason it doesn't work for gifs, which source is on internet, I mean function image is called with link. Current solution seems to work, but maybe there exist gif for what it won't. E.g. this gif has yellow background when uncomment 153 line and comment 154: http://cdn.osxdaily.com/wp-content/uploads/2013/07/dancing-banana.gif .
I should also write some tests but maybe someone else would like to do it? :) In fact I have never used RSpec before and it will take me some time to find out whats exactly going on.

PragTob:
Hi there @ducki13 !

Thanks so much for your contribution!

For the tests, yes someone of us can get to it (**everyone**: if you do please comment here, so we don't duplicate work). RSpec is pretty cool though, I highly recommend it :)

Thanks for the note about the outcommented line, will add a doc there.

About the builtin exit method: normally we prefer separate pull requests for that as it makes it easier to review, concerns are clearer and enables individual fixes to land faster. It's ok for now, though :)

I gotta go to work now but I'll review this this weekend :)

PragTob:
Ah I see you are doing this pull request from your own master branch. Generally it is advised to keep the master branch clean and do all work on separate branches, then make a pull request from that branch to our master and after the PR is merged update your master again to our master. That allows the separate pull requests as every pull request then comes from a different branch and you always have a clean master (with no pening changes).

Tobi

ducki13:
I wonder why instead of using this pretty and clear solution I was doing ninja-tricks whith my repo to avoid manual merging of changes :)) Thank you for advice! :) I will use it in the future :)
I think it is hight time to start learning for my exams (however I think I will do one or two small issues), so I will probably learn RSpec, but on holidays :) If someone could do tests it would be great :)

shoes-swt/lib/shoes/swt/image.rb
```diff
@@ -77,10 +78,18 @@ def default_height
       end
 
       def download_and_display_real_image(url)
-        @tmpname = File.join(Dir.tmpdir, "__shoes4_#{Time.now.to_f}.png") unless @tmpname_or_data
+        if @is_image
```
PragTob: The repeated usage of `@is_image` suggests that polymorphism might be a better fit here, e.g. either an image subclass or another object we compose that takes care of this :)

shoes-swt/lib/shoes/swt/image.rb
```diff
@@ -89,23 +98,90 @@ def restore_width_and_height
...
+          # repeat constantly (loader.repeatCount==0) or repeat_count times
+          while loader.repeatCount == 0 or repeat_count > 0
+
+            ::Swt.display.asyncExec do
+              @off_screen_image = ::Swt::Graphics::Image.new(::Swt.display, loader.logicalScreenWidth, loader.logicalScreenHeight)
+              @gc_shell = ::Swt::Graphics::GC.new(app.shell)
+              @off_screen_img_gc = ::Swt::Graphics::GC.new(@off_screen_image)
+              @off_screen_img_gc.setBackground(app.shell.getBackground)
+              @off_screen_img_gc.fillRectangle(0, 0, loader.logicalScreenWidth, loader.logicalScreenHeight)
+              @off_screen_img_gc.drawImage(::Swt::Graphics::Image.new(::Swt.display, frame), 0, 0, frame.width, frame.height, frame.x, frame.y, frame.width, frame.height)#dsl.element_width, dsl.element_height)
```
PragTob: we prefer to call java method the_ruby_way - I'll adjust this :)

ducki13:
I tried to do an inheritance, as you said, but I met a problem: where should I distinguish which class to use (Image or Gif)? I tried to do it after image function call in dsl.rb in shoes-core, but in order to do it there, I have to add also empty gif.rb file. Is it ok?

shoes-core/lib/shoes/builtin_methods.rb
```diff
@@ -37,6 +37,13 @@ def error(message = '')
       Shoes.logger.error message
     end
 
+    def exit(message = '')
+      Shoes::LOG << ['exit', message]
```
jasonrclark: I'm not sure that we actually need to have a `message` passed with exit. Did you pick that up from the other built-in methods nearby, or some other spot?

shoes-core/lib/shoes/builtin_methods.rb
```diff
@@ -37,6 +37,13 @@ def error(message = '')
...
+      for app in Shoes.apps
```
jasonrclark: It's a minor point, but you rarely see Ruby code actually using `for`. The more common expression for this would be `Shoes.apps.each do |app|`.

shoes-core/lib/shoes/dsl.rb
```diff
@@ -203,6 +203,10 @@ def image(path, styles = {}, &blk)
       create Shoes::Image, path, styles, blk
     end
 
+    def gif(path, styles = {}, &blk)
+      create Shoes::Gif, path, styles, blk
+    end
```
jasonrclark: I think this method can go away now, since the `Image` is handling it, right?

shoes-swt/lib/shoes/swt/image.rb
jasonrclark: I agree there's an opportunity there. I'd be most in favor of actually having a separate pair of classes -- `ImageDisplayer` and `GifDisplayer` maybe?-- that we use within `Shoes::Swt::Image`.

A subclass of `Shoes::Swt::Image` would be nice in some ways, but it fights against the structure of how the rest of Shoes wants to work DSL class -> Backend class.

shoes-core/lib/shoes/dsl.rb
PragTob: yep :)

PragTob:
@ducki13 probably having a separate class internally in Swt::Image is the best solution, the creation can branch on the file type. Don't worry though, I hope that I'll get to it (I already started, but on the train and then I spent the afternoon outside) :)

shoes-core/lib/shoes/builtin_methods.rb
ducki13: From other methods :) If someone would like to pass a message before exit, why not? Then it would be better to use "puts" and display it on command line instead of "Shoes::LOG << ['exit', message]" . What do you think? :)

PragTob:
Changes and updates in #1139 ... just when I see the `exit method improved` commit.

Closing this in favor of #1139 but going to cherry pick the new commit :+1:

Issue #638, opened 2014-04-10
Support for LinkHover


jasonrclark:
Shoes3 provided a class called `LinkHover` that you could register styling against with the `style` method. When links get moused over, they'd apply that styling.

This isn't present in Shoes4 at this point at all, so we should get that addressed before the full release. Doesn't seem like an alpha blocker to me.

jasonrclark:
Fixed by #1134.

PR #1134, opened 2015-05-19
Hover all the things!

jasonrclark:
WIP for getting hovering supported, with a specific goal of getting to #638.

This gets the initial implementation for hovering built out and tested. It's done as a module, and leans on the already in place slot support for `hover`/`leave`.

Presently only works through method calls (i.e. `@p.hover do end`), not style options or global styles, but will fix up before getting this merged.

Note that `in_bounds?` for several of the objects are off per #778, so you will see some oddities to hovering because of that which won't be addressed during this PR specifically.

**Things to do**
- [x] Basic hover/leave functionality on:
  - [x] Link
  - [x] Arc
  - [x] Line
  - [x] Oval
  - [x] Rect
  - [x] Star
  - [x] Image
- [x] Shared example for hovering 
- [x] Initialize `hover` and `leave` from `initialize` options on appropriate classes
- [x] Support registering styles against `LinkHover` that gets applied and removed on hover/leave
- [x] Links aren't properly responding to styling changes--investigate, and ~~either~~ log separate issue ~~or fix (depending on difficulty)~~ (entered #1136)
- [x] Can `Slot` use `Common::Hover` instead of having its own implementation?
- [x] ~~Can we find a way to get mouse events when we're over buttons and input boxes? Doesn't appear supported on Shoes 3 and SWT is eating the mouse move events.~~ (entered #1137)


PragTob:
Looks good so far :+1:

jasonrclark:
This is done as far as I'm concerned, and covered most of what I was aiming for. Entered issues for the outstanding parts I won't handle here.

One interesting bit was that I extended the class-based hover styling to apply to anything that got an `include Shoes::Common::Hover`, so you can do the same thing with `Shoes::RectHover` that was supported previously only with `LinkHover` on Shoes 3. Seemed like a nice generalization of the pattern.

Let me know what you think. Given the link styling issues from #1136, still not ready to put `samples/simple-menu.rb` back in the mix as fully functional, but getting close!

KCErb:
`RectHover`, very cool. Love the consistency. :+1:

shoes-core/lib/shoes/common/hover.rb
```diff
@@ -0,0 +1,87 @@
+class Shoes
+  module Common
+    module Hover
+      attr_reader :hover_blk, :leave_blk
+
+      def self.included(base)
+        clazz = Class.new {}
+        name = base.name.split("::").last
+        Shoes.const_set("#{name}Hover", clazz)
```
PragTob: magic

PragTob:
:heart:

PR #1138, opened 2015-05-27
backend_for unification

### LINKED ISSUE #741: Unify code for generating backend objects

jasonrclark:
Fixes #741

This settles on one version of `backend_for`. I found, though, that we had a number of elements (`Packager`, `ImagePattern` and `Color`) where usage was not really appropriate to expect an `app.gui` (it doesn't have it today, doesn't really need it), so I made the method flexible enough to allow for that.

Might do another pass through to pull some of this out into a separate class instead of leaving it in `Shoes::Configuration`, which seems like an odd home for it. However, that'd just be a basic extraction on top of what's here.

### DIFF
--- a/shoes-core/lib/shoes/animation.rb
+++ b/shoes-core/lib/shoes/animation.rb
@@ -17,10 +17,10 @@ class Shoes
       @blk = @app.current_slot.create_bound_block(blk)
       @current_frame = 0
       @stopped = false
-      @gui = Shoes.configuration.backend_for(self, @app.gui)
+      @gui = Shoes.backend_for(self)
     end
 
-    attr_reader :current_frame, :framerate, :blk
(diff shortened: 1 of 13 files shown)


PragTob:
Looks great :+1:

Issue #1073, opened 2015-03-16
Animated gifs


PragTob:
gifs currently work when embedded through `image` but are not animated :( For someone who loves gifs, such as myself of course that's not feasible :)

Of course this a far off feature and here is [a stack overflow to get someone started](http://stackoverflow.com/questions/13479833/java-swt-animated-gif)

ducki13:
Hello, I would like to work on this enhancement. As newcomer I have some troubles with understanding of the code. Could someone give me any tips, where I should start my work? I've already read websites with gifs implementation in Java and began to read Image class implementation.

ducki13

jasonrclark:
Hey @ducki13, glad to hear you're interested to help!

So it sounds like you're on the right track looking at how to do gif's with Java. In particular, I'd keep your search restricted to things related to SWT. That's the Java library we use for our graphics (through the `swt` gem).

If I were looking to handle this, I'd actually start by seeing if I could get a basic JRuby + `swt` app working, and then look at how to translate that to Shoes' structure. The Stackoverflow and other examples would be what I'd start with, and plug them into this shell of an SWT app that I keep around for testing:

```
#! /usr/bin/env jruby

require 'swt'

class SwtExample
  def initialize
    # A Display is the connection between SWT and the native GUI.
    display = Swt::Widgets::Display.get_current
    @shell = Swt::Widgets::Shell.new
    @shell.text = "Example"

    # A Shell must have a layout. FillLayout is the simplest.
    layout = Swt::Layout::FillLayout.new
    @shell.setLayout(layout)
    @shell.set_size 200, 200

    # !!!Add your code here!!!
    @shell.add_paint_listener do |event|
      event.gc.draw_text "Yo", 0, 0
    end

    # And this displays the Shell
    @shell.open
  end

  # This is the main gui event loop
  def start
    display = Swt::Widgets::Display.get_current

    # until the window (the Shell) has been closed
    while !@shell.isDisposed

      # check for and dispatch new gui events
      display.sleep unless display.read_and_dispatch
    end

    display.dispose
  end
end

Swt::Widgets::Display.set_app_name "Example"

app = SwtExample.new
app.start
```

If you find that you need other SWT classes that aren't around by default, you can check out https://github.com/shoes/shoes4/blob/master/shoes-swt/lib/shoes/swt.rb#L6-L25 for an example of how we're pulling those in.

[...]

PR #1139, opened 2015-05-31
Gif adjustments rebased

PragTob:
This deprecates #1129 
- Squashed/reworded a couple of the original changes
- added a gif sample
- started specs

This isn't done, making the PR (changes lay around here since last week) so no one starts this afresh.

I have a couple of things that still need love:
- [ ] some more tests around the innards of `display_gif`and in general (you can refer to [jason's notes about testing this](https://github.com/shoes/shoes4/pull/1129#issuecomment-102920347) )
- [ ] extract the differences between images/gifs into a separate class so we can use polymorphism instead if `@is_image`
- [ ] refactor `display_gif`  (need: smaller methods)
- [ ] get the gif rendering back in the same shell as the normal image and make it flow with the rest (or figure out the problem, problem description below)
- [ ] ideally a check for raw image data whether it is a PNG or a GIF :|

I still want to finish this, but am not sure when (health sadly is still sort of day-to-day and I feel itchy to play with crystal) :)

Tobi


PragTob:
Cherry picked the exit method adjustments from @ducki13 :+1:

PragTob:
This has been open for way too long and obviously I haven't gotten to it. I'll close this for now but the branch will and should remain so whenever someone tackles #1073 it is here to work with.

Sorry :'(

Issue #392, opened 2013-08-19
Command Key on Mac for keypress


PragTob:
Right now the command key on Mac maps to nothing for `keypress`.

So an idea would be to map the command key to the same thing as the "super" key on Linux (which is the Windows key on Windows).

I don't know what the Mac behaviour is here yet, could someone try that out? Just run samples/simple-keypress.rb with Shoes3 and press just command, command + letter, command + alt + letter etc.

Thanks!

ccoupe:
On Linux, that super or meta  key is often grabbed by the Window manager/theme and not passed to the application (i.e. Shoes). Very dependent on the windows manager and the all the possible theme tweaks and user settings. Consider it a unusable key, cross platform.

PragTob:
Thanks for your input cecil, can confirm that Mint used to swallow it but it works these days :) Maybe we could add docs that cross platform it might be an issue, I like the general possibility though

PragTob:
Sadly it seems like the windows key isn't even supported by SWT: https://bugs.eclipse.org/bugs/show_bug.cgi?id=55236 :-/

I'm not quite sure but it seems like ot could be accessible as MOD4, whatever that means... seems like a shaky thing to do, though :-/

jasonrclark:
Boo. Well, if that's the case do we just want to call this with the `super_` on the Mac as is? I kinda still like having it available, since there are some very natural Command+ shortcuts I personally crave in my apps...

PragTob:
We _could_ look into what MOD4 is or if the state mask changes in any way on Windows/Linux and do stuff manually - I fear it either doesn't work or will be shaky at best :(

jasonrclark:
Yeah, with that being the case, I'd be in favor of closing this as it is. If some future version SWT or a different backend supports it, there's now a home for those type of keypresses. Sound reasonable?

PragTob:
:+1:

PR #1142, opened 2015-06-13
Add super_ prefix for command key on Mac

jasonrclark:
Part of a fix for #392, this handles the Mac side of things for mapping Command to a prefix.

I've done some searching, and I don't see a ready-made way to sense the press of a Windows key to map to the same `super_` prefix. Might try to test on my actual Windows machine this weekend to see what, if any, event is fired to determine if we can even support that.

No clue what, if anything, is the right thing to map over in Linux-land.

:key:


jasonrclark:
Bit of discussion on #392 about Linux... any objections to merging this since it handles the Mac case neatly. I will research the Windows side of things before calling #392 closed, along with whatever call we make on Linux.

PragTob:
Seems good to me, I'll give Linux a look :)

Issue #446, opened 2013-10-02
Swt::Common::Child mixin


rjack1201:
There is a Swt::Common::Child mixin module which appears to be broken in some classes because it doesn't climb the parent tree. I'm not sure the mixin is actually used for anything because if I comment out its app method, all tests still pass. 

We should either get rid of it or add tests and make it work properly.

jasonrclark:
Closed via #1143

PR #1143, opened 2015-06-15
Remove Shoes::Swt::Common::Child

jasonrclark:
Per #446, this module was a little mixed up and backward. It assumed that the `parent` passed to an SWT class would return the right `app`, which it happened to... but `parent` was actually already the `app` itself we wanted!

This straightens that all out, which simplifies the SWT backends that were using `Child` a bit, and gets rid of a whole module. Huzzah for :hocho: code!


PragTob:
:heart:

PR #1147, opened 2015-07-04
Load up backend if not already done when app starts

### LINKED ISSUE #1146: Runnign scripts through ruby myscript.rb is broken
Take this script:

```
require 'shoes'

Shoes.app {button 'hello'}
```

trying to run it via `ruby my_script.rb` (or prepended with `bundle exec`) yields this error:

```
NameError: uninitialized constant Shoes::Swt::App
  const_missing at org/jruby/RubyModule.java:2733
      setup_gui at /home/tobi/github/shoes4/shoes-core/lib/shoes/internal_app.rb:30
     initialize at /home/tobi/github/shoes4/shoes-core/lib/shoes/app.rb:48
            app at /home/tobi/github/shoes4/shoes-core/lib/shoes/app.rb:19
```

Running through `bin/shoes` of course works fine. I'm guessing that when reworking the backend selection we forgot about good old running scripts with ruby and requiring shoes. We should get that back to possible again :)

Buck courtesy of [this stackoverflow](http://stackoverflow.com/questions/31040742/shoes-cannot-find-its-own-backend) - I'm going through stackoverflow right now searching for these kind of things (really I'm just putting of unpleasant work but whatta you know.. :) )

Tobi

jasonrclark:
Fixes #1146.

This helps when not running through `bin/shoes`. We end up with a defaulted backend value, but nothing actually loaded yet.

This also required allowing multiple attempts to set the backend to the same value.

### DIFF
--- a/shoes-core/lib/shoes/configuration.rb
+++ b/shoes-core/lib/shoes/configuration.rb
@@ -29,6 +29,8 @@ class Shoes
       # @example
       #   Shoes::Configuration.backend = :swt # => Shoes::Swt
       def backend=(name)
+        return if @backend_name == name
+
         unless @backend.nil?
           fail "Can't switch backend to Shoes::#{name.capitalize}, Shoes::#{backend_name.capitalize} backend already loaded."
         end
--- a/shoes-core/lib/shoes/internal_app.rb
+++ b/shoes-core/lib/shoes/internal_app.rb
@@ -27,6 +27,7 @@ class Shoes
     end
 
     def setup_gui
+      ensure_backend_loaded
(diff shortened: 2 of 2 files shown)


PragTob:
BEAUTIFUL! Thank you <3

aagradwan:
> This helps when not running through `bin/shoes`. We end up with a defaulted backend value, but nothing actually loaded yet.
>
> This also required allowing multiple attempts to set the backend to the same value.
**soryy im on arch linux and i cant use this its stel** 

> Traceback (most recent call last):
> hello.rb:3:in `<main>': uninitialized constant Shoes (NameError)

PR #1154, opened 2015-08-01
Restore idea of needs_positioning for backgrounds

### LINKED ISSUE #599: background color hides text_block
When setting the background color after creating a text block, the text block gets hidden.

```ruby
Shoes.app { banner '0:00' ; background '#ff0000' }
```
I'm on Linux. Same behavior on current Shoes 4 and Red Shoes. Green shoes not affected. The reverse works fine:

```ruby
Shoes.app { background '#ff0000' ;  banner '0:00'  }
```

jasonrclark:
Fixes #599

Background and Border derive their whole positioning from their parent,
so they need to be skipped in the explicit positioning logic that slots
run over elements.

### DIFF
--- a/shoes-core/lib/shoes/common/background_element.rb
+++ b/shoes-core/lib/shoes/common/background_element.rb
@@ -5,6 +5,11 @@ class Shoes
         @dimensions = ParentDimensions.new @parent, @style
       end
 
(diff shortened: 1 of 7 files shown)


PragTob:
Seems great :+1:

PR #1163, opened 2015-10-05
Restore separate Guardfiles for each gem

### LINKED ISSUE #950: Guardfile fix after separating out shoes gems
The changes from #949 have broken our Guardfile. Part of it is just the paths not matching anymore, but it didn't seem to be picking up changes when I edited the file to align with our relocated specs. Needs more investigation.

jasonrclark:
Fixes #950 

With the new gem layout, we needed to revise things to get Guard running
again. It also occurred to me that the way that Guard was always running
under SWT wasn't ideal, so now you can choose which component to run
Guard against:

```
rake guard:core
rake guard:swt
rake guard:package
```

I'm seeing some oddities on packaging, but no more than I see running
individual spec files outside of Guard, so just going forward with it
for now.

*Worth noting that this (at last check) was only working consistently on
JRuby 9000, not 1.7.x. Will be double-checking soon, and potentially see
about getting a patch on the necessary Guard project, but otherwise,
might be worth moving on forward to JRuby 9k for most development.*


PragTob:
:heart: :+1: :) Thanks!

ruby-version is already set to 9k, we might want to add a clearer note to the README then though.

PR #1164, opened 2015-10-07
Fix typo.

rin:


PragTob:
thanks!!! (+ *cough*)

I'll add a test after work today to make sure that doesn't happen again...

Issue #710, opened 2014-05-18
Use the default system colors for a more native look and feel


PragTob:
Every OS has their own default background color and maybe even button colors. In fact [a whole bunch of them starting at COLOR_INFO](http://help.eclipse.org/indigo/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fswt%2FSWT.html) - scroll down for more. E.g. COLOR_WIDGET_BACKGROUND seems interesting for setting the default background color of an app (as we ntoiced these are different between Linux, Mac and Windows).

That would mean to implement methods/constants referring to those values in a shoes friendly way in the backend.

Relates to #601

KCErb:
Looks like @PragTob's link is broken, here's the link to the Swt docs that give those color ints

http://help.eclipse.org/luna/index.jsp?topic=/org.eclipse.platform.doc.isv/reference/api/org/eclipse/swt/package-summary.html

jasonrclark:
Closed by #1166

PR #1166, opened 2015-10-15
first go at reading in a system color for background

KCErb:
Woohoo! First PR in more than 6 months; feels good to look back at shoes-things after so much time.

This is a start at addressing #710. It just pulls in the one system color but I can imagine being a little more general and letting `Swt::Color` see more system colors.

It also pushes all of the background handling to the backend but we could let the frontend have access through something like `:system_gray` and then have the backend grab this in response...

I'm open to ideas for improvement but also OK with this just landing as is :)

Specs look OK to you guys? I played around a bit to make sure they do what I think they do . . .


PragTob:
welcome back :heart: (at work, no review, will do later, ping me if I'm lazy)

shoes-swt/lib/shoes/swt/app.rb
```diff
@@ -154,6 +154,14 @@ def wait_until_closed
 
       private
 
+      def initialize_background
+        @background = if @dsl.opts[:background]
+          Color.new(@dsl.opts[:background]).real
+        else
+          Color.system_background_color
```
jasonrclark: If we're including the background in the `DEFAULT_OPTIONS` above, do we hit this branch in practice? Would be great to avoid the conditional if we could know we're going to get a color set  in the DSL.

jasonrclark:
So great to see you back! :sparkles: 

Looks pretty reasonable to me. The specs seem good to me for this particular implementation.

That said, another potential way we could approach this is by adding some `system` colors to the primary DSL's color list. Rather than getting defined by the numeric values like https://github.com/shoes/shoes4/blob/master/shoes-core/lib/shoes/color.rb#L223, these would do the system lookup in some way.

The upside of this is that then these system colors aren't any different than other colors. They get a special loading path, but after that they're just colors. Does that make sense? Sound at all appealing to try?

KCErb:
Yeah, I'll give it a go. But which colors? SWT has a short list so that's a start. Then add something like `system_background`?

In that case do you envision putting the default background back onto the front end?

jasonrclark:
So what I'd envision is the front-end colors list having a small set of `system_background`, `system_foreground`, `system_*` colors in it, which we'd load up specifically.

If you want, totally just start with `system_background`. I'm not sure what others would actually make sense to use.... did you have a link to the list of what SWT presents that we could review?

KCErb:
The first color in the list is here:

http://help.eclipse.org/luna/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/swt/SWT.html#COLOR_BLACK

and here's a copy paste of all the colors

```
static int	COLOR_BLACK
  Default color black (value is 2).
static int	COLOR_BLUE
  Default color blue (value is 9).
static int	COLOR_CYAN
  Default color cyan (value is 13).
static int	COLOR_DARK_BLUE
  Default color dark blue (value is 10).
static int	COLOR_DARK_CYAN
  Default color dark cyan (value is 14).
static int	COLOR_DARK_GRAY
  Default color dark gray (value is 16).
static int	COLOR_DARK_GREEN
  Default color dark green (value is 6).
static int	COLOR_DARK_MAGENTA
  Default color dark magenta (value is 12).
static int	COLOR_DARK_RED
  Default color dark red (value is 4).
static int	COLOR_DARK_YELLOW
  Default color dark yellow (value is 8).
static int	COLOR_GRAY
  Default color gray (value is 15).
static int	COLOR_GREEN
  Default color green (value is 5).
static int	COLOR_INFO_BACKGROUND
  System color used to paint tooltip background areas (value is 29).
static int	COLOR_INFO_FOREGROUND
  System color used to paint tooltip text (value is 28).
static int	COLOR_LINK_FOREGROUND
  System color used to paint link text (value is 36).
static int	COLOR_LIST_BACKGROUND
  System color used to paint list background areas (value is 25).
static int	COLOR_LIST_FOREGROUND
  System color used to paint list foreground areas (value is 24).
static int	COLOR_LIST_SELECTION
  System color used to paint list selection background areas (value is 26).
static int	COLOR_LIST_SELECTION_TEXT
  System color used to paint list selected text (value is 27).
static int	COLOR_MAGENTA
  Default color magenta (value is 11).
static int	COLOR_RED
  Default color red (value is 3).
static int	COLOR_TITLE_BACKGROUND
  System color used to paint title background areas (value is 31).
static int	COLOR_TITLE_BACKGROUND_GRADIENT
  System color used to paint title background gradient (value is 32).
static int	COLOR_TITLE_FOREGROUND
  System color used to paint title text (value is 30).
static int	COLOR_TITLE_INACTIVE_BACKGROUND
  System color used to paint inactive title background areas (value is 34).
static int	COLOR_TITLE_INACTIVE_BACKGROUND_GRADIENT
  System color used to paint inactive title background gradient (value is 35).
static int	COLOR_TITLE_INACTIVE_FOREGROUND
  System color used to paint inactive title text (value is 33).
static int	COLOR_WHITE
  Default color white (value is 1).
static int	COLOR_WIDGET_BACKGROUND
  System color used to paint background areas (value is 22).
static int	COLOR_WIDGET_BORDER
  System color used to paint border areas (value is 23).
static int	COLOR_WIDGET_DARK_SHADOW
  System color used to paint dark shadow areas (value is 17).
static int	COLOR_WIDGET_FOREGROUND
  System color used to paint foreground areas (value is 21).
static int	COLOR_WIDGET_HIGHLIGHT_SHADOW
  System color used to paint highlight shadow areas (value is 20).
static int	COLOR_WIDGET_LIGHT_SHADOW
  System color used to paint light shadow areas (value is 19).
static int	COLOR_WIDGET_NORMAL_SHADOW
  System color used to paint normal shadow areas (value is 18).
static int	COLOR_YELLOW
  Default color yellow (value is 7).
```

KCErb:
OK, so here's another thing to discuss: does this mean we want to directly tie the dsl color to the gui color?

Right now the frontend and backend of a color are indirectly linked via the element.

When I create a color on the front end, it never gets a backend representation. Instead the element that uses the color passes that info to *its* backend-representation which then makes a gui color.

So right now for example, if I create a shoes color called `system_background` then there's no way to say `system_background.red` unless we have that call instantiate a backend class.

I vote that we create a new class `Shoes::SystemColor` which has a backend representation. It uses that backend to determine its rgb values and then creates a `Shoes::Color` version of itself and disposes of the backend version.

How does that sound?

jasonrclark:
Realizing what I wrote wasn't very clear. The danger of late-night commenting :grin: 

When I said `system_background`, I didn't mean a method for setting the background. What I meant was that `system_background` is an instance of `Shoes::Color`, just like `red`, or `blue`. The only difference is in how it gets initialized.

This would mean the `Shoes::Color` DSL object would now have two ways to represent a color--either by the numeric RGB, or by a symbol that maps to a system color. Then [when the backend loads the color]( https://github.com/shoes/shoes4/blob/master/shoes-swt/lib/shoes/swt/color.rb#L13) we'd conditionally load based on which type of color was requested. (There may be some object-oriented, subclassing approach with a `Shoes::SystemColor`, but I wouldn't that to leak out to end users that there's a difference between the color implementations).

KCErb:
OK, cool that makes sense. I'll see if I can't put something together along those lines. Thanks @jasonrclark!

KCErb:
Wow, ok, that took more than expected. Adding a new dsl element is sort of a big deal. But I'm happy with the result. Now `system_background` is a full-fledged shoes color just like `tomato`.

Let me know how this looks, I'm happy to clean up / roll back things that are undesirable.

shoes-core/lib/shoes/color/dsl.rb
```diff
@@ -0,0 +1,169 @@
...
+        color = Shoes::COLORS.fetch(color_name)
+        return color if alpha == Shoes::Color::OPAQUE
+        Shoes::Color.new(color.red, color.green, color.blue, alpha)
+      end
+    end
+
+    # define the system colors
+    def system_background
+      Shoes::SystemColor.new(:system_background)
+    end
```
PragTob: Personally I didn't envision it like this, more a bit like really one of the other colors, I'll try to elaborate :)

shoes-core/lib/shoes/color/dsl.rb
```diff
@@ -0,0 +1,169 @@
...
+      [:turquoise, 64, 224, 208],
+      [:violet, 238, 130, 238],
+      [:wheat, 245, 222, 179],
+      [:white, 255, 255, 255],
+      [:whitesmoke, 245, 245, 245],
+      [:yellow, 255, 255, 0],
+      [:yellowgreen, 154, 205, 50],
+      [:shoes_background, 237, 237, 237],
+    ]
+
```
PragTob: I envisioned the system color being looked up somewhere around here (e.g. on load time of shoes, as the color should not change). Then we go to the backend one time and ask it for its background color and then add it to the list. I.e. something like (pseudo code!):

```ruby
system_background = @gui.system_color(:background) # especially uncertain here what to call/how to get it)
colors << [:system_background, system_background.red, system_background.green, system_background.blue]
```

That way we look up the system background color once on startup and from that point on it is just one normal color with the normal values assigned just like any other color.

The backend could even know the format (or maybe it should?) and we could do something like `app.gui.system_colors` or just `gui.system_colors` that then returns arrays like we'd expect them to. So like: `[[:system_background, red, green, blue], [:system_font, red2, green2, blue2]]`

Or the backend could have the responsibility of adding system colors so we'd do:

```
colors = some_backend_thing.add_system_colors(colors)
```

So many possibilities...

PragTob:
Hey @KCErb ! Thanks for your work on this. I wrote up my thoughts in [this comment](https://github.com/shoes/shoes4/pull/1166#discussion_r42319233). Basically all colors are defined on startup (e.g. as we require files etc.) and I'd like system colors to be the same. So when we start defining all the colors, we make some call to the backend (have no good idea for the API yet) to get the system colors, then just add it to the `colors` data structure and treat them like plain old colors and don't care that they are connected to the backend at all once shoes has booted up :)

Does that make sense? :)

Cheers and thanks a lot!
Tobi

KCErb:
Oh man Tobi, that sounds like a WAY better approach than mine! :smile: 

I'll definitely switch over to that. I'm glad I did it my way first though because it was a great re-intro to the codebase.

Are you guys for or against my hanging on to the `Swt::SystemColor` class to support that `@gui.system_color(:background)` call. I'm for it.

jasonrclark:
Also a big fan of your idea to load it up at boot time @PragTob. Much nicer than what I was proposing with the conditional in the backend color.

@KCErb I kind of wouldn't be inclined to create a full, separate backend class, just because it requires a DSL class to support it. That plus the fact that we don't expect any user of the DSL to actually interact with these--they're just an artifact of how we load things.

That said, the relationship between `Shoes::Color` and `Shoes::Swt::Color` doesn't actually follow the normal DSL->backend layout. There's some reason for it behind needing to dispose of resources vs the DSL colors being app-lifetime constants... anyway, all that to say if you find yourself in a snarl, do holler for help, and maybe a new backend class *would* be easier :)

KCErb:
So the first trick I'm running into is that just at the point where we'd like to call `gui.system_color` or `app.gui.system_color` there is no access to the backend. Colors don't have a backend and app isn't available from the module, just the class that includes it.

So I see two options and I'm happy to have someone point out a third.

1. Go a little down the road I have already and create a frontend and backend class for a system color so that I can call it directly. I'd still use that class the way we talked about here so that we get the values once only.

2. Do something clever so that I can talk to `@__app__` in the including class.

```ruby
module DSL
  define_method # ... do
    # ...
    @__app__.gui.system_color(:background) # etc
    # . . .
  end
end
```

Suggestions, ideas, preferences?

jasonrclark:
Given the weirdness around colors and backends, I'd be inclined to go with either 1) or the additional option 3) of changing how `Shoes::Color` relates to it's backend.

`Shoes::SystemColor` seems fine under the circumstances, though, as long as it isn't surfaced by the DSL directly. We do have at least one other case of this type of class in the relationship from front-end to packaging, so there is precedent.

PragTob:
Hello, hellooo!

Hm. Of course we don't have a backend yet! Dam you software, why are you so complex?

How about, that we make system color initiailization part of the app initialize process? It just calls something like "initialize_system_colors" (before it sets its own background of course). That then gets the colors from the backend (which we have at that point) and defined the appropriate color methods (using some color help method).

KCErb:
OK, that seems reasonable, I'll try from there. Thanks guys!

PragTob:
Also another thought I had, we could move color initialization as a whole more down the road (not sure to where). 

Also, getting the system colors could easily be a class method and you could use `Shoes.backend::SomeClass` as the entry point to get them, even in the current colors (maybe, not sure when exactly the backend is determined so that Shoes.backend returns Shoes::Swt)

jasonrclark:
Actually really like the `Shoes.backend::SomeClass` idea, although I'm not sure that the backend is loaded when we're requiring the `shoes/color.rb` file.

In thinking about it further, I also like @PragTob's suggestion to just have the SWT app initialize the colors... it could easily accomplish this by calling a (as yet undefined) method on `Shoes::Color` to add system colors in there. This feels better to me, since it's the backend feeding system-specific info to the DSL instead of the other way around.

Sorry to keep flinging options around @KCErb and looking forward to what you come up with!

KCErb:
It's all good. Fun to discuss some of the corners of the codebase.

So how about something like this

1. Init the app
2. app creates its own backend
3. backend digs up system colors and calls front end method for inserting them into the dsl

So I'm creating two pieces:

```ruby
class InternalApp
  def initialize
    # ...
    @gui.fetch_system_colors
  end
end
```

```ruby
class Swt
  class App
    def fetch_system_colors
      background_color = # . . .
      Shoes::Colors.insert_system_color(:system_background, background_color)
    end
  end
end
```

Something along those lines?

PragTob:
looks good to me, the only potential problem that I see with the backend defining them is that some colors might not be defined (depending on the backend) - we could easily prevent that though by first defining some default colors and then redefining them or something.

KCErb:
So yes, putting defaults on the frontend that get over written is one failsafe. Isn't another to just have a well-defined list of system colors that a backend must support?

That way it's more like the sort of problem where the backend doesn't support `oval`, it *has* to, so we put it in the mock backend.

I'm fine either way, I'm just wondering from a design stand point which is preferable.

Also right now we only have one color we're interested in supporting right?

PragTob:
There is a color of more colors that we might want to support so it's good to build with that in mind, if it isn't too hard.

The difference here is if the backend doesn't find it it crashes with undefined method on something from the DSL layer. Plus, we will use it as the default background color so no app for a new backend will ever start without it being defined which is a bit, meeehh. Plus the error is a bit harder/different. Normally the DSL layer _defines_ all methods and then just uses backend classes to paint them. So you get the helpful error that we expected there to be a Shoes::Swt::Oval class. With this we just get an error that there is no method called like that defined for App (and/or the main object).

Do I sort of communicate what the difference is for me? :)

KCErb:
Ah yes, I see. The way a system color would be integrated would make it a little *too* mission critical to not make sure there's a fallback value.

OK cool, thanks for explaining :sunglasses:

PragTob:
Hey @KCErb - any chance you'll get to work at this soon? Don't worry if not, just checking in.

KCErb:
Yes, I just finished up a big project yesterday and am hoping to get some homework finished today. So I anticipate having at least an hour or two between now and Sunday if not more. 

Thanks for checking in!

KCErb:
OK, here's a new approach, let me know how it looks :)

I'm struggling with the tests though! I'll comment in-line.

shoes-core/spec/shoes/shared_examples/dsl/rgb.rb
```diff
@@ -5,7 +5,8 @@
   let(:alpha) { 133 } # cornflowerblue
 
   it "sends args to Shoes::Color" do
-    expect(Shoes::Color).to receive(:new).with(red, green, blue, alpha)
+    expect(Shoes::Color).to receive(:new).ordered
```
KCErb: This test is failing now because when the app gets created it calls `Shoes::Color` to setup the system color and this expectation gets fired.

The solution I found online is to use `ordered` like this so that you say: I want the second time it gets called to be the following.

The only trouble is that the shoes specs (with a Mock backend) only call `Shoes::Color` one time so it fails this way too. Anyone know the right way to handle this?

shoes-swt/spec/shoes/swt/app_spec.rb
```diff
@@ -122,4 +128,20 @@
       it_behaves_like 'reports client area dimensions'
     end
   end
+
```
KCErb: Here are the only two specs I tried to add to spec this out. I've commented out the second one because it has been tough to figure out because this whole system color thing requires a full front and backend.

I could stub in something that acts like a `Shoes::Color`  on the double . . . is that useful though?

shoes-core/spec/shoes/shared_examples/dsl/rgb.rb
PragTob: You could go `allow(Shoes::Color).to receive ...`

Then after you did whatever you wanted to do you can do: `expect(Shoes::Color).to have_received(:something).exactly(n).times` (there is also at_least and others)

shoes-swt/spec/shoes/swt/app_spec.rb
PragTob: Please comment below the code you want to discuss (then it shows up on the comments overview :) )

shoes-swt/spec/shoes/swt/app_spec.rb
```diff
@@ -122,4 +128,20 @@
...
+      expect(background).to eq swt_salmon
+    end
+
+    # Don't see a way to spec this since it requires a full front end
+    # it 'should have the default system background when unspecified' do
+    #   plain = Shoes::Swt::App.new plain_app
+    #   default_background = ::Swt.display.getSystemColor(::Swt::SWT::COLOR_WIDGET_BACKGROUND)
+    #   background = plain.shell.background
+    #   expect(background).to eq default_background
+    # end
```
PragTob: Why don't you just do a full app? `Shoes::App.new` should be fine, there are also a couple of specs that do that (and one that is called `integration_spec.rb` or soemthing).

shoes-swt/spec/shoes/swt/app_spec.rb
PragTob: See the comment below for one approach. However mocks are also just fine. You test an object in isolation and otherwise you just test that it sends the correct messages and then that the other object receives and handles those messages correctly. So, here for instance:

one spec:
* stub Shoes.display.getSystemColor to return a well known color (would be different depending on OS otherwise)
* assert that `::Shoes::DSL` receives `define_shoes_color` with the right arguments (rgb, shoes_background_color)

In another test:
* assert that Shoes::DSL.define_shoes_color does what you'd expect it to do given the right inputs

You don't need to explicitly test Shoes.display.getSystemColor as it is a library method :)

Does that help?

You can do the integration style test or the mock or both :)

KCErb:
> Please comment below the code you want to discuss

Thanks for this, I've often gotten this wrong and wondered why I can't tell github which lines I'm commenting on!

Thanks for the other suggestions too! I'll try implementing them and get back to ya.

KCErb:
OK, I think I've got some decent specs in here now. There's only one I'm still not sure of. I'll comment over there.

shoes-swt/spec/shoes/swt/app_spec.rb
```diff
@@ -122,4 +124,36 @@
...
+        .to receive_messages(app_title: 'double',
+                             width: 0,
+                             height: 0,
+                             opts: { background: Shoes::COLORS[:system_background] })
+
+      default_background = ::Swt.display.getSystemColor(::Swt::SWT::COLOR_WIDGET_BACKGROUND)
+      app = Shoes::Swt::App.new(Shoes::App)
+      background = app.shell.background
+      expect(background).to eq default_background
+    end
```
KCErb: I claim that this test is when background is unspecified. But then I stub in a bunch of methods on `Shoes::App` I did this just to make the test work :(

I guess I still just don't get RSpec in a fundamental way. I couldn't understand why things like `opts` and `app_title` were `nil` when I was using a "real" Shoes App in stead of a double.

wasnotrice: @KCErb looks like you are using the class `Shoes::App` instead of an instance `Shoes::App.new`

KCErb: Oh yeah, thanks! You're absolutely right.

I switched that our for this now:

```ruby
app = Shoes::Swt::App.new(Shoes::InternalApp.new(Shoes::App.new, {}))
```

unfortunately the test is now failing because the app is using the fallback color: `[237, 237, 237]` instead of the swt color, which for me is `[232, 232, 232]`.

I'll look around and see if I can understand why.

PragTob: Any luck figuring it out? Want to push newest state so we can help? :)

(sorry for the long silence from my side, had a hard week with 2 presentations)

KCErb:
Hey @PragTob, I just pushed up what's not working. I haven't figured this out yet :\

jasonrclark:
So dug a bit and figured out the trouble. What's happening with the defaulting is that [this line](https://github.com/shoes/shoes4/pull/1166/files#diff-4aa8d1cb0b6621b713f277f0c6e6b4d8R20) sets the default before the backend has been loaded and it never gets reassigned.

I've sent a [PR against your branch](https://github.com/shoes/shoes4/pull/1171) @KCErb that would fix it, but it shuffles things around a bit so wanted your thoughts on it before just rushing forward.

KCErb:
I think your changes are great. Only two things were a little mystifying. 

1. You removed the app method from the mock backend but it looks like your code still uses it. I thought that backend needed methods like that.

2. How did you create a pull request against a pull request? It says the location was `shoes/jclark/system_background`. I've never seen such a thing; it seems really useful.

jasonrclark:
1. So the `setup_system_colors` method as I wrote it is called from within `Shoes::Swt`'s own initialization. Since the call is within the backend entirely, not DSL -> backend, the mock doesn't need it.

That's actually a bit we lose with my changes--"updating your system colors" is left entirely up to the backend, and not triggered from the DSL. A new backend wouldn't know to do it unless they read the code... we could preserve that (probably a method we'd call on the backend class immediately after calling `initialize_backend`) but I'm not too worried about it.

2. When you start a PR there should be a couple of dropdowns on the top to select branches. You can change the one pointed to `master` to point to any arbitrary branch. :sparkles:

PragTob:
regarding 1) I think it's fine to sort of drop that, that way backends can start handling that when they are at that point in the implementation, it's not the most critical part... we made it such a long time without this :)

Looks good the way it is now to me. I'd prefer it to be rebased/squashed a bit as the implementation was changed some times and those commits are potentially confusing on master. So I'd like to be cut down a bit, not necessarily to just one commit but a couple of commits or so :) If you wan to you can try yourself, otherwise I might if everyones's cool with that :)

jasonrclark:
Squash away! :athletic_shoe:

KCErb:
Huh, just googled what squash and rebase mean in terms of git. It's surprising how much there is to learn. Sounds like a great idea.

PragTob:
shoes 4 _209

looks marvellous!

PR #1167, opened 2015-10-18
Fix up handling of explicitly sized text blocks

### LINKED ISSUE #664: TextBlock#width should set width of text block
Let's say I want to mono-space a mono-space font, by setting each character in its own little para with a fixed width. 
```ruby
Shoes.app do
  style(Shoes::Para, font: "Monospace 12px")
  s = "Here is some monospaced text"
  s.each_char do |char|
    para char, width: 10
  end
end
```

Instead of a flow of paras with a fixed width, I get a flow with a fixed width:

weird

There are probably better ways to monospace a mono-spaced font (kerning? shouldn't shoes monospace it?) but the bug stands!

jasonrclark:
Fixes #664

If we were provided a text block with an explicitly declared width, that
wasn't being properly respected in deciding where to position the next
element after it, or in how much space we asked for. These conditions
are now properly handled.

### DIFF
--- a/shoes-core/lib/shoes/text_block_dimensions.rb
+++ b/shoes-core/lib/shoes/text_block_dimensions.rb
@@ -29,15 +29,31 @@ class Shoes
       10
     end
 
-    # This is the width the text block initially wants to try and fit into.
-    def desired_width(containing = containing_width)
-      parent.absolute_left + containing - absolute_left - margin_left - margin_right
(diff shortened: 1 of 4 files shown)


KCErb:
Looks clean, simple, and logical. I say it's good to merge.

jasonrclark:
Thanks @KCErb!

PragTob:
:+1:

Issue #1036, opened 2015-01-18
Slots content should be hidden when the slot is too small to display them


PragTob:
Right now we paint elements no matter what the dimensions of a slot are:

``` ruby
Shoes.app do
  stack height: 0 do
    para 'Y U SEE ME?'
  end
end
```

Shoes 4
![screenshot from 2015-01-18 19 41 27](https://cloud.githubusercontent.com/assets/606517/5793604/5a13f464-9f4a-11e4-894a-786db9eff6e8.png)

Shoes 3.2
![screenshot from 2015-01-18 19 41 52](https://cloud.githubusercontent.com/assets/606517/5793607/628ea710-9f4a-11e4-9cb8-5c876ca2358f.png)

It gets more interesting if you have something big and a smaller slot:

``` ruby
Shoes.app do
  stack height: 25 do
    title 'Y U SEE ME?'
  end
end
```

Shoes 4:

![screenshot from 2015-01-18 19 45 21](https://cloud.githubusercontent.com/assets/606517/5793616/bd33be58-9f4a-11e4-8197-ab1943c706b3.png)

Shoes 3.2:

![screenshot from 2015-01-18 19 46 02](https://cloud.githubusercontent.com/assets/606517/5793617/c8d1a1bc-9f4a-11e4-9245-19bcc60a94f4.png)

I guess to correctly tackle this we'd need a backend for slots #905 

Discovered during lookign at #1021

jasonrclark:
Good find. Curious what the behavior looks like for `button` and `rect` (which each use a bit different scheme from `TextBlock` which is it's own special :snowflake: )

jasonrclark:
Double-checked and Shoes 3 does apply that behavior to the other element types I checked (rect and button). Shouldn't be too tough to apply to the painter-type elements (text and rect), but not sure how we'd accomplish that with SWT-widget-based things. Should be fun :grin:

jasonrclark:
@KCErb won't be tromping on anything you're doing if I fiddle with the painting side of this, will I? Those won't be reliant at all on a `Slot` backend to function--they're done a whole different way.

KCErb:
Definitely not, haven't started on this because I'm still on the colors part :P

jasonrclark:
This is addressed for text and art elements by #1170, but work remains for the control elements. Punting that to a later release.

jasonrclark:
[...]

PR #1170, opened 2015-10-21
Slot clipping for painted elements

jasonrclark:
This PR will cover what we need on #1036 for non-control elements--all art and text elements in Shoes.

Basically all that's necessary is to set the "clipping" on any graphics context before we draw with it. We set the clipping back to it's original state afterward, using a helper method added on the (poorly named) `Shoes::Swt::Common::Resource`.

WIP because this deserves some actual specs, not just to pass the existing ones, and I might rename `Resource` while I'm at it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,6 @@ group :development do
   gem "kramdown"
   gem 'simplecov'
   gem "codeclimate-test-reporter"
-  gem 'jruby-lint'
   gem 'webmock'
   gem 'hometown'
   gem 'rubocop'
--- a/shoes-swt/lib/shoes/swt/common/resource.rb
+++ b/shoes-swt/lib/shoes/swt/common/resource.rb
@@ -24,10 +24,12 @@ class Shoes
           @graphic_contexts << graphics_context
         end
 
-        def clip_context_to(graphics_context, element)
+        def clip_context_to(graphics_context, element, use_element_height = true)
           clipping = graphics_context.clipping
+          height   = use_element_height ? element.height : clipping.height
+
           graphics_context.set_clipping(element.absolute_left, element.absolute_top,
-                                        element.width, element.height)
+                                        element.width, height)
           yield graphics_context
         ensure
           if clipping
--- a/shoes-swt/lib/shoes/swt/common/resource.rb
+++ b/shoes-swt/lib/shoes/swt/common/resource.rb
@@ -26,7 +26,7 @@ class Shoes
 
         def clip_context_to(graphics_context, element, use_element_height = true)
           clipping = graphics_context.clipping
-          height   = use_element_height ? element.height : clipping.height
+          height   = use_element_height ? element.height : element.app.height
 
           graphics_context.set_clipping(element.absolute_left, element.absolute_top,
                                         element.width, height)
--- a/shoes-swt/lib/shoes/swt/common/resource.rb
+++ b/shoes-swt/lib/shoes/swt/common/resource.rb
@@ -25,7 +25,7 @@ class Shoes
         end
 
         def clip_context_to(graphics_context, element)
-          clipping = graphics_context.get_clipping
+          clipping = graphics_context.clipping
           graphics_context.set_clipping(element.absolute_left, element.absolute_top,
                                         element.width, element.height)
           yield graphics_context
--- a/shoes-swt/lib/shoes/swt/common/resource.rb
+++ b/shoes-swt/lib/shoes/swt/common/resource.rb
@@ -23,6 +23,18 @@ class Shoes
         def track_graphics_context(graphics_context)
           @graphic_contexts << graphics_context
         end
+
+        def clip_context_to(graphics_context, element)
+          clipping = graphics_context.get_clipping
+          graphics_context.set_clipping(element.absolute_left, element.absolute_top,
+                                        element.width, element.height)
+          yield graphics_context
+        ensure
+          if clipping
+            graphics_context.set_clipping(clipping.x, clipping.y,
+                                          clipping.width, clipping.height)
+          end
+        end
       end
     end
   end
--- a/samples/expert-game-of-life-adjusted.rb
+++ b/samples/expert-game-of-life-adjusted.rb
@@ -192,7 +192,8 @@ end
 Shoes.app(title: "The Game of Life", width: 800, height: 620, resizable: false) do
   background white
   @animate = false
-  stack(margin: 10) do
+
+  stack(margin: 10, width: 650, height: 620) do
     @new_world = World.new(40, 40, self)
     animate(10) do
       if @animate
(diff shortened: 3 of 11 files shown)


PragTob:
Looks good, looking forward to the specs :+1: 

Love it how our painter inheritance centralizes this! :D

jasonrclark:
Added specs, should be ready for prime time...

```
Shoes.app do
  stack width: 200, height: 100 do
    para "text " * 100
    rect 50, 50, 300, 300, fill: green
  end
end
```

**Before**
image

**After**
image

jasonrclark:
Hrm, failure seems to be related to some gem changes:

```
Gem::InstallError: tins requires Ruby version >= 2.0.
Using shoes-swt 4.0.0.pre4 from source at shoes-swt
An error occurred while installing tins (1.7.0), and Bundler cannot continue.
```

`tins` just released 1.7 which has a Ruby 2.0 requirement. Appear that we get that requirement through the use of `jruby-lint` which is in our Gemfile but I don't see much usage of. Will try dropping that gem...

Gemfile
```diff
@@ -23,7 +23,6 @@ group :development do
   gem "kramdown"
   gem 'simplecov'
   gem "codeclimate-test-reporter"
-  gem 'jruby-lint'
```
PragTob: :+1: removing jruby-lint, it's a gem to detect incompatible gems and bad usage patterns etc... so it's nice but hasn't been updated since 2013 :( Also so far it never reported anything for us

jasonrclark:
Ran through all the samples, and found two with problems. Latest gets those in line. They were both relying on drawing outside of their containing slots. The solution was set better slot dimensions.

This does make me wonder about applying the clipping to art elements though. While they do "belong" to a parent in one sense (we do know the current slot where they're being added, and margins do seem to influence them), they also are almost always absolutely positioned. Anyone have second thoughts on this change in behavior there? (Still seems totally spot-on for text elements...)

jasonrclark:
Strike my comments about art elements from the record... ran it on Shoes3 and got this result:

image

I think this is set to go. Will merge in a couple days unless I hear some objection.

PragTob:
This is weird... I remember that one could always throw around art elements and it'd just be ok. It seems that this is related to whether or not a slot as width/height explicitly set. First example:

```ruby
Shoes.app do
stack do
  para 'omg lol'
  rect 20, 20, 500, 500, fill: green
end
end
```

this branch:

shoes 4 _013

Shoes 3.2:

shoes_014

Shoes 3.1 (it still compiles over here, wow!):

shoes_016


Now when we set an explicit size:

```ruby
Shoes.app do
stack width: 100, height: 100 do
  para 'omg lol'
  rect 20, 20, 500, 500, fill: green
end
end

```

this branch:

shoes 4 _017

Shoes 3.2:

shoes_019

Shoes 3.1:

shoes_021


I.e. behavior does not seem to be right when there is no size set. But that might as well be because the slot sizes are set differently in the versions... let's check that:

```ruby
Shoes.app do
s = stack do
  para 'omg lol'
  rect 20, 20, 500, 500, fill: green
end

button 'report' do
  debug "height: #{s.height}"
  debug "width: #{s.width}"
end
end
```

this branch (Shoes 4):

```
DEBUG: height: 19
DEBUG: width: 600
```

Shoes 3.2:

```
DEBUG: height: 35
DEBUG: width: 600
```

Shoes 3.1:

```
DEBUG: height: 35
DEBUG: width: 600
```

Which seems... weird arbitrary. Or well... it just does the clipping if explicit sizes are set. Which is weird but also somehow understandable... thoughts? @jasonrclark

shoes-swt/lib/shoes/swt/common/resource.rb
```diff
@@ -23,6 +23,18 @@ def set_defaults_on_context(graphics_context)
         def track_graphics_context(graphics_context)
           @graphic_contexts << graphics_context
         end
+
+        def clip_context_to(graphics_context, element)
+          clipping = graphics_context.get_clipping
```
PragTob: can't we do just `clipping` here? :)

shoes-swt/lib/shoes/swt/common/resource.rb
```diff
@@ -23,6 +23,18 @@ def set_defaults_on_context(graphics_context)
         def track_graphics_context(graphics_context)
           @graphic_contexts << graphics_context
         end
+
+        def clip_context_to(graphics_context, element)
```
PragTob: love how this helper works!

PragTob:
So there's still that behavior question to fix up... otherwise I really LOVE how this came together and how relatively little code changes it required :heart:

shoes-swt/lib/shoes/swt/common/resource.rb
jasonrclark: Yep. Too close of reading SWT docs :O

jasonrclark:
Great catch on the implicit vs explicit behavior difference @PragTob! It kind of makes sense to me, so I'll see how hard it'd be to match that.

jasonrclark:
All right, futzed around and got this working in a way that I believe is more consistent (if not 100% like Shoes 3) and satisfying to me.

Here stand the Rules of Clipping (according to Jason, tonight):

1) All non-art elements get clipped to their containing parent dimensions (only implemented so far on `TextBlock`)
2) If a slot dimension was not explicitly fixed, then allow art elements to skip clipping in that direction
3) Otherwise, apply clipping to the explicit slot boundaries

Based on what Shoes 3 did *most* of the time, if you set either `height` or `width` (or both), then that particular dimension will get trimmed for art elements. But the un-set dimensions let the art fly free :balloon:.

One caveat to my implementation is how to determine whether the height/width has been set. I'm kind of cheating by basing it only off the initial height value (which was already used to flag the slot "fixed height" even if the height could be later edited). We also luck into working right on width because it defaults to `1.0` which buys us the desired behavior out of the box.

Tracking for whether `height`/`width` has been set *by the user but not by Shoes positioning code* sounds complicated, not fun, and not that valuable so I'm skipping it (at least for the time being).

That said, will do one final pass through the good samples hunting for any weirdness... oh, and push a commit to fix the red tests that I totally ran before force pushing on my branch here :trollface:

jasonrclark:
Hmm, seems to be some interference with non-art element drawing and the GC clipping :grimacing: 

image

One day I'll finally get this merged...

jasonrclark:
Interesting bug. This related to my recent change to use the clipping height rather than element height for art elements in an unsized slot.

Turns out the `clipping.height` is sometimes a small value. Controls like buttons and textboxes then get triggered for drawing with that value (odd timing, but SWT's does its own thing for controls).

Intent here is really to leave the height unrestricted, so will set it to the app height instead and all is well. :crosses_fingers:

jasonrclark:
Finally looking good on all the samples. Merging so we can get this :bullettrain_front: rolling!

PR #1171, opened 2015-11-23
Load system colors during backend initialization

jasonrclark:
Proposed changes for #1166. **Note this merges onto `system_background`
NOT onto `master`!**

This fixes the defaulting problems we were having and prevents us from
having to do the color loading multiple times (which could happen in the
prior scheme if you had multiple windows.)

Downside, it does remove the defaulting in the DSL layer since there's
not an elegant way to get the value when we need it. Assigning it in the
backend just ends up cleaner.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,6 @@ group :development do
   gem "kramdown"
   gem 'simplecov'
   gem "codeclimate-test-reporter"
-  gem 'jruby-lint'
   gem 'webmock'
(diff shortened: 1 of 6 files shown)


KCErb:
This is great! Thanks so much @jasonrclark!

I was sure it was a problem with the test, I should have spent more time checking that the code did what I expected. :blush: 

Thanks for taking the time to dig into this. So glad you're getting work in on this!

Issue #1057, opened 2015-02-11
Shoes - window always on top


PragTob:
Came in through the mailing list:

> what about setting the 'always on top' state of a shoes window? 

what is meant here is probably that the window is always on top and can not be sent to the background. This seems to be rather simple in SWT (just a style for the shell): [swt thread](http://www.eclipse.org/forums/index.php/t/123254/)

seethemhigh:
This might be not so easy to implement for Linux. There is [this bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=36128) which makes styles other than ``SWT.ON_TOP`` to disappear. I tested it on Ubuntu 14.04 and the app window looks like a square, i.e no close, minimize, maximize buttons, etc.

Not sure about other platforms.

PragTob:
That seems totally weird, that bug is so old (also the duplicated one) that it seems so weird that it isn't fixed in an up to date SWT we use... :/

Since it's an upstream bug I believe all we can do is use it and reference that bug that needs fixing.

AStrangeEnigma:
I would love support for this for a project I'm working on!

AStrangeEnigma:
I'm also happy to try adding it myself if you could give me a little direction

PragTob:
:D 

@AStrangeEnigma that sounds great! My general guidance would be this (and I should probably also write that up some time...): The data if something should be "always on top" should be held in the DSL class `Shoes::App`, the actual implementation of being always on top that is backend specific goes into `Shoes::Swt::App`. And I'd like the option to be passed on startup so that I can do

```ruby
Shoes.app always_on_top: true do
  # ...
end
```

Feel free to open a pull request even with broken tests. We can help you fix them up :)

AStrangeEnigma:
Great!

What I ended up doing was just adding the one line

```style |= ::Swt::SWT::ON_TOP if @dsl.opts[:always_on_top]```

in ```main_window_style```

To get exactly the behavior you mentioned. Is there somewhere else I should add that?

PragTob:
that seems perfectly valid :) Would you care make a PR? Otherwise we'll have to wait until Jason or I get to it :) And you'd miss out on all the glory! :)

AStrangeEnigma:
I'll make the PR tonight!

AStrangeEnigma:
#1172

PR #1172, opened 2015-11-24
Add always on top support to App configuration

AStrangeEnigma:
Please help me figure out the tests because I am lost here! I added a test, but I want to make sure it's ok and that I'm not breaking anything else.

Solves issue #1057 

Thanks,
Richard


jasonrclark:
That test looks just right to me, and fits in neatly with the other tests. You could potentially test the negative case (i.e. `always_on_top = false`), but it is already implicitly covered by the other specs so I'm convinced the conditional is correctly implemented.

Tested it locally and all looks :rocket: 

Did you have any other specific concerns before I get this merged?

AStrangeEnigma:
If you're convinced it's good, then you're all set to merge it!

Thanks for being so helpful!

jasonrclark:
Thanks for the contribution! :sparkles:

Issue #506, opened 2013-12-20
Provide spec guidance


PragTob:
As identified in #505 it would be great to provide further guidance with spec writing. Most likely in the wiki.

E,g, go into stuff like:

* stubbing/mocking
* which specs are executed with what backend
* an example
* tips/tricks

What do you think @chuckremes ? Any additions? =)

Cheers,
Tobi

KCErb:
I've been away from the project for far too long. I think I'll try to jump back in on the last thing I was doing which was some tutorials. I actually left off working through the specs so that fits in nicely with this issue.

jasonrclark:
I think this is done on the basis of https://github.com/shoes/shoes4/wiki/Testing-Shoes

PR #1175, opened 2015-11-29
Updated README to delegate the nitty-gritty to the wiki

jasonrclark:
So I wrote a thing over at https://github.com/shoes/shoes4/wiki/Testing-Shoes that I believe covers the bases for #506

In light of that article, I'm trimming down the README. Will still give the basic "how to run things", but hand off to the wiki for the rest.

How do we all feel about this style? I honestly get a bit bogged down with the README sections right now when I glance over them... maybe something more like this ("here's how to do the most basic thing, here's the rest of it spelled out over there") could help out? What do you all think?

Also keen for any commentary on the wiki entry... although it being a wiki, feel free to just edit in what you like :grin:

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,46 +163,15 @@ Refer to the following section for information on how to run the specs, that wer
 
 ## Running Specs
 
-Shoes 4 is developed in a TDD style. You should be writing and running the specs :)
+Shoes 4 is developed in a TDD style using RSpec. You should be writing and running the specs :)
 
-There are two kinds of Shoes 4 specs:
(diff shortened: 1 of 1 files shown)


PragTob:
I like it, thought about it myself and also heard people saying READMEs should be short and sweet and details should go into the wiki, personally I do like long READMEs though :P

wasnotrice:
thanks @jasonrclark, long overdue. I like a README that's complete but skimmable, and gives me links to further information :star2:

Issue #904, opened 2014-10-03
Hiding a slot hides too much


PragTob:
Take this script:

``` ruby
Shoes.app do
  my_stack = stack do para 'hide me' end
  button 'hide stack' do my_stack.hide end
end
```

For All that I know, the button shouldn't be hidden after clicking it. Just the 'hide me' should go away. Unfortuantely, both go away :-(

![screenshot from 2014-10-03 14 20 44](https://cloud.githubusercontent.com/assets/606517/4504515/365e58f0-4ae7-11e4-8b5b-c7b58c8b93ea.png)

_click_

![screenshot from 2014-10-03 14 20 48](https://cloud.githubusercontent.com/assets/606517/4504516/38217fdc-4ae7-11e4-91f8-820719284aa9.png)

PR #1177, opened 2015-11-29
Only touch contents visibility on actual state change

### LINKED ISSUE #1174: simple-guess-game shows answers when started
To duplicate:

```
bin/shoes samples/simple-guess-game.rb
```

Result:
image

The `hide` call on [this line](https://github.com/shoes/shoes4/blob/2085c9e2b3babb69e13f517a31f398bdf5ed872e/samples/simple-guess-game.rb#L7) ought to work. Not sure why it isn't.

jasonrclark:
Fixes #1174

`update_visibility` gets called plenty of times when nothing has changed, so hiding/showing the children forcibly in those cases overrides their own defaults unnecessarily.

Additionally, this doesn't break the following sample from #904 which is why this whole mechanism was introduced in the first place:

```
Shoes.app do
  my_stack = stack do para 'hide me' end
  button 'hide stack' do my_stack.hide end
end
```

### DIFF
--- a/shoes-swt/lib/shoes/swt/slot.rb
+++ b/shoes-swt/lib/shoes/swt/slot.rb
@@ -25,7 +25,10 @@ class Shoes
       # responsibility, although this is more DSL code
       # #904 #905
       def update_visibility
-        if dsl.hidden?
+        # Only alter contents on a visibility change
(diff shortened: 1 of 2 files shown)


PragTob:
Hm seems good to me and simple enough.

Issue #1183, opened 2015-12-03
:state/.state= broken in edit_line and edit_box


lmrodriguezr:
Hi all,

It looks like the `:state` argument to `edit_line` and `edit_box` is being silently ignored, and the `.state=` method works but disables selection (regardless of the value passed). This little app demonstrates the problem:
```ruby
Shoes.app do
   stack margin_left: 40, margin_right: 20  do
      %w(readonly disabled).each do |s|
         title s
         edit_line("Always enabled", state: s)
         edit_line(:state => s).text = "Always enabled"
         edit_line("Always disabled").state = s
         para ""
      end
      title "This really should be enabled:"
      edit_line("but is not!").state="enabled"
      edit_line("but is not!").state=:enabled
      edit_line("but is not!").state=true
      edit_line("This one is!").state=nil
   end
end
```

The first two methods `edit_line("txt", state: s)` and `edit_line(state: s)` produce regular editable `edit_line`s. The third method `.state=s` seems to ignore the value passed (except if `s.nil?`) and produces greyed-out and un-selectable boxes, which kind of defeats [the original purpose](http://comments.gmane.org/gmane.comp.lib.shoes/2152) of these methods.

In case the problem is specific to my system: I'm running shoes4 out of the box (7fc6c6c5685a6a04ef3d677289c3525e0ad46719) with jruby 9.0.3.0 in Mac 10.10.5.

Thanks!
--Miguel.

PragTob:
Hey there,

thank you for the great bug report with a nice app to reproduce it!

Sorry about that, might be related to one of our style related changes, it should be rather easy to debug I hope :)

Tobi

PragTob:
Official [docs](http://shoesrb.com/manual/Styles.html#:state):

>The :state style is for disabling or locking certain controls, if you don't want them to be edited.
>
>Here are the possible style settings:
>
>    nil - the control is active and editable.
>    "readonly" - the control is active but cannot be edited.
>    "disabled" - the control is not active (grayed out) and cannot be edited.

PragTob:
Hi there,

this should be fixed over in #1186 !

It should now (after it lands on master) only be disabled for "disabled"/:disabled and it also works when passing them right in.

I'll make an issue for the readonly style

Thanks again for the bug report!

PragTob:
That was merged and fixed, will be in pre5 whenw e release it

PR #1186, opened 2015-12-05
Fix enabled state for edit line box

PragTob:
This includes 4 separate changes related to the state of elements, for the changes themselves I'll refer you to the commit messages :)

Fixes (most of) #1183

### DIFF
--- a/shoes-core/lib/shoes/common/state.rb
+++ b/shoes-core/lib/shoes/common/state.rb
@@ -1,15 +1,29 @@
 class Shoes
   module Common
     module State
-      attr_reader :state
+      DISABLED_STATE = "disabled"
+
+      def after_initialize(*_)
+        super
+        update_enabled
+      end
 
       def state=(value)
-        @state = value
-        @gui.enabled value.nil?
+        style(state: value)
+        update_enabled
       end
 
       def state_options(opts)
         self.state = opts[:state]
+      end
+
+      private
+      def enabled?
+        !(state.to_s == DISABLED_STATE)
+      end
+
+      def update_enabled
+        @gui.enabled(enabled?)
       end
     end
   end
--- a/shoes-core/spec/shoes/shared_examples/state.rb
+++ b/shoes-core/spec/shoes/shared_examples/state.rb
@@ -1,7 +1,8 @@
 shared_examples_for "object with state" do
   let(:input_opts) { {state: "disabled"} }
 
-  it "should initialize" do
+  it "should initialize in the right state" do
+    expect_any_instance_of(Shoes.configuration.backend_class(described_class)).to receive(:enabled).with(false)
     expect(subject.state).to eq("disabled")
   end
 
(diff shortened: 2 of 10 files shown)


shoes-core/lib/shoes/common/state.rb
```diff
@@ -1,16 +1,30 @@
 class Shoes
   module Common
     module State
```
PragTob: curiously we had this, but it wasn't included anywhere o_O might have gotten lost in the big style PR or somewhere - another reason against big PRs :D

shoes-swt/lib/shoes/swt/input_box.rb
```diff
@@ -32,7 +32,7 @@ def text=(value)
       end
 
       def enabled(value)
-        @real.enable_widget value
+        @real.enabled = value
```
PragTob: this change is a bit unrelated, I couldn't find any reference to `enableWidget` in the SWT docs so I changed it to the documented enabled

shoes-core/spec/shoes/shared_examples/state.rb
```diff
@@ -1,7 +1,8 @@
 shared_examples_for "object with state" do
   let(:input_opts) { {state: "disabled"} }
 
-  it "should initialize" do
+  it "should initialize in the right state" do
+    expect_any_instance_of(Shoes.configuration.backend_class(described_class)).to receive(:enabled).with(false)
```
jasonrclark: Was this added just to test that the initialization calls through to the backend?

jasonrclark:
Had one question just to clarify about specs, but otherwise this looks :sparkling_heart: Thanks for getting to this @PragTob! I'll get the CHANGELOG bumped once we merge this.

shoes-core/spec/shoes/shared_examples/state.rb
PragTob: Yep, indeed it was. I.e. testing the `after_initialize` bit. Proved its worth, i.e. The one in list_box didn't call super, hence did not call the `State` module method and hence it failed :)

jasonrclark: :+1:

PragTob: I'd have loved for this to be tested on the GUI instance and not as `Shotgun Surgery` on any class, but I couldn't find a suitable way since the gui is constructed in initialize and then the method is straight up called.

There might be a way, now that I think about it again - stubbing the method that creates the gui the return a double. But it should be fine as is imo :)

PragTob:
:) felt great to do some coding again, also this was an easy one. Super happy with how little code I needed to change for these changes and how easy it was.

jasonrclark:
Huzzah! :tada:

shoes-core/lib/shoes/common/state.rb
KCErb: I found where I did this:

https://github.com/shoes/shoes4/pull/803/files#r15267351

 Not sure I understand my reasoning then . . . what we have here seems good :+1:

PragTob: Cool that you dug that up!

I guess the reasoning was most methods are auto generated by style, which is true. But now we needed more stuff to reinstate it - software always changes and it is not uncommon to go back and forth on a decision (at least in my experience) :)

PR #1188, opened 2015-12-06
Update JRuby version

PragTob:
This should finally test with 9k, 1.7 intentionally omitted as we'll likely drop it for the pre6


jasonrclark:
:rocket: :sparkles:

Issue #1148, opened 2015-07-06
Can't package the app: NameError: uninitialized constant Bundler


Le6ow5k1:
When I try to package an app:

```
shoes -p swt:app app.rb
```

I get the following error:

```
 Packaging swt:app...
 NameError: uninitialized constant Bundler
      const_missing at org/jruby/RubyModule.java:2726
```

[full stack trace](https://gist.github.com/Le6ow5k1/fa09e95e6abb5c8b499c)

I'm running it on Ubuntu 14.04 and using rvm with jruby-1.7.19.

It seems like the problem is not on shoes side, but...

cfurrow:
Can confirm this is happening on OSX 10.10.3, java 1.8.0_45, jruby 1.7.19, shoes 4.0.0.pre4

```
$ rbenv local
jruby-1.7.19
$ gem list

*** LOCAL GEMS ***

after_do (0.3.1)
axiom-types (0.1.1)
bundler (1.10.5)
coercible (1.0.0)
descendants_tracker (0.0.4)
equalizer (0.0.9)
ffi (1.9.3 java)
furoshiki (0.3.1)
ice_nine (0.11.1)
jar-dependencies (0.1.2)
jruby-jars (1.7.20.1)
jruby-launcher (1.1.1 java)
jruby-openssl (0.9.5 java)
jruby-rack (1.1.19)
json (1.8.0 java)
krypt (0.0.2)
krypt-core (0.0.2 universal-java)
krypt-provider-jdk (0.0.2)
maven-tools (1.0.7)
nokogiri (1.6.4.1 java)
plist (3.1.0)
rake (10.1.0)
rdoc (4.1.2)
ruby-maven (3.1.1.0.8)
ruby-maven-libs (3.1.1)
rubyzip (1.1.7)
shoes (4.0.0.pre4)
shoes-core (4.0.0.pre4)
shoes-highlighter (1.0.1)
shoes-manual (4.0.0)
shoes-package (4.0.0.pre4)
shoes-swt (4.0.0.pre4)
swt (4.4)
thread_safe (0.3.4 java)
virtus (1.0.4)
warbler (1.4.8)
```

cfurrow:
Got a little further by defining a Gemfile and bundling:

Gemfile:
```
source "https://rubygems.org"

gem 'shoes', '4.0.0.pre4'
gem 'shoes-swt'
gem 'warbler'
```

```
$ bundle
...snip..
[...]

PR #1189, opened 2015-12-06
Always set up Bundler

jasonrclark:
This deals with the first issue in #1148 where folks were getting
constant not found errors on `Bundler`.

We only `Bundler.require` if we find that we have a Gemfile, but parts of
packaging really, really want to have Bundler's code available, so
requiring `bundler/setup` keeps those bits happy.

### DIFF
--- a/shoes-swt/bin/shoes-swt
+++ b/shoes-swt/bin/shoes-swt
@@ -2,8 +2,8 @@
 lib_directory = File.expand_path('../../lib', __FILE__)
 $LOAD_PATH << lib_directory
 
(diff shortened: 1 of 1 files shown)


PragTob:
Cool to merge if green :+1:

jasonrclark:
Turns out this was a very, very bad idea. This makes it so `shoes` won't execute from a directory that doesn't have a Gemfile with shoes in it, which is definitely not what we want.

Will need to just look harder at this to resolve more systematically for #1148 later but gonna revert for `pre5`

PR #1196, opened 2015-12-08
Detect Mac properly on i386

### LINKED ISSUE #1180: Make bash script recognize i386 mac as a mac and not as Linux
.@lmrodriguezr  posted a couple of bash setting he had to have in order to [get shoes running](http://pastebin.com/a0LihAx1)

```
export JRUBY_OPTS='-J-XstartOnFirstThread' # Just to remove annoying warnings
export PATH="/usr/local/Cellar/jruby/9.0.3.0/libexec/bin/:$PATH" # I installed jruby from homebrew
export MACHTYPE=darwin-x86_64 # the default is i386, which makes shoes believe it's a linux and use readlink -f
alias ruby=jruby # I also have ruby installed
alias gem=jgem
```

This issue is about the third one of those, as I think that's the only one we can fix. The first one should be fixed when using the `bin/shoes` script, but maybe detection is broken there as well so it'd be the same problem

jasonrclark:
Fixes #1180

We've gotten reports that `$MACHTYPE` is occasionally `i386`
instead of the more full-formed `x86_64-apple-darwin14` we're counting on.
Include the `$OSTYPE` so we're sure to catch `'darwin'` somewhere in the
string.

### DIFF
--- a/shoes-core/bin/shoes-stub
+++ b/shoes-core/bin/shoes-stub
@@ -42,7 +42,7 @@ mac_readlink_f () {
   echo $RESULT
 }
 
-case "${MACHTYPE:-}" in
+case "${MACHTYPE:-}${OSTYPE:-}" in
   (*darwin*)
     SCRIPT=$(mac_readlink_f $0);;
   (*)


PragTob:
LGTM

PR #1197, opened 2015-12-08
refactoring tests for builtin_methods and adding missing tests

jethroo:
increasing test coverage to 100% for `builtin_methods.rb`

### DIFF
--- a/shoes-core/spec/shoes/builtin_methods_spec.rb
+++ b/shoes-core/spec/shoes/builtin_methods_spec.rb
@@ -4,6 +4,7 @@ describe Shoes::BuiltinMethods do
   let(:input_blk) { Proc.new {} }
   let(:app) { Shoes::App.new({}, &input_blk) }
   let(:logger) { double("logger") }
+  let(:dialog) { double('dialog') }
 
   before :each do
     Shoes::LOG.clear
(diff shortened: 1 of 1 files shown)


shoes-core/spec/shoes/builtin_methods_spec.rb
```diff
@@ -22,63 +22,35 @@
...
-  describe "warn" do
-    before :each do
-      allow(logger).to receive(:warn)
-      app.warn("test")
-    end
+  %w(info debug warn error).each do |level|
+    describe "##{level}" do
+      before :each do
+        allow(logger).to receive(level)
+        app.send(level, "test")
```
PragTob: I'd prefer `public_send` here, otherwise we might make them private and not even notice :)

PragTob:
Thank you!

Looks good to me, I'd just change `send` to `public_send`. 

Other than that I think I'd like slightly less stubs/mocks but that'd rather be hard to pull off I guess and and `Shoes::Dialog` should have appropriate unit tests so it should be fine.

Cheers,
Tobi

shoes-core/spec/shoes/builtin_methods_spec.rb
jethroo: indeed, thx for spotting, will change to public_send :santa:

shoes-core/spec/shoes/builtin_methods_spec.rb
```diff
@@ -106,4 +78,49 @@
       end
     end
   end
+
+  %w(alert confirm).each do |type|
+    describe '##{alert}' do
+     let(:dialog) { double('dialog') }
+     let(:message) { double('message') }
```
jasonrclark: Do we actually need these to be doubles? Seems like a simple string would be sufficient since that's what the actual arguments are passed as.

shoes-core/spec/shoes/builtin_methods_spec.rb
```diff
@@ -106,4 +78,49 @@
...
+    let(:message) { double('message') }
+    let(:args) { double('args') }
+
+    before do
+      allow(Shoes::Dialog).to receive(:new).and_return(dialog)
+      expect(dialog).to receive(:ask).with(message, args).and_return(dialog)
+    end
+
+    it 'creates a new dialog' do
+      expect(app.send(:ask, message, args)).to eq(dialog)
```
jasonrclark: Since we don't have any metaprogramming on the tests, what would you think of just calling `app.ask` here instead of the `send`? It makes sense in the other contexts, but I'd like to avoid it.

shoes-core/spec/shoes/builtin_methods_spec.rb
```diff
@@ -106,4 +78,49 @@
...
+    end
+  end
+
+  describe '#ask_color' do
+    let(:dialog) { double('dialog') }
+    let(:title) { double('title') }
+
+    before do
+      allow(Shoes::Dialog).to receive(:new).and_return(dialog)
```
jasonrclark: Given the repetition on this line, we could probably make a wrapping context for `"dialogs"` that would do this once, then have the separate `describe` for each method use that setup. What do you think?

jasonrclark:
Yeah @PragTob, in my experience with `Shoes::Dialog` the stubbing will be hard to avoid here. Had a few comments on tightening things up and tidying, but otherwise these are looking good. Thanks @jethroo!

shoes-core/spec/shoes/builtin_methods_spec.rb
jethroo: rather then a wrapping context I would like to propose to move this to the top, so the dialog double is only defined once, costwise this should not impact the other test which don't really need the double but will leave a more readable testsuite though (not hiding method describe blocks within the shared context)

shoes-core/spec/shoes/builtin_methods_spec.rb
jethroo: :+1:

shoes-core/spec/shoes/builtin_methods_spec.rb
jasonrclark: Seems reasonable. Any new spec in this file that happens to touch dialogs will get the stub, which is a good thing I think.

shoes-core/spec/shoes/builtin_methods_spec.rb
jethroo: sure, I sometimes prefer doubles since the also can catch changes a normal String wouldn't: so if you pass in `'message'` and someone introduces `message.lowercase` in the implementation, the double will catch that (reminding the implementor to adept the test accordingly), but as said a matter of preference

jethroo:
addressed your comments @PragTob & @jasonrclark

PR #1201, opened 2015-12-10
remove duplication via dynamic method definition in builtin_methods.rb

jethroo:
The logging methods all behaved in the same way, so this PR defines them dynamically thus reducing code duplication.


PragTob:
:+1: thank you, good eyes!

PR #1203, opened 2015-12-10
Fix the positioning of backgrounds along with slots

### LINKED ISSUE #1193: Margins for slot and background broken (samples/good-vjot)
selection_024
(Shoes 4 to the left, 3.2 to the right, if you go back to #255 you see that it used to work)

Might investigate further later on, seems like an annoying breakage :(

PragTob:
* fixes #1193

I'll have another look at it (it irritates me that it tries to do a lookup of extent before it is even positioned) + I probably gotta walk through all samples + the old PR mentioned in #1193 to check out if this really fixes it, but this is how it should work in my head and first examples proclaim that I'm right.

Have a look and review if you like, but please don't merge yet :)

(github file upload seems to be broken... so believe me for now until that's fixed, or run yourself)

### DIFF
--- a/shoes-core/lib/shoes/common/background_element.rb
+++ b/shoes-core/lib/shoes/common/background_element.rb
@@ -5,11 +5,6 @@ class Shoes
         @dimensions = ParentDimensions.new @parent, @style
       end
 
-      # We derive everything from our parent, so we skip slot positioning.
-      def needs_positioning?
-        false
-      end
-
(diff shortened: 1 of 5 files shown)


PragTob:
Should have looked at that old PR, of course this breaks that behavior again... will need to adjust. Sigh.

PragTob:
Deprecated by #1204

PR #1204, opened 2015-12-13
Correctly using parents margin for background

### LINKED ISSUE #1193: Margins for slot and background broken (samples/good-vjot)

PragTob:
This rethinks ParentDimension a bunch, from 3 main observations:

* we should not position backgrounds, it works when they are
  the first element in a slot but not after that
* backgrounds need to respect the margins of parents, which means
  their absolute_starts should be the parents element_starts
* beyond that, there doesn't seem to be much need to get
  parent position values, as they all somehow end up
  influencing absolute_top/left (I hope, this might come back
  later)

That also removes a lot of ambiguity and difficult code.
Better code through increased domain understanding :+1:

* fixes #1193 

succeeds #1203 

A look at a couple of examples between shoes 4 and shoes 3.2:

```ruby
Shoes.app do
  stack margin: 20, height: 100 do
    background green
  end

  stack margin: 40, height: 100 do
    background blue
  end
end
```

selection_031

```ruby
Shoes.app do
  banner 'What up'
  background red
end
```

(see #599)

selection_032

```ruby
Shoes.app do
  stack height: 150, bottom: 50 do
    background red
  end
end
```

(some more advanced positioning)

selection_033

```ruby
Shoes.app do
  stack margin: 20, height: 200 do
    background blue
    background green, margin: 20
  end
end
```

(double margin)

selection_034

last but not least `good-vjot.rb` which still has some issues but those are of another nature (or at least I believe so ;))

good-vjot

### DIFF
--- a/shoes-core/lib/shoes/dimension.rb
+++ b/shoes-core/lib/shoes/dimension.rb
@@ -238,11 +238,7 @@ class Shoes
 
   class ParentDimension < Dimension
     def absolute_start
-      @absolute_start ? super : parent.absolute_start
-    end
-
-    def start
-      @start ? super : parent.start
+      parent.element_start
     end
 
     def extent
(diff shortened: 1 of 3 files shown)


shoes-core/lib/shoes/dimension.rb
```diff
@@ -253,27 +249,18 @@ def extent
...
-      if parent.element_end
-        # Why subtracting an absolute from an element dimension value? A
-        # diagram helped me reason out what we wanted.
-        #
-        # parent.      parent.      self.       self.    parent.      parent.
-        # abs_start    elem_start   abs_start   abs_end  elem_end     abs_end
-        # |   margin   |            |           |        |   margin   |
-        #
-        # To get our extent respecting the parent's margins, it's our absolute
-        # start, minus parent's element end (so we don't blow past the margin)
```
jasonrclark: As the author of this comment, I'm so glad to see it going away :)

jasonrclark:
Code here looks good. I'm going to do some testing on it this afternoon and evening, and assuming everything is happy will merge and :boat: to pre5!

PR #1207, opened 2015-12-13
Correct slot height when margins are in play

### LINKED ISSUE #1205: Margins on elements and effect with parent margins
This is reduced from a good-vjot.rb example, e.g. this one:

good-vjot

The sample can be broken down to:

```ruby
Shoes.app do
  stack margin: 40 do
    background green
    edit_box height: 200, margin_bottom: 20
  end
end
```

selection_035

We can clearly see that for shoes 3.2 the margin_bottom put on the edit_box correctly influences the height of the slot, wheras on shoes4 it doesn't and causes the edit_box to awkwardly extend over the parent slot :sweat:

PragTob:
* elements are positioned inside the parent only, so set
  element_height, not height
* Same thing goes for calculating our height, it's the max
  absolute_bottom minus our element_top :)
* fixes #1205 

Sample from #1205 (ignore the over extending background, that is fixed in #1204 the important thing is the space underneath the edit_box)

selection_039

And finally good-vjot.rb (again the background over extending is fixed in #1204):

good vjot

It is curios/fun for me to see how both these issues in good-vjot.rb came from not fully embracing/adjusting calls to our `element_* dimensions from the legacy back when we made that change, which is sort of a recurring theme :)

### DIFF
--- a/shoes-core/lib/shoes/slot.rb
+++ b/shoes-core/lib/shoes/slot.rb
@@ -279,7 +279,7 @@ class Shoes
 
     def determine_slot_height
       content_height = compute_content_height
-      self.height = content_height if variable_height?
+      self.element_height = content_height if variable_height?
       content_height
     end
 
@@ -291,7 +291,7 @@ class Shoes
                     max
 
       if max_bottom
-        max_bottom - self.absolute_top + NEXT_ELEMENT_OFFSET
+        max_bottom - self.element_top + NEXT_ELEMENT_OFFSET
       else
         0
       end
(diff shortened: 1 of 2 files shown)


jasonrclark:
LGTM. Along with #1204 I'll test things up, merge assuming it's all good and get `pre5` out the :door:

PragTob:
@jasonrclark shouldn't we rather merge the two together and then just check that? Seems like tripling the checking amount otherwise (one time for each, one time for the merge)?

jasonrclark:
@pragtob I might merge first although if anything comes up it'll be extra work them to figure which change caused it. Tradeoffs!

PragTob:
@jasonrclark I see, well good luck then :)

PR #1208, opened 2015-12-14
Revert "Always set up Bundler"

jasonrclark:
Reverts shoes/shoes4#1189

### DIFF
--- a/shoes-swt/bin/shoes-swt
+++ b/shoes-swt/bin/shoes-swt
@@ -2,8 +2,8 @@
 lib_directory = File.expand_path('../../lib', __FILE__)
 $LOAD_PATH << lib_directory
 
-require "bundler/setup"
 if File.exist?("Gemfile")
+  require "bundler/setup"
   Bundler.require
 end
 


PragTob:
.@jasonrclark from over at #1189 (for less jumping):

> Turns out this was a very, very bad idea. This makes it so shoes won't execute from a directory that doesn't have a Gemfile with shoes in it, which is definitely not what we want.
> 
> Will need to just look harder at this to resolve more systematically for #1148 later but gonna revert for pre5

I personally have no idea why that'd cause it (unless we need the bundler code anywhere else, in which case maybe bundler should be an explicit dependency) - but for now Jason deserves some sleep and I go to work :)

jasonrclark:
The problem is that `require 'bundler/setup'` patches things such that Bundler will restricts where code can be required from to just what's in your Gemfile. If you've got a Gemfile, that's exactly what you want. But if you don't--say you're just running the `shoes` executable on a file somewhere--this blocks you from actually loading the rest of shoes!

Not to self-promote too much, but actually have a [lightning talk on how this works](https://www.youtube.com/watch?v=o_VHvwjFcmQ) at pdx.rb a while back. Hopefully explains in a little more detail how that happens :movie_camera:

Issue #1210, opened 2015-12-26
Button::text=() doesn't update the text of buttons


lmrodriguezr:
This little script shows the problem:
```ruby
Shoes.app do
   stack margin: 10 do
      @b = button("Hello") do
        stack margin: 10 do
           para strong("pre:"), @b.text
           @b.text = "Bye"
           para strong("post:"), @b.text
        end
      end
   end
end
```

Which, after clicking the button twice, looks like this:
<img width="250" alt="screen shot 2015-12-25 at 10 54 45 pm" src="https://cloud.githubusercontent.com/assets/12363/12005097/9c85013e-ab5c-11e5-816e-f1e26e4a7142.png">

As you can see, the `text` slot of the object changed, but the button itself still shows the same (original) text.

jasonrclark:
Closed by #1211

PR #1211, opened 2015-12-26
Buttons implement .text=() method

lmrodriguezr:
Solves #1210.

### DIFF
--- a/shoes-swt/lib/shoes/swt/button.rb
+++ b/shoes-swt/lib/shoes/swt/button.rb
@@ -13,7 +13,6 @@ class Shoes
       end
 
       def text=(value)
-        @last_text = @real.text
         @real.text = value
       end
     end
--- a/shoes-swt/lib/shoes/swt/button.rb
+++ b/shoes-swt/lib/shoes/swt/button.rb
@@ -11,6 +11,11 @@ class Shoes
           button.set_text @dsl.text
         end
       end
+
+      def text=(value)
+        @last_text = @real.text
+        @real.text = value
+      end
     end
   end
 end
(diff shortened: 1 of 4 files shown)


shoes-swt/lib/shoes/swt/button.rb
```diff
@@ -11,6 +11,11 @@ def initialize(dsl, app)
           button.set_text @dsl.text
         end
       end
+
+      def text=(value)
+        @last_text = @real.text
```
jasonrclark: Any particular reason you're holding onto this here? I don't see other uses of `@last_text` anywhere else in button-related code.

jasonrclark:
Great catch, and seems like a fairly reasonable to me on first reading through. Appears that we do the same for our handling of input boxes already.

Any chance you could add some basic specs around this? A couple basic examples would be enough to handle it in https://github.com/shoes/shoes4/blob/master/shoes-core/spec/shoes/button_spec.rb and https://github.com/shoes/shoes4/blob/master/shoes-swt/spec/shoes/swt/button_spec.rb

So appreciate all your :bug: reporting and now this PR! :sparkles:

shoes-swt/lib/shoes/swt/button.rb
lmrodriguezr: None at all. I did it based on edit lines, and preferred to be conservative by keeping that line, but I've been sweeping through the code and I think it can be removed. I'll fix that in a PR including specs.

lmrodriguezr:
Shoes made GUI development actually enjoyable for me, it's my pleasure to help with this endeavor!

jasonrclark:
Looks great, thanks @lmrodriguezr. So glad that you're enjoying Shoes!

PR #1214, opened 2016-01-29
Simple App to execute all samples

fidelisrafael:
Hi guys, 

I wanted a way to run each of samples just by pressing a key, so I developed this very simple app. I hope this will be useful for people who want to see what shoes4 can do in a very quickly way ;)

Very simple:

`bin/shoes samples/simple-sample-executor-all.rb`

App Running

Thats it's ;)


PragTob:
Some of the samples will still be broken (see [README](https://github.com/shoes/shoes4/blob/master/samples/README)) but seems like a useful thing to have. Thank you!

Issue #1219, opened 2016-02-26
Improve README


mperham:
> That's the worst README I've ever read. I still don't know what Shoes is. I know how to install it, how to use its samples, and its dependencies. I also (thank God!) know that Shoes 4 is the most recent version. But they never bother to actually say what it is or post a screenshot.

https://www.reddit.com/r/ruby/comments/47oc3n/have_there_been_any_big_apps_made_with_shoes/d0ejrt9

PragTob:
Thanks for bringing this to our attention @mperham :) Definitely a good point... what happens when you been working on a project for way too long :sweat_smile:

PR #1220, opened 2016-02-28
Update README

Gangwolf:
@PragTob Here's a small improvement for the README. #1219


PragTob:
Thank you very much, much appreciated! :D I'll also add some front line example and how it looks like to lure people in, think that this might be a good idea.

Issue #1194, opened 2015-12-07
Manual: First line of code is weirdly indented 


PragTob:
Running the manual and clicking around I noticed that the first line in a code block is always weirdly indented. Could be either a bug in manual code or a bug in our shoes code.

![selection_025](https://cloud.githubusercontent.com/assets/606517/11636744/df31b2f4-9d1d-11e5-9a20-65c825ca1d68.png)

![selection_026](https://cloud.githubusercontent.com/assets/606517/11636746/e0c175c8-9d1d-11e5-92dd-44dd23296ec8.png)

jasonrclark:
Pretty sure that this has been that way for a while, so I don't think it's a recent regression. Marking as TextBlock too which will keep it on my radar longer-haul (since it probably changed when the text flowing work landed a long while back).

jasonrclark:
Looks more like a manual problem to me after than not. We get the odd indent on the first line because of the "right" aligned `"Run this"` link.

Boiling it down to a simpler example, this code yields the following result on both Shoes 3 and Shoes 4:

```
Shoes.app do
  flow do
    para "What", align: "right"
    para "ever..."
  end
end
```

![image](https://cloud.githubusercontent.com/assets/130504/12082744/37ddcefe-b24c-11e5-9b35-d969c6687c1c.png)

As you can see, the following element picks up the right-aligned element's space. Weird, but at least consistent.

This behavior is a problem with how the current manual does positioning via newlines + margins :disappointed: . I've got a small update in https://github.com/shoes/shoes-manual/blob/master/lib/shoes/manual/app.rb#L123-L128 which will get it back looking correct.

What do you think @PragTob? You ok with leaving the behavior (since it's consistent) and fixing the manual directly?

PragTob:
hey @jasonrclark - thanks for looking into this!

I'd actually like to break with Shoes3 for this, because of overall consistency. I.e. another way to do something like this is setting `right: 0` and when we set `right: 0` I believe we do not advance the current position so it'd be weird for me if align does it bit right/left/top/bottom don't.

jasonrclark:
If you're willing to break with consistency, it does seem like a defensible change to make. Will take a look when I've got time. Unfortunately injured my shoulder, so I've been minimizing off-hours computering and will be for a while :(

jasonrclark:
Fixed by https://github.com/shoes/shoes4/pull/1227

PR #1227, opened 2016-03-10
Don't advance position for right aligned text

jasonrclark:
As noted in https://github.com/shoes/shoes4/issues/1194 we get odd effects in flowing if we allow right-aligned elements bump the current position when doing slot alignment.

While this is a breaking change with Shoes 3, it behaves more consistently, and we'll be happier with it in the long haul this way.

image

### DIFF
--- a/shoes-core/lib/shoes/text_block.rb
+++ b/shoes-core/lib/shoes/text_block.rb
@@ -33,6 +33,10 @@ class Shoes
       @gui.in_bounds?(*args)
     end
 
+    def takes_up_space?
+      @style[:align] != "right"
+    end
+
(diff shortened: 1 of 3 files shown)


jasonrclark:
Build failures are from https://github.com/ruby/rake/commit/e76242ce7ef94568399a50b69bda4b723dab7c75, which is incompatible with the latest version of `yard` pending https://github.com/lsegal/yard/pull/946.

We already have some code to be conditional about defining the yard tasks, so I'm extending that to allow for more general failures without blocking things from running.

PragTob:
Looking good, thanks Jason!

PR #1228, opened 2016-03-13
Add basic usage with screenshot for README friendliness

PragTob:
Added one of my favorite demo scripts in there, as well as the samples as it's fairly important that it works :D 

What do you think @jasonrclark ? Also thought about putting some animation sample in as people keep asking me if Shoes can do that, might a be a bit much though.


jasonrclark:
:heart: this.

What do you think about eventually pruning the longer instructions down and linking to wiki with them? I know some folks like having the whole thing in one place, but for something trying to draw beginners in, the deeper details might be a bit intimidating.

PragTob:
@jasonrclark definitely up for cleaning the README and either going to the wiki or another .md file!

PR #1229, opened 2016-03-15
Be able to create a borderless window (by popular demand)

### LINKED ISSUE #1052: Borderless Shoes App
Just in fromt he mailing list:

>I would like to make a shoes app which has a borderless window (i.e. no chrome on the sides/top/bottom, no title bar, no minimize/maximize close buttons). Is this within the realm of what shoes can do? Does someone have a (platform-independent) way of doing this?

PragTob:
* fixes #1052

Not too sure about calling the option `borderless` - thoughts @jasonrclark ?

```ruby
Shoes.app borderless: true, width: 400, height: 287 do
  image 'http://moonmoonwolf.weebly.com/uploads/2/5/5/8/25581328/2095238.jpg', width: 400, height: 287
end
```

selection_004

### DIFF
--- a/samples/README
+++ b/samples/README
@@ -12,6 +12,7 @@ simple-anim-shapes.rb
 simple-animate.rb
 simple-anim-text.rb
 # simple-arc.rb  Pending on #423
+simple-borderless.rb
 simple-bounce.rb
 simple-brightness-transitions.rb
 simple-button-animate.rb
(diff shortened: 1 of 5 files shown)


jasonrclark:
Border definitely seems like the right term, but it's often clearer to avoid negatives in setting names. Could we have border: true as the default, so you'd pass border: false on the window to disable?

Hoping to get a moment to test on my Mac soon. Maybe tonight...

PragTob:
Sounds like a good idea, will adjust! :+1:

PragTob:
Let me know what you think :)

jasonrclark:
Looking good on Mac. Here comes the merge! :dancer: @PragTob

PR #1230, opened 2016-03-15
Lighten the load on Travis a bit

PragTob:
* one JDK7 test should be enough (given that JRuby and SWT
  do their testing their already)
* for a pre release project worrying about that bit of backwards
  incompatibility from 9.0.5.0 to 9.0.4.0 that JRuby might
  introduce seems like a bit overkill in hindsight


jasonrclark:
Totally sensible to me.

PR #1231, opened 2016-03-25
Tighten versioning requirements to avoid breakages

PragTob:
* Noticed that if I were to release an after_do version
  0.x.y that breaks compatibility people would install
  it when gem installing shoes. That might happen so
  I rather tightened the dependency (along with the
  one of the swt gem)
* Also updated after_do to newest version


jasonrclark:
Legit. Feel free to merge on :green_heart: from my perspective!

PragTob:
Thanks for the quick review, I'm waiting for the last green knob then and will merge then :)

PR #1234, opened 2016-05-11
Jruby 9.1

PragTob:


PragTob:
Seems like there is some dependency that starts failing with old JRuby 1.7 (Ruby equivalent version less than 2.0) so question is to drop 1.7.x support or make version restrictions.

```
Gem::InstallError: ruby_dep requires Ruby version >= 2.0.0, ~> 2.0.

Using rspec-core 3.4.4

Using rspec-expectations 3.4.0

Using rspec-mocks 3.4.1

Installing rubocop 0.40.0

Installing shoes-manual 4.0.0

Installing webmock 2.0.1

Installing codeclimate-test-reporter 0.5.0

Installing furoshiki 0.3.1

Installing pry 0.10.3

An error occurred while installing ruby_dep (1.3.1), and Bundler cannot

continue.

Make sure that `gem install ruby_dep -v '1.3.1'` succeeds before bundling.

```

jasonrclark:
Which of our dependencies actually is pulling in `ruby_dep`? While I'm longer-term with cutting over to 9.x, I'm not sure if I'm ready quite yet to make the cut. Open to discussion, though...

PragTob:
Pinned listen (required by guard) down to a version that doesn't break on 1.7 :)

PR #1236, opened 2016-05-28
Add require 'shoes' to the welcome script in README.md

rubycoder:
Trivial change, but without the require 'shoes', JRuby gives the error:
NameError: uninitialized constant Shoes
which could be confusing to someone new to Ruby.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ Shoes is a little DSL for cross-platform (Mac, Windows, and Linux) GUI programmi
 Want to see what Shoes looks like? Well, here you go! Given the script:
 
 ```ruby
+require 'shoes'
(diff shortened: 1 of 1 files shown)


PragTob:
Hi there, thanks for your contribution. The program is meant to be run with the `shoes` executable, probably we should be more explicit about that :)

rubycoder:
Sorry, my bad!   I kinda skimmed README.md and didn't read it closely
enough. It does say about running with the shoes executable, and even
mentions the option of running with require 'shoes', so I think what you
have is OK.

Thanks,

Wayne

PragTob:
Okay cool, thanks :)

PR #1237, opened 2016-05-31
"Run" button in expert-game-of-life sample now correctly toggles between "Run" and "Stop"

rubycoder:
Looking at the source for expert-game-of-life.rb, the intent was that when the "Run" button was clicked, its title would change to "Stop", and when "Stop" was clicked, its title would change back to "Run".  However, this didn't work, and the "Run" button always had the title "Run".  This changes and simplifies the implementation so that it now correctly toggles between "Run" and "Stop".

Corrected a typo, "Glinder" -> "Glider".

Minor code cleanups.

Tested on Windows 7 Professional SP 1 and Ubuntu 15.10.

Full Disclosure: The para "Species" usually appears above the "Glider" button.  But occasionally "Species" does not appear.  This happens sporadically, and I can't recreate the problem except by trial and error.  I'm reasonably certain this problem occurred before I made any changes to the code, and I don't see how my changes could have caused this, so I don't think I introduced this problem.

If these changes look OK, I'll make the corresponding changes to expert-game-of-life-adjusted.rb.

### DIFF
--- a/samples/expert-game-of-life.rb
+++ b/samples/expert-game-of-life.rb
@@ -165,47 +165,53 @@ class World
     result.compact
   end
 
-  def add_specie(*cells)
+  def add_species(*cells)
     cells.each do |y,x|
       @board[y][x].toggle_state
     end
   end
 
-  def add_glinder
-    add_specie [4,2], [4,3], [4,4], [3,4], [2,3]
+  def add_glider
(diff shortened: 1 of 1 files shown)


PragTob:
LGTM, thank you very much :)

PR #1243, opened 2016-06-09
Resolve 8 RuboCop issues

rubycoder:
The remaining RuboCop issues are ones where, IMHO, we arguably don't want to change the code simply to avoid a RuboCop message.

```
shoes/app.rb:60:7: C: Performance/RedundantMerge: Use options[:owner] = self instead of options.merge! owner: self.
      options.merge! owner: self
      ^^^^^^^^^^^^^^^^^^^^^^^^^^
shoes/slot_contents.rb:27:7: C: Performance/RedundantBlockCall: Use yield instead of blk.call.
      blk.call
      ^^^^^^^^
shoes/dsl.rb:586:7: C: Performance/RedundantBlockCall: Use yield instead of blk.call.
      blk.call if blk
      ^^^^^^^^
shoes/slot.rb:288:21: C: Style/MultilineMethodCallIndentation: Align select with contents. on line 287.
                    select(&:takes_up_space?).
                    ^^^^^^
shoes/slot.rb:289:21: C: Style/MultilineMethodCallIndentation: Align reject with contents. on line 287.
                    reject(&:hidden?).
                    ^^^^^^
shoes/slot.rb:290:21: C: Style/MultilineMethodCallIndentation: Align map with contents. on line 287.
                    map(&:absolute_bottom).
                    ^^^
shoes/slot.rb:291:21: C: Style/MultilineMethodCallIndentation: Align max with contents. on line 287.
                    max
                    ^^^
shoes/text_block.rb:132:21: C: Style/MultilineMethodCallIndentation: Align .split with .gsub on line 131.
                    .split(',').map { |x| x.strip.gsub(/["]/, '') }
                    ^^^^^^
```


shoes-core/lib/shoes/dsl.rb
```diff
@@ -583,7 +583,7 @@ def keyrelease(&blk)
     end
 
     def append(&blk)
-      blk.call if blk
+      yield if blk
```
PragTob: if we change this we could/should remove the `&blk` and instead check `block_given?` - this avoids the capture of the whole block and therefore also gives a small speed boost :)

shoes-core/lib/shoes/slot_contents.rb
```diff
@@ -24,7 +24,7 @@ def add_element(element)
     def prepend(&blk)
       @prepending_index = 0
       @prepending = true
-      blk.call
+      yield
```
PragTob: same about not capturing the blk in the method definition imo

PragTob:
Thank you, looks good! I have a couple of further optimizations/additions but then it should be good to go :)

rubycoder:
Good catch!  ```block_given?``` did cross my mind as I was making this change, but I didn't pursue it, mostly because I'm not familiar with the Ruby style conventions in this code base so I wanted to make the most conservative change that would address the immediate issue, which was getting RuboCop to stop complaining about this line.

Let me know if there's anything you'd like me to do on this pull request.

PragTob:
@rubycoder as said, if you'd change it to remove the blk argument and use block_given? that'd be fantastic otherwise I'll do that :)

rubycoder:
Sounds good!  Thanks.

Done!

PragTob:
Thanks!

PR #1244, opened 2016-06-16
Fix several issues in simple-calc-2 sample

rubycoder:
In any physical, real-world calculator I can recall using, if I wanted to compute 2+2 and 3+3 I would click the following buttons:

2 + 2 = 3 + 3 =

But if I do that in this program,

2 + 2 =

shows "4", as expected, but then clicking 3 changes the display to "43".

I appreciate that the sample is intended as an example of using Shoes, not an example of how to write calculator logic.  But this problem makes it hard to actually use the calculator.

Another fix is that on most calculators, doing:

1 + 2 + 3 + 4 + 5 + 6 + 7 + 8 + 9 =

shows the running total every time + is clicked.  This program did not display any result until = was clicked.

Renamed several instance variables to be more precise.

Corrected 4 RuboCop issues.

If we'd like, I can make the corresponding changes to simple-calc.rb, also.

### DIFF
--- a/samples/simple-calc-2.rb
+++ b/samples/simple-calc-2.rb
@@ -1,31 +1,45 @@
 Shoes.app height: 250, width: 198 do
-  def do_calc
-    @number = @previous.send(@op, @number)  if @op
+  def clear
+    @accumulator = 0
+    @display = 0
     @op = nil
   end
 
-  @previous, @number, @op = 0, 0, nil
+  def do_calc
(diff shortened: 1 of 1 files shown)


PragTob:
LGTM. I don't we need to make the changes to simple-calc.rb as ideally we'd only have one of them lying around either way and we'd remove the other. There are soemtiems 2 versionsw because of different code bases and/or features not yet implemented that we needed to work around but still get it to work.

Thanks!

PR #1246, opened 2016-09-10
Renamed samplexx.rb files appropriately (Issue #353).

rodenmonte:
Fixed minor Rubocop Issues (Issue #945)


rodenmonte:
"rubocop-0.42.0 requires ruby version >= 2.0.0, which is incompatible with the current version, ruby 1.9.3p551 (jruby 1.7.25)"
and
"/home/travis/build.sh: line 179: bundle: command not found"
are the two error messages I receive. I didn't touch anything that should affect what version of rubocop runs in my commit, I don't think I broke anything that is. Perhaps the project lost backwards compatibility somewhere else, maybe its a problem on Travis's end?

samples/simple-bounce.rb
```diff
@@ -1,24 +1,31 @@
...
-    ydir *= -1 if y > self.height - size[1] or y < 0
+      xdir *= -1 if x > 300 - size[0] or x < 0
+      ydir *= -1 if y > 300 - size[1] or y < 0
 
-    @icon.move x.to_i, y.to_i
+      img.move x.to_i, y.to_i
+    end
+    pause = n
   end
 end
```
PragTob: Seems to me that you accidentally overrode somple-bounce.rb with some other sample?

PragTob:
First thanks for the PR @rodenmonte !

You did nothing wrong, shoes has been a bit idle and in the meantime new rubocop versions were released, Travis picked that up installed a new version and there was a deprecation. So it's our thing to fix :D 

Looks good overall, thanks for making the PR and naming the samples. From my point of view there is 2 issues here left to be resolved:

- [ ] simple-bounce.rb seemed to have been overridden - would be cool if you could restore the vintage version and rename the other (or tell me what I missed :) )
- [ ] fix the build, which is more my thing to do on master

PragTob:
Trusting this works now that .travis.yml is adjusted on master, thanks :)

PR #1247, opened 2016-09-11
Upgrade jruby and drop 1.7

PragTob:
I think it's time as rubocop was breaking in #1246 and this is just going to get more and more as we go.

Maybe @jasonrclark would take a look? :)


PragTob:
Dropped the updated listen as it started to fail on appveyor and I dunno if/when I'd get them to install a new JRuby version or get it to work with installing myself (which failed before). 

Might also want to move listen into some dev group not installed on the CI.

README.md
```diff
@@ -41,13 +41,15 @@ Hi there, thanks for checking by! Shoes 4 is in the preview stage. It currently
 
 There are two ways to get your hands on Shoes 4 - the preview gem release and installing it straight from github. For both you need a current [JRuby](http://www.jruby.org/) installation.
 
+As a not on versions, currently 1.7.x should still work (along with JDK 6) but we removed testing against it as dependencies started acting up and JRuby 9k has been out fairly long and the upgrade is relatively painless. **We recommend using JRuby 9+**.
```
jasonrclark: The first phrase doesn't make sense to me, but I think we could drop it anyway. Just start with "Currently 1.7.x should still work..."

jasonrclark:
Minor README tweak, but otherwise this looks great to me. Totally happy to make the bump on this now. Thanks @PragTob !

README.md
PragTob: haha that was supposed to read "note" - will just throw it away, thanks :)

PR #1248, opened 2016-09-12
Re-added samples/simple-bounce, accidentally removed in #1246

rodenmonte:


PragTob:
This is the same as #1246 right? Then I'll close here

PR #1253, opened 2016-11-04
Partial fix for issue shoes4/issues/945

brennx0r:
Diving into rubocop fixes. I focused mainly on this error:

http://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/MutableConstant

There was one other type of Rubocop error I fixed (instead of gsub use delete method). See changes in ```tasks/changelog.rb``` for details.


brennx0r:
Hmm, not sure why there's jruby specific errors. I didn't add or modify anything regarding jruby, as far as I know.

brennx0r:
Hmm, I think the cause of the error is the application of the frozen method in the Rakefile. Removing the method now.

brennx0r:
Welp, the failure is in a different place now, which is good. But, it does mean that additional work is needed on the PR. I'm going to head to bed, but I plan on continuing work on this PR tomorrow when time permits. Thanks for your patience! :)

jasonrclark:
Hey @brennx0r thanks for looking into this. Glancing over the output, it seems like we might be doing something bad modifying base styles in the code. I don't understand it yet myself (and can't look much further at the moment), but you might have uncovered a bug with this!

PR #1254, opened 2016-11-04
Rubocop, my old friend

jaredmoody:
Wow, a year and a half between rubocop commits.

I regenerated `.rubocop-todo.yml` with the newest rubocop.

Unfortunately rubocop has some bugs in how it handles exclusions in the todo file, causing not all cops in the todo file to be properly ignored. The other fixes in this PR address those cops to make all rubocop checks pass again.

Also [filed a bug](https://github.com/bbatsov/rubocop/issues/3699) with rubocop for the false positive it's throwing on `shoes-core/lib/shoes/app.rb` for the `Style/VariableNumber` cop and temporarily added the exclusion to `.rubocop.yml`

#945


PragTob:
Yeah shoes hasn't been very active recently, but we might pick it up again :muscle: 

Thanks for the help! :)

PR #1257, opened 2016-11-05
Remove listen restriction as we don't support 1.7 anymore

PragTob:


PragTob:
GRML JRuby 9k still on appveyor which is why it won't pass there (although I don't completely understand why...) that ruby_dep thing is weird ;P

jasonrclark:
Ugh, `ruby_dep` again :(

Looking on Rubygems, looks like 1.3.1 of that gem is where it last supported Ruby 2.0+. Later versions are only for Ruby 2.2.5, and JRuby 9.0.5.0 is pinned at 2.2.3 and doesn't seem likely to ever change.

If we're going all-in on JRuby 9k do we want to just say 9.1.5.0 is the thing and roll forward? Would cut down our testing times as well, and if you're going 9k, not much reason to hang back on 9.0 vs 9.1.

jasonrclark:
Another plus on going 9.1 only... Travis times would be sliced considerably. My last build too 50minutes to get all through, but only running on 9.1 and head should cut that down a lot.

PragTob:
yeah really starting to dislike `ruby_dep`...

seems like `listen` depends on `ruby_dep` in `~> 1.2` so we could restrict it to that I guess... :|

PragTob:
argh of course appveyor fails cause they only got 9.0... they installed it back in the day just for me. Wonder if one could get them to install a new JRuby version, however guessing that I'm the only (unpaying) person who uses it....

PragTob:
with a newer JRuby on appveyor, this might actually pass :)

jasonrclark:
Did you want to rebase to see if it'll go? I think we'll still have problems with 9.0.x on Travis because of the `ruby_dep` versioning problem, though. Now that Appveyor is on 9.1.x, do we want to just write off 9.0.x entirely?

PragTob:
Well yeah, let's do that. You suggested it earlier here and it might be the most pragmatic way forward - especially considering we don't even have a "REAL" release it's reasonable to use the latest JRuby.

jasonrclark:
🚢

PR #1260, opened 2016-11-07
Wanted: Rubocop Fixes, Dead or Alive

jaredmoody:
More rubocop autocorrected fixes (with the occasional manual adjustment) for #945


PragTob:
Still sort of proud that most of them are in the samples, which are random scripts random people wrote ages ago :D LGTM leaving it open a bit so that any of the others could chime in and offer objections to the styles and/or praise :)

jasonrclark:
Works for me. Will see if the trailing if/unless rule gets to me as we go forward ;)

Thanks again! :sparkles:

PR #1262, opened 2016-11-08
Rubocop: Looking for me?

jaredmoody:
A collection of rubocop autocorrected syle fixes for #945


PragTob:
the each with object bit was a bit weird but it seems to be something [important to bbastov](http://batsov.com/articles/2013/12/04/using-rubys-each-with-object/) and seems ok although we should now also remove the object return in the end - but I'll let it through as is as not too important :)

Thanks!

Issue #1212, opened 2015-12-27
visit doesn't remove clickable objects


lmrodriguezr:
This is a very strange bug, that I've noticed has certain... selectivity. It doesn't seem to occur with all clickable objects, but it turns out that some objects are persistent (and clickable) after the window has been cleared out by `visit`.

I've prepared a script that showcases the bug:

``` ruby
class BadBoxes < Shoes
   url "/",     :index
   url "/(.+)", :color_rand

   def index
      stack(margin:30) do
         stack do
            background darkblue
            title "Testing blue...", stroke: white, margin: 30
         end.click{ visit "/blue" }
      end
      stack(margin:30) do
         $s = stack do
            background darkred
            title "Testing red...", stroke: white, margin: 30
         end
         $s.click{ visit "/red" }
      end
      stack(margin:10) do
         para link(strong("Green"), click: "/green")
         para strong(link("Green", click: "/green"))
         para link("Green", click: "/green")
      end
   end

   def color_rand(col)
      para.parent.hidden = false
      para link("Home", click: "/")
      stack(margin:40){ para rand(1e3), stroke: eval(col) }
   end
end

Shoes.app
```

If you run this script and click in any of the big boxes, you'll notice (unless it's something specific to my system) that the cursor keeps switching to clickable (the little hand) whenever you hover the place where the boxes used to be. Not only that, if you click there you'll see that the click event is actually fired (I've put a random number, so you can tell when the page is reloaded).

The three "Green" links show something interesting. Here's the relevant piece of code:

``` ruby
         para link(strong("Green"), click: "/green")
         para strong(link("Green", click: "/green"))
         para link("Green", click: "/green")
```

The first and the third links go away nicely when any `visit` is fired (as they should). However, the second link stays to haunt the `color_rand` page. I know that `strong(link(...))` is not proper syntax, and it's not actually rendered as `strong`, but that's actually how I found out about this bug and it might give some extra clues.

lmrodriguezr:
By the way, I've found a workaround to this problem, but it's rather cumbersome: hide clickable objects. Here's a very manual way to do it just to showcase the workaround:
```ruby
class BadBoxes < Shoes
   url "/",     :index
   url "/(.+)", :color_rand

   def index
      stack(margin:30) do
         ($b = stack do
            background darkblue
            title "Testing blue...", stroke: white, margin: 30
         end).click{ visit "/blue" }
      end
      stack(margin:30) do
         $r = stack do
            background darkred
[...]

PR #1267, opened 2016-11-15
Clean up slot click listeners

jasonrclark:
When an outer slot cleared an inner slot, it ended up calling `clear`
instead of `remove` on the inner slot. This in turn didn't remove the
click listeners properly, which left them dangling.

Because clearing a slot shouldn't remove clicks, but when your parent
clears you should, this meant that slots needed to have a separate
`remove` method. Wonderfully, this lined up the handling around slots
during clear and got rid of a stray conditional.

Sample program to expose the issue:

```
Shoes.app do
  @outer = stack do
    @inner = stack do
      background darkblue
      title "Testing blue...", stroke: white, margin: 30
    end

    @inner.click do
      puts "whatever"
    end
  end

  button "clear" do
    @outer.clear
  end
end
```

Partial fix for #1212, which has issues with links as well before we're done.

### DIFF
--- a/shoes-core/lib/shoes/mock/slot.rb
+++ b/shoes-core/lib/shoes/mock/slot.rb
@@ -8,6 +8,9 @@ class Shoes
         @dsl, @parent = dsl, parent
       end
 
+      def remove
+      end
+
       attr_reader :dsl
     end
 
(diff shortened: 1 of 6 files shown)


PragTob:
:+1: :heart:

PR #1269, opened 2016-11-15
Rubocop: Serve the public trust, protect the innocent, uphold the law.

jaredmoody:
A collection of rubocop autocorrected style fixes for #945


samples/good-arc.rb
```diff
@@ -26,8 +26,8 @@
       10.times do |j|
         strokewidth sw * j
         size = (basesize + step * j) / 3
-        top = (self.height - size) / 2
-        left = (self.width - size) / 2
+        top = (height - size) / 2
+        left = (width - size) / 2
```
jasonrclark: Does this sample still run properly? I'm generally in favor of excluding self where possible, but I remember some samples hit problems because of our fussing with self inside the blocks.

samples/original-manual-adjusted.rb
```diff
@@ -30,7 +30,7 @@ def dewikify_hi(str, terms, intro = false)
       str = str
         .gsub(CODE_RE) { |x| code << x; "CODE#[#{code.length-1}]" }
         .gsub(/#{Regexp.quote(terms)}/i, '@\0@')
-        .gsub(/CODE#\[(\d+)\]/) { code[$1.to_i] }
+        .gsub(/CODE#\[(\d+)\]/) { code[Regexp.last_match(1).to_i] }
```
jasonrclark: While not a huge fan of $1, this seems worse to me. If we want to revise to it, I'd prefer to explicitly call match and use the return vs Regexp.last_match. @PragTobi what do you think?

PragTob: you know what I think :D aka yes revise and explicitly use match or some other form that doesn't have to rely on global state

samples/good-arc.rb
jaredmoody: Ran this and it looks good to me. Also ran a few of the other samples modified in this commit and they all seem ok.

samples/original-manual-adjusted.rb
jaredmoody: Hmm, some of these seem ok with `match`, but the if/case statements don't feel good.

e.g.:

```ruby
match_data = ps.match(CODE_RE)
paras.map do |ps|
  if match_data
    yield :code, match_data[1]
  else
    case ps
    when match_data = ps.match(/\A \* (.+)/m)
          yield :list, match_data[1].split(/^ \* /)
```

What do you guys think? Any suggestions on how this could be better written? Or perhaps the cure here is worse than the disease?

jasonrclark: Yeah, I don't see a tidy way to keep the `case` with that... but I do have a solution. We don't even include `original-manual.rb` or `original-manual-adjusted.rb` in the working samples list, and we've extract the "real" manual over to its own gem in https://github.com/shoes/shoes-manual. How about we just get rid of these two files entirely and adjust the other instances to use `match`?

jaredmoody: Sounds fine to me 👍

jaredmoody:
OK, removed original shoes-manual files and used `match` in one case and the return value of `gsub!` in another to remove global back references.

jasonrclark:
Will merge this once we're 🍏

jaredmoody:
And then I'll send you another PR 😄 

3 more of 10 fixes each and then all the autocorrectable cops are done 🎉 👟 🎉

PR #1270, opened 2016-11-18
Clarify our recommendation about JRuby 9k

### LINKED ISSUE #1179: Ditch JRuby 1.7 in favor of 9k? (also means dropping 1.9 in favor of 2.x)
1.7 still receives maintenance releases but 9k is the way forward. If we dropped it we could make the README clearer as we're only talking about one JRuby version and its requirements then. Moreover, we could use shiny new Ruby 2 features.

It'd require shoes users to have a JDK 7+ though.

I'm gonna leave this open for a while to gather a wide range of feedback. :)

jasonrclark:
Fixes #1179 since we've already dropped the testing of it from Travis.


PragTob:
:+1:

PR #1272, opened 2016-11-18
Find nested links properly

### LINKED ISSUE #1212: visit doesn't remove clickable objects

jasonrclark:
When we're doing operations like removing link handlers, we need to
traverse the potentially multiple levels of a text block defined by
calling things like `strong(strong(link()))`. These were only looking at
the top-most level.

Fixes #1212 other case that we didn't get with previous slot work.


shoes-core/lib/shoes/link.rb
```diff
@@ -62,6 +62,17 @@ def visible?
...
 
+    def links
+      [self]
+    end
+
+    def self.find_links(texts)
+      texts.to_a
+           .map { |text| text.links if text.respond_to?(:links) }
+           .flatten
+           .compact
```
PragTob: maybe select responds_to links first? That way we'd get rid of the if and most likely the compact as well?

shoes-core/lib/shoes/text.rb
```diff
@@ -17,6 +17,10 @@ def app
       parent && parent.app
     end
 
+    def links
+      Shoes::Link.find_links(@texts)
+    end
```
PragTob: not 100% sure about using a singleton method to find the links... e.g. that is publicly accessible but doesn't make much sense outside of text/text_block. I'd have probably gone with a module to include... that said, we could also put it into its own singleton or sort of helper object `LinkFinder`... not leaning strong in any way - any ideas?

jasonrclark: Yeah, the fact that we needed similar finding between the `TextBlock` (full Shoes element) and the `Text` "segment" classes which aren't real Shoes elements, makes this awkward.

We've definitely stuck more to the module based approach in Shoes, so although I'm not 100% a fan, that'd probably be more expected. I'll give it a try.

PR #1273, opened 2016-11-18
Let link hover define own target

### LINKED ISSUE #1266: Exception when hovering on links nested in spans
Found while looking into #1212. The following app:

```
Shoes.app do
  para(strong(@l = link("yo") { puts "yo" }))

  @l.hover do
    print "."
  end
end
```

Results in the following when you hover over the link:

```
NoMethodError: undefined method `eval_hover_block' for (Shoes::Span:0x000012b8 "yo"):Shoes::Span
                 mouse_hovered at /Users/jason/source/shoes4/shoes-core/lib/shoes/common/hover.rb:43
  block in mouse_hover_control at /Users/jason/source/shoes4/shoes-swt/lib/shoes/swt/mouse_move_listener.rb:44
                          each at org/jruby/RubyArray.java:1734
           mouse_hover_control at /Users/jason/source/shoes4/shoes-swt/lib/shoes/swt/mouse_move_listener.rb:42
                     mouseMove at /Users/jason/source/shoes4/shoes-swt/lib/shoes/swt/mouse_move_listener.rb:15
                run_event_loop at /Users/jason/.rbenv/versions/jruby-9.1.6.0/lib/ruby/gems/shared/gems/swt-4.4/lib/swt/minimal.rb:109
                    event_loop at /Users/jason/.rbenv/versions/jruby-9.1.6.0/lib/ruby/gems/shared/gems/swt-4.4/lib/swt/minimal.rb:92
                          open at /Users/jason/source/shoes4/shoes-swt/lib/shoes/swt/app.rb:49
                      open_gui at /Users/jason/source/shoes4/shoes-core/lib/shoes/internal_app.rb:120
                    initialize at /Users/jason/source/shoes4/shoes-core/lib/shoes/app.rb:50
                           app at /Users/jason/source/shoes4/shoes-core/lib/shoes/app.rb:19
                        <main> at /Users/jason/source/shoes4/testing/nested_link.rb:1
                          load at org/jruby/RubyKernel.java:977
                        (root) at /Users/jason/source/shoes4/shoes-core/lib/shoes/ui/cli.rb:1
                   execute_app at /Users/jason/source/shoes4/shoes-core/lib/shoes/ui/cli.rb:58
                        <main> at /Users/jason/source/shoes4/shoes-core/lib/shoes/ui/cli.rb:72
                          load at org/jruby/RubyKernel.java:977
                        <main> at /Users/jason/source/shoes4/shoes-swt/bin/shoes-swt:11
```

Note that if you take out the `strong` call around the `link`, things are happy.

jasonrclark:
Fixes #1266

We were assuming that the immediate parent of a link would be the proper
hover handler (which needs dimensioning information, and thus needs to
be the TextBlock). But in fact, you can nest links within other span
types, so the single-parent lookup we were doing was insufficient.

### DIFF
--- a/shoes-core/lib/shoes/link.rb
+++ b/shoes-core/lib/shoes/link.rb
@@ -18,9 +18,9 @@ class Shoes
       super texts, @style
     end
 
-    # Force hovering evaluation up in parent where we have actual dimensions
-    def eval_in_parent?
-      true
+    # Hover must use containing text block since we're missing dimensions
+    def target
+      text_block
     end
 
     # Doesn't use Common::Clickable because of URL flavor option clicks
--- a/shoes-core/lib/shoes/common/hover.rb
+++ b/shoes-core/lib/shoes/common/hover.rb
@@ -23,10 +23,6 @@ class Shoes
         @hovered
       end
 
-      def eval_in_parent?
-        false
-      end
-
       def hover_class
         return @hover_class if @hover_class
 
@@ -57,9 +53,7 @@ class Shoes
       end
 
       def target
-        target = self
-        target = parent if eval_in_parent?
-        target
+        self
       end
 
       def eval_hover_block(blk)
--- a/shoes-core/spec/shoes/link_spec.rb
+++ b/shoes-core/spec/shoes/link_spec.rb
@@ -9,10 +9,12 @@ describe Shoes::Link do
   let(:parent) { double("parent") }
   let(:internal_app) { double("internal app", app: app, gui: gui, style: {}, element_styles: {}) }
   let(:texts) { ["text", "goes", "first"] }
+  let(:text_block) { double 'text block', visible?: true, hidden?: false }
 
   subject {
     link = Shoes::Link.new(app, texts, color: :blue)
     link.parent = parent
+    link.text_block = text_block
     link
   }
 
@@ -21,7 +23,7 @@ describe Shoes::Link do
       element_styles[clazz] = styles
     end
 
-    allow(parent).to receive(:eval_hover_block) do |blk|
+    allow(text_block).to receive(:eval_hover_block) do |blk|
       blk.call(subject) if blk
     end
   end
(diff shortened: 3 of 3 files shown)


shoes-core/lib/shoes/link.rb
```diff
@@ -18,9 +18,9 @@ def initialize(my_app, texts, styles = {}, blk = nil)
       super texts, @style
     end
 
-    # Force hovering evaluation up in parent where we have actual dimensions
-    def eval_in_parent?
-      true
+    # Hover must use containing text block since we're missing dimensions
+    def target
+      text_block
     end
```
PragTob: won't that be a problem though? I.e. don't we need dimensions? Which, I know, is  a grueling thought but if we have a text block that goes link, no link, link won't we trigger hover for all of them? Not in this PR as it's a fix but in general and in the future?

PragTob:
LGTM major question about dimensions though :)

shoes-core/lib/shoes/link.rb
jasonrclark: Tested this out, and the hover detection talks to the elements directly. The only thing the `target` is used against is the evaluation of the hover block. I just cribbed the dimensions comment from my earlier work, none of which explains when we need dimensions :(

That said, when I think about the odd nature of the `Link` and `Span` objects... they're not really Shoes elements, so having them be the `self` in the block evaluation might be surprising in a different sense.

I'm inclined to revise the comment but leave things as they are, or maybe check Shoes 3 to see what it has as the hover block evaluation target.

jasonrclark:
Upon further testing and reflection, the evaluation doesn't care about the target at all either so I tried removing it and everything remained happy. Tested out on the manual, which makes heavy use of links, and I'm pleased to say it worked fine.

Will keep an eye out, but happy to see this simplified. Thanks for the questioning on it @PragTob!

PragTob:
Then this seems cool to go :)

PR #1274, opened 2016-11-18
Rubocop: Come quietly or there will be... trouble.

jaredmoody:
A collection of rubocop autocorrected fixes for #945.


jasonrclark:
These all look fine to me. Will see how I feel about the `let do` in specs as we go along, but generally I think it looks better now than before. Thanks again @jaredmoody!

jaredmoody:
👍 👍

PR #1275, opened 2016-11-19
Fix overlapping fill on rounded rectangles

### LINKED ISSUE #1013: Rounded rect stroke doesn't cover fill at corners
image

(samples/simple-flashing.rb)

jasonrclark:
Fixes #1013

Integer math being slightly off between inner fill and outer draw on
borders sometimes let a single pixel show in the corners, which looked
ugly.

Want to give this some thorough sample testing and more thought before
calling it done.

### DIFF
--- a/shoes-swt/lib/shoes/swt/rect_painter.rb
+++ b/shoes-swt/lib/shoes/swt/rect_painter.rb
@@ -2,10 +2,6 @@ class Shoes
   module Swt
     class RectPainter < Common::Painter
       def fill(gc)
-        # If drawing a stroke around the shape, inset the fill so the draw
-        # isn't inside bounds of fill because of integer division remainders.
-        inset = inset_fill? ? 1 : 0
-
         gc.fill_round_rectangle(@obj.element_left + inset,
                                 @obj.element_top + inset,
                                 @obj.element_width - inset * 2,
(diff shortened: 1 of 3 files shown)


PragTob:
Can I haz before/after screenshots with the sample please? That way I can think better about why the changes were made and how we do it. :)

jasonrclark:
**Before**
image

**After**
image

Will look into those test failures.. looks like `Background` uses the `RectPainter`. I suspect it's just test setup problems, as I had to tweak the `RectPainter` tests to get them working.

Suspecting too in the light of morning that I don't need to offset vs the whole `strokewidth`. The fringe that shows is just a single pixel because of the rounding problems in our math. So if there's a stroke around the shape we can just offset the fill by 1 pixel, regardless the stroke width, and that'll be enough I think.

jasonrclark:
All right, took the WIP off as I think this is ready. Ran through samples and looking good. Also applied a restriction so we only do this insetting when we're rounded (i.e. have `corners` set in the painter) which should also limit the scope of where this might change the appearance of things.

PragTob:
Admittedly, I don't fully understand the change and effect but this is one of those changes where you need to sit and draw in your mind and on paper and I'm just trusting Jason and his testing here :)

couple of smaller questions/suggestions otherwise looks good :)

shoes-swt/lib/shoes/swt/rect_painter.rb
```diff
@@ -1,13 +1,17 @@
...
-        graphics_context.fill_round_rectangle(@obj.element_left,
-                                              @obj.element_top,
-                                              @obj.element_width,
-                                              @obj.element_height,
-                                              @obj.corners * 2,
-                                              @obj.corners * 2)
+      def fill(gc)
+        # If drawing a stroke around the shape, inset the fill so the draw
+        # isn't inside bounds of fill because of integer division remainders.
+        inset = inset_fill? ? 1 : 0
```
jasonrclark: I like it!

PR #1276, opened 2016-11-19
Reverse order of tests and Rubocop

jasonrclark:
For reasons I haven't quite sussed out yet, running Rubocop locally gets me tons of errors that aren't happening in CI.

I probably have something misconfigured, but in the meantime it'd be nice for Travis to fail faster when I do miss something Rubocop-related. This can be accomplished by just running Rubocop first (which is much faster than the specs.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: ruby
 script:
-  - bundle exec rake spec:all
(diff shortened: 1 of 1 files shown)


PragTob:
:+1: definitely linting before test execution

PR #1280, opened 2016-11-20
Rubocop: Fix most Style/MutableConstant violations

jaredmoody:
There are several areas in shoes where constants are defined that aren't _really_ constants. For example Shoes::LOG, Shoes::FONTS and Shoes::COLORS are all arrays that can be modified at runtime - this was the source of the CI failing when those constants were frozen.

Maybe each of these constants should become methods that expose the underlying data structure (e.g. Shoes.log) but that would be a breaking API change that would require more discussion so for now I've just excluded the files that contain those mutable constants.


jasonrclark:
Looks like there's some legit failures over on Travis with this. (Appveyor isn't failing because it doesn't run the SWT portion of the suite... hard enough getting the core tests passing on Windows at the time 😬 )

Totally agree with postponing those other cases. Care to document those in an issue so we don't forget to come back around to them?

jaredmoody:
My bad, I wasn't careful enough with the changes I made to `shoes-swt/lib/shoes/swt/key_listener.rb`, fixed up now.

Also opened #1282 to address the exceptions.

PragTob:
> Looks like there's some legit failures over on Travis with this. (Appveyor isn't failing because it doesn't run the SWT portion of the suite... hard enough getting the core tests passing on Windows at the time :grimacing: )

it's not? That seems wrong... they should definitely run over there as well... ok looked into the history and it seems like we were limited on handles...

```
Failure/Error: Unable to find matching line from backtrace
Java::OrgEclipseSwt::SWTError:
No more handles
```

:|

PragTob:
I'll merge this for now, but I'm gonna open a ticket to use the magic/autofreeze string comment in our files then we don't need to freeze the strings per hand.

Issue #1136, opened 2015-05-25
Link doesn't take styling properly


jasonrclark:
Noticed during #1134 but doesn't actually have to do with hovering.

This app, if you comment in/out various styling lines, will give you a variety of different crashes:

```
Shoes.app do
  pa = para("Text", l = link("yO"))

  l.size = 42
  l.stroke = blue
  l.fill = yellow
end
```

`size=` gives this:

```
AfterDo::CallbackError: A NoMethodError: undefined method `element_left' for (Shoes::Link:0x00003eb4 "yO"):Shoes::Link was raised during an after_do callback block for method 'update_style' on the instance Shoes::Common::Style with the following arguments: {:size=>42} defined in the file /Users/jclark/source/shoes4/shoes-swt/lib/shoes/swt/redrawing_aspect.rb in line 81. This is the backtrace of the NoMethodError:
 /Users/jclark/source/shoes4/shoes-swt/lib/shoes/swt/redrawing_aspect.rb:99:in `redraw_element'
/Users/jclark/source/shoes4/shoes-swt/lib/shoes/swt/redrawing_aspect.rb:83:in `add_redraws'
org/jruby/RubyProc.java:271:in `call'
/Users/jclark/.rbenv/versions/jruby-1.7.16.1/lib/ruby/gems/shared/gems/after_do-0.3.1/lib/after_do.rb:113:in `_after_do_execute_callback'
/Users/jclark/.rbenv/versions/jruby-1.7.16.1/lib/ruby/gems/shared/gems/after_do-0.3.1/lib/after_do.rb:107:in `_after_do_execute_callbacks'
org/jruby/RubyArray.java:1613:in `each'
/Users/jclark/.rbenv/versions/jruby-1.7.16.1/lib/ruby/gems/shared/gems/after_do-0.3.1/lib/after_do.rb:106:in `_after_do_execute_callbacks'
/Users/jclark/.rbenv/versions/jruby-1.7.16.1/lib/ruby/gems/shared/gems/after_do-0.3.1/lib/after_do.rb:100:in `update_style'
/Users/jclark/source/shoes4/shoes-core/lib/shoes/common/style.rb:30:in `style'
/Users/jclark/source/shoes4/shoes-core/lib/shoes/common/style.rb:103:in `size='
```

Where `stroke=` gives this:

```
NameError: undefined local variable or method `update_stroke' for (Shoes::Link:0x00003eb4 "yO"):Shoes::Link
[...]

PR #1281, opened 2016-11-20
Look to redraw_target in RedrawingAspect

jasonrclark:
Going with @PragTob suggestion of having the `RedrawingAspect` look for an explicit `redraw_target` object that `Link` can override. This nicely allows us to keep this all in SWT-land too, which is great!

~~Simplest (but possibly hacky?) proposal for fixing #1136.~~

~~As was noted in an earlier PR, links and spans have trouble if they do anything which triggers a redrawing against them, which expects them to have `element_left`, `element_top`, etc dimensions solely for the purpose of calculating a redrawing area.~~

~~This change gets the code from #1136 fully working (no idea what the `strokewidth` problem was at the time, but it doesn't manifest now) at the cost of `Link` and `Span` advertising `element_*` methods that aren't actually quite right.~~

~~Looking for feedback whether this is an acceptable compromise to get link styling working. I can see two alternatives:~~

~~1) Create a new `redraw_left`, `redraw_top`... set of methods which the `RedrawingAspect` could use instead of `element_*`. The majority of Shoes elements would just plumb `redraw_*` -> `element_*`, but it would give us the wiggle-room for `Link` and `Span` to point it to the broader `TextBlock`~~

~~2) Actually get the `Link` and `Span` set up with true, correctly calculated dimensions. This turns out to look modestly hard, because the lower-level text fitting is the only thing which knows about those dimensions, and it's forgotten about the specific segment objects long before it gets to really drawing things.~~

FWIW, I'm in favor of either just rolling with this (for simplicity and momentum), or doing option 1) which is relatively low impact. @PragTob what do you think?

### DIFF
--- a/shoes-swt/lib/shoes/swt/redrawing_aspect.rb
+++ b/shoes-swt/lib/shoes/swt/redrawing_aspect.rb
@@ -88,15 +88,23 @@ class Shoes
       end
 
       def redraw_element(element, include_children = true)
-        redraw_target = element.gui&.redraw_target || element
-        redraw_area redraw_target.element_left, redraw_target.element_top,
-                    redraw_target.element_width, redraw_target.element_height,
+        target = redraw_target(element)
+        redraw_area target.element_left, target.element_top,
+                    target.element_width, target.element_height,
                     include_children
       end
 
       def redraw_area(left, top, width, height, include_children = true)
         unless app.disposed?
           app.redraw left, top, width, height, include_children
+        end
+      end
+
+      def redraw_target(element)
+        if element.gui
+          element.gui.redraw_target
+        else
+          element
         end
       end
 
--- a/shoes-swt/lib/shoes/swt/redrawing_aspect.rb
+++ b/shoes-swt/lib/shoes/swt/redrawing_aspect.rb
@@ -88,8 +88,9 @@ class Shoes
       end
 
       def redraw_element(element, include_children = true)
-        redraw_area element.element_left, element.element_top,
-                    element.element_width, element.element_height,
+        redraw_target = element.gui&.redraw_target || element
+        redraw_area redraw_target.element_left, redraw_target.element_top,
+                    redraw_target.element_width, redraw_target.element_height,
                     include_children
       end
 
--- a/shoes-core/lib/shoes/dimensions.rb
+++ b/shoes-core/lib/shoes/dimensions.rb
@@ -219,5 +219,9 @@ class Shoes
     extend Forwardable
 
     def_delegators :dsl, *DimensionsDelegations::CANDIDATE_METHODS
+
+    def redraw_target
+      @dsl
+    end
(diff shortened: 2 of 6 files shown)


PragTob:
Delegating to TextBlock doesn't seem quite right. Trying to think of the ways it could go wrong I come up with:

* Positioning could be off - shouldn't happen as atm only text blocks are positioned afaik
* Redrawing - this is the intended side effect here
* users grabbing Links/Spans and relying on the element_* dimensions and then are in for a surprise (wouldn't think that people would do that though.. or well hope, I'm sure someone will).

It got me thinking though... instead of pretending we are as big as our parent element isn't what we really want to do tell the redrawing "Hey, you can't redraw me but instead please just redraw this other bigger thing that contains me"? That'd be similar to the idea of having a redraw_left/right but instead a `redraw_target` or something that for most would return `self` but here it'd return the `text_block`.

Not sure though. Best would be real dimensions but that seems like a real horror to do and can be put off imo. So I guess it's this here, 1.) or 1.b.) (naming the `text_block` idea that) - WDYT?

jasonrclark:
1b, the redraw_target is just what I was flailing around for. Thank you @PragTob! Might even be able to isolate things to the SWT side of the world which would be ideal since the redrawing is purely in shoes-swt.

jasonrclark:
So updated to use your approach @PragTob and I like it. Let me know what you think.

SWT elements don't have any common base class, so I ran through samples to attempt to ensure I've found all the separate, discrete types involved.

shoes-swt/lib/shoes/swt/redrawing_aspect.rb
```diff
@@ -99,6 +100,14 @@ def redraw_area(left, top, width, height, include_children = true)
         end
       end
 
+      def redraw_target(element)
+        if element.gui
+          element.gui.redraw_target
+        else
+          element
+        end
+      end
```
PragTob: I can only assume that this was done because sometimes we get DSL elements here, sometimes we get backend elements? Might be worth unifying if that's the case.. not here though probably in another ticket.

PragTob:
LGTM! I added one little question that I'd like to resolve, not so it can be fixed up here but so we can create a new ticket or decide it's not important enough to track :)

shoes-swt/lib/shoes/swt/redrawing_aspect.rb
jasonrclark: So the actual reason is that we occasionally trigger the redrawing during initialization of some elements (Image in particular), so we'll get here before the `gui` is actually set in a class. Haven't dug too deeply on why we get that redraw at this point, since this seemed relatively benign.

PragTob: Filed #1289 - thanks for the explanation!

PR #1283, opened 2016-11-20
Rubocop: Your move, Creep

jaredmoody:
Low hanging rubocop fixes for #945

I’m not 100% sure on the Lint/EmptyWhen fix if someone can double check that, I’d appreciate it.


shoes-core/spec/shoes/framework_learning_spec.rb
```diff
@@ -18,8 +18,10 @@ def gui_init
   end
 end
 
-class Learning::App
-  include TextPlugin::App
+class Learning
+  class App
+    include TextPlugin::App
+  end
```
PragTob: This seems to break the tests, cause Learning is a `module` not a class :)

shoes-core/lib/shoes/common/positioning.rb
```diff
@@ -17,7 +17,7 @@ def _position(left, top)
         gui.update_position if gui && gui.respond_to?(:update_position)
       end
 
-      # displace(left: a number, top: a number) » self
+      # displace(left: a number, top: a number) >> self
```
PragTob: :+1:

samples/expert-pong.rb
```diff
@@ -38,8 +38,10 @@
 
     # move the @comp paddle, speed based on `compuspeed` variable
     @comp.left +=
-      if nx + (ball_diameter / 2) > @comp.left + paddle_size; compuspeed
-      elsif nx < @comp.left; -compuspeed
```
PragTob: wow that was ugly :D

samples/expert-irb-adjusted.rb
```diff
@@ -44,10 +44,7 @@ def run(str)
     $stdout.rewind
     [output, obj]
   rescue Object => e
-    case e
-    when Empty, Continue
-    else @line = ""
-    end
+    @line = "" unless e == Empty || e == Continue
```
jaredmoody: Rubocop was complaining about the empty `when`, but I made the change here - we're through all the auto-correctable cops now. I wasn't sure about this change though, so I mentioned it in the PR comments.  Thanks for catching that I should have used `===`. How do you want this to look though, would this be better?

`@line = "" unless [Empty, Continue].include?(e.class)`

PragTob:
Rubocop bumms out... 

```
samples/expert-irb-adjusted.rb:47:25: C: Avoid the use of the case equality operator ===.

    @line = "" unless e === Empty || e === Continue

                        ^^^

samples/expert-irb-adjusted.rb:47:40: C: Avoid the use of the case equality operator ===.

    @line = "" unless e === Empty || e === Continue

                                       ^^^

samples/expert-irb.rb:47:25: C: Avoid the use of the case equality operator ===.

    @line = "" unless e === Empty || e === Continue

                        ^^^

samples/expert-irb.rb:47:40: C: Avoid the use of the case equality operator ===.

    @line = "" unless e === Empty || e === Continue

                                       ^^^
```

I guess we could substitute it through an `e.is_a?(clazz)`

jaredmoody:
OK, used `is_a?` to fix up that exception comparison and fixed that module definition.

PragTob:
Thank you!

PR #1285, opened 2016-11-20
See if we can get a custom appveyor build running again

PragTob:
* it used to just time out, see #1051 - will try to run it a
  couple of times and see if improvements they have made or
  JRuby has made might have gotten rid off it.

edit: Of course this also updates JRuby on the windows machine so that we can have a more up to date version there and get rid off some baggage.... hopefully
edit2: also so happy I managed to dig out that windows script to install it... wouldn't have started implementing it myself again


PragTob:
omgz... it's running rubocop on all of JRuby/stdlib :|

PragTob:
Both passed on the first run.. trying again.

PragTob:
I'll let that be enough, let's see if it fails out on us

PR #1290, opened 2016-11-22
Implement arrow

### LINKED ISSUE #798: Implement Arrow
I see arrow described in the [manual](http://shoesrb.com/manual/Art.html#arrow) but I don't see it in our dsl.

I'd also be interested to discuss changing the way arrow works. I know that in Shoes3 it's styled after the NodeBox arrow ([ref](http://nodebox.net/code/index.php/Reference_%7C_arrow%28%29)) but it seems like the more natural way would be to do it like a line with a special end. For example `arrow(left, top, x2, y2, end, size)` would draw an arrow where `end` specifies which ends of the line get an arrow head, and `size` specifies the relative size of the arrowhead compared to the stroke width (for example 1 could mean same size, 2 could mean 200% of the strokewidth etc.).

Unfortunately SWT doesn't support arrows natively, so we'll have to create our own. Here's some relevant [code](http://www.coderanch.com/t/339505/GUI/java/drawing-arrows).

jasonrclark:
Fixes #798 

Basic implementation of the arrow. Slight oddities you may notice:

* Height isn't obeyed, although by being a dimensioned object we allow setting/getting it. This matches Shoes 3's behavior FWIW.
* We allow `rotate` in the hash keys like other Shoes 4 elements, although Shoes 3 didn't support that
* Unlike shape where the SWT element holds onto the path, I chose to keep that in the `ArrowPainter` and just have the element notify the painter of changes at which point we regenerate it.
* Specs largely cribbed from other shapes. If we had lots more shapes to do, there's a lot of redundancy we could extract, but doesn't feel worth it at this stage and given how simple the existing code is generally.
* Path generation is a fair bit of kind of slightly ugly, but don't know if pulling variables would actually make it clearer or not. For now, tweaking Rubocop to let it slide.


shoes-swt/lib/shoes/swt/arrow_painter.rb
```diff
@@ -0,0 +1,36 @@
...
+          path.move_to(@obj.left - @obj.width / 2, @obj.top)
+          path.line_to(@obj.left - @obj.width / 2, @obj.top - @obj.width * 0.20)
+          path.line_to(@obj.left + @obj.width * 0.10, @obj.top - @obj.width * 0.20)
+          path.line_to(@obj.left + @obj.width * 0.10, @obj.top - @obj.width * 0.20 - @obj.width * 0.20)
+          path.line_to(@obj.left + @obj.width * 0.50, @obj.top)
+          path.line_to(@obj.left + @obj.width * 0.10, @obj.top + @obj.width * 0.20 + @obj.width * 0.20)
+          path.line_to(@obj.left + @obj.width * 0.10, @obj.top + @obj.width * 0.20)
+          path.line_to(@obj.left - @obj.width / 2, @obj.top + @obj.width * 0.20)
+          path.line_to(@obj.left - @obj.width / 2, @obj.top - @obj.width * 0.20)
+          path.line_to(@obj.left - @obj.width / 2, @obj.top)
```
PragTob: I think I'd prefer all the floats (0.10, 0.20, 0.5) to be constants.

Not totally sure how to make this smaller... it seems though that `width * 0.x` is rather repeated so putting that into local variables might help as it might also have a meaning like... dunnno `arrow_thickness` or something like that.

Looking at it from that angle one _maybe_ could also extract methods like draw_arrowhead/draw_arrowbody. Sadly, I don't really understand it so I'm more guessing :D

PragTob:
Looks good to me overall, left a couple of ideas at the method you mentioned :)

shoes-swt/lib/shoes/swt/arrow_painter.rb
jasonrclark: So was actually underway making a stab at giving the positions names at least rather than the constants. Ran with body/head naming and gave a diagram to help illustrate where the various points are located. What do you think does that help?

On plus side, passes Rubocop without modification now :+1:

jasonrclark:
Ha, spoke too soon on Rubocop there... still had issues with it over in `shoes-swt/lib/shoes/swt.rb`. Given that file has long sequences of requires, I'm just excluding the file entirely rather than bumping the counts for it.

PragTob:
I still feel a bit like some of those floats might better live in constants as they are repated and to my understanding a change in one would change the other... that said not important enough to stop this. Thanks for the work @jasonrclark ! :)

PR #1292, opened 2016-11-22
Rubocop: Fix all double negation

jaredmoody:
How’s this look?

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -104,11 +104,6 @@ Style/ClassVars:
 Style/Documentation:
   Enabled: false
 
-# Offense count: 1
-Style/DoubleNegation:
-  Exclude:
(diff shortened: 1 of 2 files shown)


PragTob:
Hmm, form an introspection point of view I'd like the variable to be a boolean.

How about...  `@fixed_height = !height.nil?` and then just remove the `!!` ? Does this work? Otherwise we could also do something like `@variable_height = height.nil?` and then adjust the methods

jasonrclark:
Does anything other than these predicate methods is @fixed_height? On phone so I can't search easily but I don't see use in the file, or att exposure.

If that's the case, how about just one nil? check in initialize and then the predicates are just pure returns?

PragTob:
@jasonrclark was the only usage I could find, I'm with you there one nil check in initialize :)

jasonrclark:
❤️

PR #1293, opened 2016-11-22
Rubocop: Fix all Style/FileName

jaredmoody:
This one seems like it might be questionable - rubocop wants all ruby file names snake cased.

I just did a quick script rename, so no worries if we want to disable this cop.


PragTob:
Hm yes we definitely wanted to rename those to fit ruby naming conventions, I thought there was a ticket for that couldn't find it but in #353 we talk quite a bit about samples naming.

I believe you also need to change them in `samples/README.md` :)

I think this is good, @jasonrclark what do you think?

Btw. :+1: on small PRs!

jasonrclark:
Hey @jaredmoody! Sorry I appear to have gotten things even further snarled by recent changes I've made in the `samples`. Did you want to carry forward with this or shall I right my wrongs? :)

jaredmoody:
Sorry I ate too much 🦃 and fell asleep for a few days. I can probably look at this this weekend but if you want to try and untangle before then, go for t :)

jasonrclark:
Picked these changes up and integrated with my sample work in #1314. Thanks @jaredmoody!

PR #1295, opened 2016-11-23
Update to latest SWT version 4.6.1

PragTob:
~~**DO NOT MERGE YET**~~

There was a small API breakage, which already surprised me. But running through the samples it seems like there is even more breakage and I aborted at some point. Breakage seems to be related to dimensions and seemingly mostly text :|

A week back or so I tried to find a good SWT changelog but came up short :|

First an image from master, then it's SWT 4.6.1 equivalent.

#### simple-loogink-cy.rb 

photo frame _005

photo frame _003

Text is cut off at the bottom.

#### simple-logo-display.rb 

shoes 4 logo icon _001

shoes 4 logo icon _002

text is cut off at the bottom again

#### simple-color-selector.rb 

shoes 4 _006

shoes 4 _004

Text starts wandering off into oblivion.

-------------------------------------------------------------------------------------

At best this is some off by x error where we manually add/remove something that adds up... worst case a very long debuggigng session. Sadly I need to work on some Elixir projects for a presentation in 1.5 weeks so don't have the time to dig deep :(

### DIFF
--- a/shoes-swt/lib/shoes/swt/text_block/text_segment.rb
+++ b/shoes-swt/lib/shoes/swt/text_block/text_segment.rb
@@ -84,8 +84,17 @@ class Shoes
           }
         end
 
+        # Why not just use layout.bounds?  Recent versions of SWT have shown
+        # platform dependent behavior about whether spacing is included in
+        # bounds, leading to unpredictable placement.  Doing it more directly
+        # per-line in this fashion insulates us from the regression. #1295
         def height
-          layout.bounds.height - layout.spacing
+          line_count = layout.line_count
+          h = (line_count - 1) * layout.spacing
+          line_count.times do |i|
+            h += layout.get_line_bounds(i).height
+          end
+          h
         end
 
         def last_line_height
--- a/shoes-swt/lib/shoes/swt/list_box.rb
+++ b/shoes-swt/lib/shoes/swt/list_box.rb
@@ -27,7 +27,7 @@ class Shoes
       end
 
       def update_items
-        @real.items = @dsl.items.to_a.map(&:to_s)
+        @real.set_items(*@dsl.items.to_a.map(&:to_s))
       end
 
       def text
--- a/shoes-swt/shoes-swt.gemspec
+++ b/shoes-swt/shoes-swt.gemspec
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
   s.test_files    = s.files.grep(%r{^(test|spec|features)/})
   s.require_paths = ["lib"]
 
-  s.add_dependency "swt", "~>4.4.0"
+  s.add_dependency "swt", "~>4.6.1"
   s.add_dependency "after_do", "~>0.4.0"
   s.add_dependency "shoes-core", Shoes::Swt::VERSION
   s.add_dependency "shoes-package", Shoes::Swt::VERSION
--- a/shoes-swt/spec/shoes/swt/list_box_spec.rb
+++ b/shoes-swt/spec/shoes/swt/list_box_spec.rb
@@ -13,7 +13,7 @@ describe Shoes::Swt::ListBox do
   let(:block) { ->() {} }
 
   let(:real) do
-    double('real', text: "", :items= => true, :text= => true, set_size: true,
+    double('real', text: "", :set_items => true, :text= => true, set_size: true,
                    add_selection_listener: true, disposed?: false)
   end
 
@@ -33,7 +33,7 @@ describe Shoes::Swt::ListBox do
     allow(dsl).to receive(:items).and_return ["hello"]
     subject.update_items
     # creation already calls update_items once
-    expect(real).to have_received(:items=).with(["hello"]).twice
+    expect(real).to have_received(:set_items).with("hello").twice
   end
 
   it "should set real text if initialized with choose option" do
@@ -59,14 +59,14 @@ describe Shoes::Swt::ListBox do
 
   it 'sets the items on real upon initialization' do
     subject
-    expect(real).to have_received(:items=).with(items)
+    expect(real).to have_received(:set_items).with(*items)
   end
 
   it 'converts array to string' do
     allow(dsl).to receive(:items).and_return [1, 2, 3]
     subject.update_items
     # creation already calls update_items once
-    expect(real).to have_received(:items=).with(%w(1 2 3)).twice
+    expect(real).to have_received(:set_items).with("1", "2", "3").twice
   end
 
   describe "when the backend notifies us that the selection has changed" do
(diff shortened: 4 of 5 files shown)


jasonrclark:
Thanks for looking into this @PragTob! Especially since it's showing up in text stuff, I might also take a look as I've got more 👟  time at the moment. Good luck on your Elixir presentation!!!

jasonrclark:
Hmm, so good news/bad news. Pulled down the branch, updated the SWT gem with bundler and... nothing? I don't see any of the problems on the Mac that you're seeing on Linux.

Don't have a virtual Linux around anymore that I was using, so that'll take a bit to set up, but this might be OS specific? Or maybe JVM/SWT version specific?

```
~/source/shoes4:java -version
java version "1.8.0_102"
Java(TM) SE Runtime Environment (build 1.8.0_102-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.102-b14, mixed mode)
```

PragTob:
I... grrm gnargh. Please not? :| that'd be the worst. I'm travelling atm but please ping me if I don't check within a week but I'm pretty sure it was broken for me... although it might have been my old desktop with java 7 so it might work on Java 8? Not sure. We'll see.

PragTob:
Still broken on 16.04 with its package base etc. on my desktop mine, for reference java version..


```
tobi@speedy ~/github/shoes4 $ java -version
openjdk version "1.8.0_111"
OpenJDK Runtime Environment (build 1.8.0_111-8u111-b14-2ubuntu0.16.04.2-b14)
OpenJDK 64-Bit Server VM (build 25.111-b14, mixed mode)
```

:|

Rebased either way so that that's at least done... might even be wort looking for SWT issues and maybe waiting a release because if it's Linux only chances are it's a bug they might fix. Should/could also try on Windows.

jasonrclark:
Thanks for checking @PragTob. Now that I've got working VMs for both Ubuntu and Windows, I can also do some digging once I've got `pre7` out the 🚪.

PragTob:
Rebased, made a simpler test script for at least one of the errors:

```ruby
Shoes.app do
  flow do
    para 'a'
    para 'b'
    para 'c'
    para 'd'
    para 'e'
    para 'f'
    para 'g'
  end
end
```

selection_030

PragTob:
Ok, lunch time soon, so I'll write down my findings.

The culprit seems to be `TextBlock#update_current_position` (with debug output by yours truly):\

```ruby
      def adjust_current_position(current_position)
        current_position.y = @dsl.absolute_bottom + NEXT_ELEMENT_OFFSET

        last_segment = segments.last
        if last_segment && !@bumped_to_next_line
          # Not quite sure why this is necessary. Could be a problem in some
          # other part of positioning, or something about how text layouts
          # actually draw themselves.
          current_position.x -= 1

          puts "reset current_position: #{current_position.y}"
          puts "last_segment.line_height: #{last_segment.last_line_height}"
          current_position.y -= last_segment.last_line_height
          puts "and we end up at: #{current_position.y}"
        end
      end
```

which ends up with:

```
Before 0
After: 0
current_position.y: 0
reset current_position: 13
last_segment.line_height: 17
and we end up at: -4
adjusted current position y: -4
#<struct Shoes::Slot::Position x=8, y=-4, next_line_start=13>
Relative y: -4
```

we seem to get the -4 because it's the differnce between oru dsl.absolute_bottom and the last_segment.last_line_height, which is defined as:
 
```ruby
        def last_line_height
          layout.get_line_bounds(layout.line_count - 1).height
        end
```

I'm guessing that something about the `get_line_bounds` changed in SWT and we might have to do something about that, try to get back old behaviour or maybe the dsl.absolute_bottom also ends up wrong for one reason or another.

jasonrclark:
Progress, although the news doesn't look good. Along the lines of what you're suspecting @PragTob, something has changed around the layout bounds.

More specifically, it appears that on the new SWT on Linux (GTK-backed version I'm guessing?), `layout#bounds` has already subtracted the `spacing` so when [we apply that same subtraction again](https://github.com/shoes/shoes4/blob/master/shoes-swt/lib/shoes/swt/text_block/text_segment.rb#L88) things get wacky. When I remove that subtraction on the offending branch + OS, everything looks right.

Buuuuuut, that same change looks wrong on my Mac (no idea yet what it would look like on Windows 😭)

So where does this leave us? I'm dropping off for a bit, but next thing I'll look at is to see if I can detect the situation just from the layout values in some fashion. Maybe there's something in there we could conditionalize this off of.

Worst case, if we can't spot something within the text layout values, we can fall back to an OS/platform-based condition on whether to subtract spacing or not.

jasonrclark:
All right, so I lied about stepping away apparently :)

It looks as though using the `layout.get_line_bounds` on each individual line **is** consistent although to return heights like we had before we have to factor in the spacing ourselves ala:

```
        def height
          line_count = layout.line_count
          h = line_count * layout.spacing
          line_count.times do |i|
            h += layout.get_line_bounds(i).height
          end
          h
        end
```

Ugly, because `layout.get_line_bounds` isn't neatly enumerable like we'd wish for, but this appears to get it done on both OSs in brief testing (@PragTob's sample app + the manual which is text-heavy).

I'll do some further testing on this as I have a chance, maybe ring in the New Year heads down coding on version compatibility bugs! 🎉

PragTob:
Thanks for taking the time to look into this! :) Sounds like a good solution imo. Feel free to push on this branch!

jasonrclark:
Thanks @PragTob! Just pushed up what I've been testing against... nearly through a sample run on all three platforms and looking 👌 

I'm going to look a bit closer at specing out the changes around `TextSegment`, as I noticed my update didn't break any specs, which strongly suggests that the actual height isn't being covered. While it might get a bit fiddly, given our recent experience I think some type of specs there would be worthwhile.

Once that's in, I think we'll be ready to merge this and get on 4.6.1 of `swt`, and look at rev'ing that without our desired fixes for packaging.

jasonrclark:
Oh, and will drop this here since it might be useful... no changelog, but I located the bugs marked as resolved in 4.6.1 [with this query](https://bugs.eclipse.org/bugs/buglist.cgi?component=SWT&product=Platform&query_format=advanced&target_milestone=4.6.1). Still nothing shedding much light on what's altered here (and probably other intervening releases between 4.4 and 4.6.1 that would have to be searched).

jasonrclark:
Removing the DNM tagging as I *think* this is good to go. Added the specs, testing's been fine... What do you think @PragTob?

shoes-swt/lib/shoes/swt/text_block/text_segment.rb
```diff
@@ -84,8 +84,17 @@ def font_styling
           }
         end
 
+        # Why not just use layout.bounds?  Recent versions of SWT have shown
+        # platform dependent behavior about whether spacing is included in
+        # bounds, leading to unpredictable placement.  Doing it more directly
```
PragTob: the double spaces seem slightly weird to me but not weird enough :)

PragTob: thanks for the comment though, much needed!

PragTob:
Seems great thanks for fiddling with this, I suspect I couldn't have gotten it working like you did :)

Yeah let's merge it, worst case we notice later that samples break that we didn't check and we just revert it.

PR #1297, opened 2016-11-23
Fix test coverage

### LINKED ISSUE #1287: Testcoverage reporting is broken
It seems like test coverage reporting is broken, maybe at the same time that I upgraded the code climate test coverage reporter...

selection_041

if you [look at the coverage](https://codeclimate.com/github/shoes/shoes4/coverage), you can see that it only counts **2** files at the moment.

So no idea what's wrong there atm... maybe some update is broken, maybe the wrong ruby version updates and doesn't send it all along.. needs further investigation.

PragTob:
* remember kids, always load simplecov before any code (something
  somewhere was flipped around so that swt was loaded before the
  DSL spec_helper so that it couldn't be covered)
* adds command_name so that results are correctly merged
* does not run it with package, as that has a sample app and seems
  to mess with test coverage somehow (we end up with 99.xx%
  coverage then covering things that are definitely not covered
  like the logger app)
* fixes #1287

### DIFF
--- a/shoes-core/spec/spec_helper.rb
+++ b/shoes-core/spec/spec_helper.rb
@@ -2,7 +2,14 @@ SHOESSPEC_ROOT = File.expand_path('..', __FILE__)
 $LOAD_PATH << File.join(SHOESSPEC_ROOT)
 $LOAD_PATH << File.join(SHOESSPEC_ROOT, "../lib")
 
-require_relative '../../spec/code_coverage'
+# loaded again when executing DSL specs with the SWT backend, therefore
+# we souldn't start the test coverage again or override the command as it'd
+# mess up merging
+unless defined?(SimpleCov)
+  require_relative '../../spec/code_coverage'
+  SimpleCov.command_name 'spec:shoes'
+end
+
 require 'rspec'
 require 'rspec/its'
 require 'pry'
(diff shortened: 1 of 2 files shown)


jasonrclark:
Looks good here. Thanks @PragTob!

jasonrclark:
Hmm, so the coverage has updated (94%) but appears to only include `shoes-core` rather than `shoes-swt` :/ Will poke at it more later, as I've got to get down to some chores now.

PragTob:
@jasonrclark strange... it definitely ran SWT as well:

core: Coverage report generated for spec:shoes to /home/travis/build/shoes/shoes4/coverage. 2619 / 2791 LOC (93.84%) covered.

swt: Coverage report generated for spec:shoes, spec:swt to /home/travis/build/shoes/shoes4/coverage. 4663 / 4991 LOC (93.43%) covered.

Seems like the merging it together somehow failed or the codeclimate test reporter doesn't really support the merging together of results?

Also, we effectively have a race condition between the builds, but probably last build wins for reporting coverage results... which should be fine unless jruby-head breaks something

jasonrclark:
Yeah, stranger and stranger. Definitely looks like the specs generated coverage. Popped up a little debug branch (chores can wait!) and it shows that the results we created does include `shoes-swt` files... https://travis-ci.org/shoes/shoes4/jobs/178342068

FWIW, local coverage run for me is almost the same overall %... so at least it's coincidentally correct 😬

PR #1299, opened 2016-11-25
Attaching elements

### LINKED ISSUE #807: Implement Attach for flows and stacks
Flows and Stacks should be able to attach to things: http://shoesrb.com/manual/Styles.html#:attach

I played around with this in Shoes3 and it seems like the primary purpose of this style is to fix content relative to the window (I'm not sure what good attaching a slot to another slot/element is, but that's not for me to judge! :)

For example, a fixed bottom bar (in Shoes 3 only for now):

``` ruby
Shoes.app do   
  50.times do
    stack {para "blah"}
  end

  f = flow height: 30, attach: Shoes::Window, top: app.height - 30 do
    background black
    para "Footnotes!", stroke: white
  end
end
```

jasonrclark:
Revises slot positioning to take advantage of attaching an element to another slot. Fixes #807.

## Sample
Sample program exercising `attach`:

```
Shoes.app do
  stack do
    para "blah"
    para "blah"
    para "blah"
    para "blah"
  end

  @s = stack do
    background blue
    para "The Body!!", stroke: white
  end

  flow height: 60, width: 50, attach: @s, left: 20, top: 30 do
    background black
    para "Footnotes!", stroke: white
  end
end
```

**Before**
image

**After**
image

## TODO

* [x] Look into `Shoes::Window` as attachment target, although that class doesn't exist at all (yet?) in Shoes 4 (http://shoesrb.com/manual/Styles.html#:attach)
* [x] Add a sample using attach... something with real purpose behind fiddling with positioning
* [x] More testing, particularly see how things work with margins
* [x] Rename `CurrentPosition` to `Position`

### DIFF
--- a/shoes-core/lib/shoes/slot.rb
+++ b/shoes-core/lib/shoes/slot.rb
@@ -29,7 +29,7 @@ class Shoes
     end
 
     def handle_block(blk)
-      @current_position = CurrentPosition.new element_left,
+      @current_position = Position.new element_left,
                                               element_top,
                                               element_top
       @blk = blk
@@ -167,10 +167,10 @@ class Shoes
 
     protected
 
-    CurrentPosition = Struct.new(:x, :y, :next_line_start)
+    Position = Struct.new(:x, :y, :next_line_start)
 
     def position_contents
-      @current_position = CurrentPosition.new element_left,
+      @current_position = Position.new element_left,
                                               element_top,
                                               element_top
 
@@ -212,7 +212,7 @@ class Shoes
 
       position_element_at element, attached_left, attached_top
 
-      CurrentPosition.new attached_left, attached_top, attached_top
+      Position.new attached_left, attached_top, attached_top
     end
 
     def position_slotted_element(element, current_position)
--- a/shoes-core/lib/shoes/slot.rb
+++ b/shoes-core/lib/shoes/slot.rb
@@ -183,16 +183,13 @@ class Shoes
 
     def positioning(element, current_position)
       if element.needs_positioning?
-        if element.attached_to
-          align_position = position_attached_element element
-        else
-          position_element element, current_position
-          align_position = current_position
-        end
+        align_position = if element.attached_to
+                           position_attached_element(element)
+                         else
+                           position_slotted_element(element, current_position)
+                         end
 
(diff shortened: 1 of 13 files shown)


shoes-core/lib/shoes/slot.rb
```diff
@@ -205,6 +209,15 @@ def position_element(_element, _current_position)
       raise 'position_element is a subclass responsibility'
     end
 
+    def position_attached_element(element)
+      attached_left = element.attached_to.element_left + element.style[:left].to_i
+      attached_top  = element.attached_to.element_top + element.style[:top].to_i
+
+      position_element_at element, attached_left, attached_top
+
+      CurrentPosition.new attached_left, attached_top, attached_top
```
PragTob: So after we position the attached element we continue after the attached element? That is... interesting. I sort of expected it to be free floating and not influence current_position at all (like normal elements with their left/top set). Is that the behavior in Shoes 3?

PragTob:
LGTM save for the points you lined out that still need to get done :)

shoes-core/lib/shoes/slot.rb
```diff
@@ -183,15 +183,19 @@ def position_contents
...
       if element.needs_positioning?
-        position_element element, current_position
-        element.contents_alignment(current_position) if element.respond_to? :contents_alignment
-      end
+        if element.attached_to
+          align_position = position_attached_element element
+        else
+          position_element element, current_position
+          align_position = current_position
+        end
```
jasonrclark: Yeah, the lack of symmetry between attached/not there was bothering me. Will see what I can work up.

jasonrclark:
So I think this should be ready to go (getting CI running while I do a quick scan through the full samples before bed here). Things I learned:

* `Shoes::Window` exists in Shoes 3 but doesn't appear to do anything except be a marker for saying things like `attach: Shoes::Window`. The class has no instance methods of its own, and greping the old code turned up minimal usage of it.
* Margins aren't respected in positioning of attached elements. This meant a change from `element_left` to `absolute_left` in my slot code. Fair enough.

PragTob:
Ah you already fixed the one indentation thing.. all great then! :)

PR #1302, opened 2016-11-26
Remove unnecessary targetting logic in hover

jasonrclark:
PR #1281 dealt with the dependency on actual coordinates in the SWT
`RedrawingAspect`. We don't need to jig the target around in the DSL layer
anymore.

Also restore the `simple-menu` sample to good standing since it works as
well as it did on Shoes 3 (and demonstrates `hover`)!


PragTob:
:star: :heart: :star:

PR #1303, opened 2016-11-26
Fix expert-othello

### LINKED ISSUE #534: samples/expert-othello.rb text/button layout is funky
Except for the text display and the undo feature, the game plays perfectly. Issues:
- Player text isn't positioned properly, and over time, the player labels turn black
- There is no undo button

`samples/expert-othello.rb`:

image

The text layout of `expert-othello-adjusted.rb` looks better, but the play doesn't work. Note that the adjusted version was edited by @ashbb to run on green shoes...we may want to focus on the original, since it's close to working.

`samples/expert-othello-adjusted.rb`:

image

jasonrclark:
Fixes #534

* Fixes crash on startup from bad Rubocop auto-fix on `zero?`
* Addresses odd placement by setting size on player banner slots
* Fixes undo placement so it actually shows up
* Fixes undo so it doesn't let you revert initial positioning
* Fixes drawing glitch on alerting for errors


PragTob:
_the trust is strong_ :)

Issue #519, opened 2013-12-29
samples/expert_irb.rb not scrolling properly


jasonrclark:
Running this sample crashes about not having `scroll_top=` on the `Shoes::Stack`

```
✘ [jruby] (master)*
♥♥♥♥~/source/shoes4:bin/shoes samples/expert-irb-adjusted.rb

NoMethodError: undefined method `scroll_top=' for #<Shoes::Stack:0x6a7ffd16>
            (root) at /Users/jclark/source/shoes4/samples/expert-irb-adjusted.rb:108
              call at org/jruby/RubyProc.java:271
        eval_block at /Users/jclark/source/shoes4/lib/shoes/swt/key_listener.rb:51
        eval_block at /Users/jclark/.rbenv/versions/jruby-1.7.8/lib/ruby/gems/shared/gems/after_do-0.2.2/lib/after_do.rb:89
  handle_key_event at /Users/jclark/source/shoes4/lib/shoes/swt/key_listener.rb:47
        keyPressed at /Users/jclark/source/shoes4/lib/shoes/swt/key_listener.rb:118
        keyPressed at Shoes$$Swt$$KeypressListener_1241086277.gen:13
    run_event_loop at /Users/jclark/.rbenv/versions/jruby-1.7.8/lib/ruby/gems/shared/gems/swt-0.18/lib/swt/minimal.rb:109
        event_loop at /Users/jclark/.rbenv/versions/jruby-1.7.8/lib/ruby/gems/shared/gems/swt-0.18/lib/swt/minimal.rb:92
              open at /Users/jclark/source/shoes4/lib/shoes/swt/app.rb:34
        initialize at /Users/jclark/source/shoes4/lib/shoes/app.rb:47
               app at /Users/jclark/source/shoes4/lib/shoes/app.rb:15
            (root) at /Users/jclark/source/shoes4/samples/expert-irb-adjusted.rb:64
              load at org/jruby/RubyKernel.java:1103
            (root) at /Users/jclark/source/shoes4/lib/shoes/cli.rb:1
       execute_app at /Users/jclark/source/shoes4/lib/shoes/cli.rb:73
            (root) at /Users/jclark/source/shoes4/bin/ruby-shoes:6
```

jasonrclark:
While this no longer crashes, the scrolling still isn't really working properly. The console doesn't follow when we make new lines past the end of things. Oddly, once I resize the window then I get scrolling enabled manually, but it doesn't work from where we start out.

PR #1304, opened 2016-11-26
Fix up expert-irb

jasonrclark:
Sample as existed works just about as well as the adjusted, so falling back on the original.

* Fixed problem with initial console not getting splatted
* Still has scrolling issues, further notes in #519, but good enough to run it in samples


PragTob:
Fixed up merge conflicts, should be good to go now

PR #1305, opened 2016-11-26
Remove game-of-life-adjusted

jasonrclark:
Performance between old and new versions looks almost same, visuals and
placement all line up right too.


PragTob:
:+1:

PR #1310, opened 2016-11-27
Implement translate and move expert-minesweeper to unadjusted version

### LINKED ISSUE #1306: Missing proper `translate`
[Per the manual](http://shoesrb.com/manual/Art.html#translate) `translate` is a method that takes `left` and `top` as parameter to move the base where drawing starts from. We've got it defined instead as a style, which isn't right.

This is a main blocker to getting the older version of `samples/expert-minesweeper.rb` back to running, and a clear compatibility miss.

Sample program:

```
Shoes.app do
  translate 100, 150
  rect 0, 0, 100, 100
end
```

Expected display:
image

Currently just get a crash on wrong number of arguments when we run on Shoes 4.

jasonrclark:
Fixes #1306 

Lot of diffs here because this applied specially to every individual art element in the library. Whew.

* Added correct `Shoes::App#translate` methods to take expected parameters
* Added `Shoes::Common::Translate` module for more easily extracting the left/top from any translation we've applied. (Also alphabetized some things along they way when including modules for these elements)
* Spiked through on each art element type to use the translation when deciding where to draw things.
* Our classes were split between a few with named `Painter` classes (like `RectPainter`) and those with an embedded `Painter` in the SWT class. I like the separate classes better, so I split them. Bummer that that's where a bunch of noise came from... in hindsight, maybe shoulda refactored that in a separate step before/after. ~~Ah well.~~
* Flipped `samples/expert-minesweeper.rb` to being the correct version and removed the adjusted version. 🎉 
* Disabled Rubocop `Style/TrailingUnderscoreVariable` and made a change so it's easier to find failed Cops in CI output.

**TODO**
* [x] Confirm about element level `translate` behavior back in Shoes 3. Suspect we're not doing the right thing quite yet. _So I was totally wrong, element level `translate` doesn't exist. I think that's handled by just setting dimensions or using `displace`, where `translate` is intended as a higher level "move all the things together" operation._
* [x] What happens with multiple `translate` calls at the app level? _Turns out we're fine. Elements capture the app styles at their point of creation like we desire, so it will get the `translate` that's in sequence of the app evaluation. Hurray!_
* [x] Tidy up on the SWT side so we don't have to put the intervening `dsl` in `obj.dsl.translate_left`. `Shoes::Swt::Common::Translate` module probably? _Yup, much cleaner_
* [x] Should we cache the translation values for faster lookup in elements? How to invalidate if so? _Did so, and handled it via standard invalidation for styles that need a little extra love and care like dimensions and visibility_
* [x] Add a purposeful sample for translation
* [x] Rebase to fix samples conflict


shoes-core/lib/shoes/common/translate.rb
```diff
@@ -0,0 +1,19 @@
+class Shoes
+  module Common
+    module Translate
+      def translate_left
+        left, = translate
```
PragTob: `left, =` is a thing? I'd appreciate it to be `left, _ =` :)

jasonrclark:
Had a quick brainwave on getting a nicer diff... will get something pushed shortly so don't despair of catching the changes.

jasonrclark:
All right, pushed up a commit that moves the extracted painters back to their original locations, then restores the split. With that, you can view https://github.com/shoes/shoes4/compare/master...c84e3d710c72761c645633d3162b87f139aca845 and see what the actual changes looked like before the move.

shoes-core/lib/shoes/common/translate.rb
jasonrclark: All right, found the cop: http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/TrailingUnderscoreVariable and disabled that.

shoes-core/lib/shoes/common/translate.rb
```diff
@@ -0,0 +1,19 @@
...
+        left || 0
+      end
+
+      def translate_top
+        _, top = translate
+        top || 0
+      end
```
PragTob: Seeing that translate it something that we need to use on pretty much every draw, how about setting it as some instance variables? I might pre prematurely optimizing things here but the access seems a bit weird this way and it's just set one time isn't it?

PragTob:
@jasonrclark that wouldn't have been necessary, I liked the extraction - it was maybe a bit much all in here but not that bad. Plus if we don't do opportunistic refactoring, when do we refactor at all?

I like the changes, just thing that probably the translate info should end up in some instance variable of some object, e.g. sort of looking it up/destructuring it every time for something as frequent as drawing doesn't feel quite right

PragTob:
After reading the isssue I have another question... so what does translate do? Can it be called once for the entire app and that's it? It seems it could be called multiple times resetting those translate values... but it appears only on the app level - is that right?

If so, couldn't the painters etc. just look these values up in the app and don't bother the individual elements with them? Just a thought.

jasonrclark:
You raise some really good questions. I suspect (but haven't confirmed, and won't be able to until later today) that art elements themselves probably allow you to call `translate`. If you call it at that level, it'd probably just influence that element, called at the app it sets the "default" translation. If that's the case, then the individual elements need to be involved something like they are today.

I'm very curious too, though, what happens if you call it multiple times. I'll test it out, but I suspect that elements between the translate calls should get the different values. This would be a :+1: toward your idea of caching the values locally to elements not just for performance, but for actual correct behavior!

Making further TODO notes and will carry on. As always, so great to have your help @PragTob! 🍬

PragTob:
Thanks :slightly_smiling_face: 

I'm just happy to take the time and pitch in through some reviews, least I can do. You're doing all the work really so thanks so much for that :two_hearts: 

I'm hoping I can pitch in a bit more despite PR reviews/janitor work when benchee 0.6.0 & benchee_html are cut for release and the accompanying benchmarking presentation is ready :D Will probably only be after that is over which is next weekend :) We'll see.

Anyhow, thanks a lot as always :pray:

jasonrclark:
Done with TODOs, added notes to the items with how they were addressed. Samples all look good (after a brief moment where my OS was looking funny, but wasn't Shoes' fault... 😌 )

jasonrclark:
Re-reviewed this since it had been a bit. Might give it another quick test, but if I get down to doing any Shoes this weekend, I'll probably merge then if that's 🆒 .

jasonrclark:
Rebased after other merges that changed things up a bit. Will test samples and merge on 🍏

Issue #1239, opened 2016-06-04
Mac, ignoring shoes-core-4.0.0.pre5 because its extensions are not built...


maxiedaniels:
I'm trying to get shoes4 working on my Mac, system is El Capitan. First I've tried just 'gem install shoes --pre', which is the least broken for me. When I run shoes <myfile>.rb, I get "Ignoring shoes-core-4.0.0.pre5 because its extensions are not built.  Try: gem pristine shoes-core --version 4.0.0.pre5", and then a whole lot of errors about undefined methods.. ex. 'last_line_width', "contents_alignment'... etc. The app then displays, but the alignments are completely wrong.

I've tried the gem pristine command.. no dice. Nothing changes.

I've also tried the git clone + bundler method, but that seems to be completely broken for me. I'll start with a clean setup and do that method, but when I go to run shoes, it says a bunch of gems have extensions not built (ffi, gem-wrappers, nokogiri, unf_ext) and then it says /usr/local/bin/shoes: shoes4: command not found.

Any ideas? I have rvm, with ruby-head and jruby installed and ruby-head selected. I actually had to install rvm because El Capitan has the SIP feature which screws up gems in the system folder.

PragTob:
Hi there,

shoes does not work with ruby-head. Please try Jruby, preferable JRuby 9.1+ :)

maxiedaniels:
Ah - that would explain a lot. Thank you! I assumed it was running jRuby from within Shoes and that I didn't need to select jRuby as my current Ruby.

PragTob:
is this your error: #1217 ?

If so, then installing from master or a JRuby prior to 9.0.5.0 should help (they introduced a breaking change). And I should make a new gem release :)

Cheers,
Tobi

maxiedaniels:
Yup that's the error. But I'm trying to install from the master... when I run shoes now, I get:

`Ignoring executable-hooks-1.3.2 because its extensions are not built.  Try: gem pristine executable-hooks --version 1.3.2
Ignoring gem-wrappers-1.2.7 because its extensions are not built.  Try: gem pristine gem-wrappers --version 1.2.7
Ignoring jruby-launcher-1.1.1-java because its extensions are not built.  Try: gem pristine jruby-launcher --version 1.1.1
/usr/local/lib/ruby/site_ruby/2.3.0/rubygems/dependency.rb:308:in to_specs: Could not find 'shoes-core' (>= 0) among 72 total gem(s) (Gem::MissingSpecError)
[...]

PR #1314, opened 2016-12-04
Rebase and fix from "Rubocop: Fix all Style/FileName #1239"

jasonrclark:
This supercedes #1239, which needed some rebases after my fiddling in samples.

Fixes the code that was looking for `generate-backend.rb` in gems so it looks for `generate_backend.rb` which was the better name anyway.

I got most of the way through fixing up the broken cops from the renames when I remembered `.rubocop_todo.yml` had filenames in it that just didn't line up anymore. D'oh. Since I was almost done, I've just kept the fixes.

This also patches up the `samples/README` and a few other locations referencing sample files so everything keeps happy.


jasonrclark:
Failures are from a new version of Rubocop released last week. Will re-generate todo on this branch as well then get it merged today after a little more sample testing so things can get to 💚 on master.

jasonrclark:
Found one more edit in `samples/README` to make, but otherwise the samples all look 👌  Will be merging before EOD.

PragTob:
:tada:

PR #1315, opened 2016-12-05
Redraw after displace

### LINKED ISSUE #1308: Fix displace to refresh appropriately
`samples/good-displace.rb` shows initially the same between Shoes 3 and Shoes 4, but clicking the button to displace the image on Shoes 4 doesn't redraw. If you click it then resize the window, you'll see it get moved.

Neither of the versions appear to actually displace the oval or rect. Can't tell if that's a problem with the sample or a bug in both versions at this point.

jasonrclark:
Fixes #1308 and includes the sample now that it's 👌

### DIFF
--- a/samples/README
+++ b/samples/README
@@ -77,6 +77,7 @@ shape_arc_to.rb
 good_arc.rb
 good_bounce.rb
 good_clock.rb
+good_displace.rb
 good_follow.rb
 good_image_dl.rb
 good_potato_chopping.rb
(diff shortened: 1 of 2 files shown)


PragTob:
:heart: :heart: :heart:

PragTob:
Also I LOVE the small diff and the redrawing aspect. Dunno how everyone else feels about it these days but it's so great to have that all in one place

jasonrclark:
Yep. The pattern continues to hold up. ✨

PR #1316, opened 2016-12-06
Use magic comments for freezing instead of `.freeze`

jasonrclark:
~~Noticed when running `rake update_versions` to test some stuff that we were dropping it which makes Rubocop ☹️ ~~

Updates version files to include the magic comment, stops with the individual freezing there.

Needed update to tell Rubocop to target Ruby 2.3 to get it happy, but totally cool as we're targeting JRuby 9.1+


PragTob:
Couldn't we instead try to have the file where this is used be one of our first with `# frozen-string-iteral: true` or whatever the magic is?

jasonrclark:
Updated description. Since moving to Ruby 2.3 also needed to regen some Rubocop TODO's as the updated version introduces more things we're not doing yet. 👮

PragTob:
:heart:

PR #1322, opened 2016-12-09
Simple console example, right now does not work/log anything

PragTob:
Maybe the recent rubocop fixes broke it (as it's a whole it is
dangerously untested) and in general it needs some love as it uses
a constant to share data and also the updating isn't the nicest.

What doesn't work yet: the messages don't seem to show up in the log anymore :(

### DIFF
--- a/samples/simple_console.rb
+++ b/samples/simple_console.rb
@@ -3,7 +3,7 @@ Shoes.app do
 
   flow do
     %w(info debug warn error).each do |type|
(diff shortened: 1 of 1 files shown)


jasonrclark:
RUBOCOP!!!! 😏

PragTob:
Arghs grrr.. grml. 

> Prefer to_s over string interpolation.

Issue #1209, opened 2015-12-25
Packages ship with ruby 1.9


lmrodriguezr:
**Disclaimer**: This is probably not a :bug:, not sure.

I need to use `Process.spawn` options in my app, which was not supported by JRuby 1.7<sup>1</sup>. This works fine with `bin/shoes` or with `require 'shoes'`, so long as I have 9K in my system. The problem comes when I try to package my app, because the jar ships with ruby 1.9.3 instead of 2.x (I believe it's jruby 1.7.23), so the `Process.spawn` options are not understood when running from the jar. Here's a tiny session:

``` bash
$> cat samples/t.rb
# Just show the ruby version
$stderr.puts RUBY_VERSION
$> # With bin/shoes it gets my system's jruby
$> bin/shoes-swt t.rb
2.2.2
$> # With a jar it runs with 1.7
$> bin/shoes-swt -p swt:jar samples/t.rb
$> java -XstartOnFirstThread -jar samples/pkg/t.jar
1.9.3
```

Is there any way to tell shoes (or furoshiki) that I want the package to ship with the jruby-jars of 9K? I saw discussions on restricting support to 9K (with 2.x) over at #1179, so I'm assuming I'm not the only one that wants to play with 9K features :)

Happy holidays `shoes-people`!

---

**<sup>1</sup>** `Process.spawn` supported options in ruby 1.9, it was just not implemented in JRuby 1.7. Some discussion on this at jruby/jruby#985.

jasonrclark:
Thanks for the report! As you mention, I think there's something weird going on between the installed version of JRuby vs what gets into the package.

jasonrclark:
Fixed by https://github.com/shoes/furoshiki/pull/21

PR #1324, opened 2016-12-11
DNM: Packaging love :heart:

jasonrclark:
This is a WIP PR to start thrashing out packaging. Fair warning, this PR is probably gonna get long and involved, so buckle your seatbelts! 🚗 💥 

First things first, I'm aiming to get packaging from `bin/shoes` to generate a run-able jar.

## furoshiki update
This relies on https://github.com/shoes/furoshiki/pull/21 to pull in newest `warbler`. This gets the generated package up onto JRuby 9.1, which I believe will address #1209.

(Side note for those following along at home, testing this packaging is a pain in the 🍑! One of the biggest causes is that warbler picks up installed gems (which makes sense). But that means if we make a change to a Shoes gem to get things working, gotta re-build and install to test it. Aw, yeah.)

## Problem with `shoes-core` extensions
Next stall that I hit was that the generated jar refused to load `shoes-core` with this error:

```
✔ ♥♥♥♥ [jruby-9.1.6.0] (master)*
~/source/shoes4:java -XstartOnFirstThread -jar samples/packaging/pkg/package_me.jar
Ignoring shoes-core-4.0.0.pre7 because its extensions are not built.  Try: gem pristine shoes-core --version 4.0.0.pre7
LoadError: no such file to load -- shoes
  require at org/jruby/RubyKernel.java:959
  require at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:55
   <main> at uri:classloader:/META-INF/init.rb:5
  require at org/jruby/RubyKernel.java:959
   (root) at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:1
   <main> at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:55
ERROR: org.jruby.embed.EvalFailedException: (LoadError) no such file to load -- shoes
```

Done a little digging but haven't gotten a good solution to this quite yet in configuring warbler. Will see if I can get a minimal test-case together and open an issue over there to see if it's something known and/or a bug.

That said, we do have a workaround in the form of 1ec80f9b9e06c6d93666d577ff7f25df3e69c637. We only have "native extensions" to do our funky shell-script installation and avoid double-loading JRuby ([sigh](https://github.com/shoes/shoes4/blob/master/shoes-core/ext/install/Rakefile#L4)). If I can't get traction on the warbler/rubygems end of this, we could instead move the executable generation over into the `shoes` metagem, which we don't even bother pulling into our packaged apps. Then `shoes-core` could be extension-free, install happily, and no one would be the wiser. It actually makes a little sense to me that `shoes` would be the home for executable shenanigans vs `shoes-core` which is intended as the squeaky clean home for our DSL.

## Packaging detection issue
So, once dropping the extension building temporarily to get things chugging along, ran into a second snag with detecting when we're in a packaged state from the SWT app code for writing out the icon image. This changed from JRuy 1.7 to 9.x. We already properly handle it the new way elsewhere, so this will update it to be happy.

## Success?
I did get a packaged jar to actually run at the end of this. Wahoo! 🎉 🌮 💃 

## Next steps
Done for tonight, but will work next on getting a minimal duplication of our pain with `warbler` + `shoes-core`. Will probably do the move of executable fun from `shoes-core` to `shoes` in parallel on speculation since it's just mechanics.

After that, may play with testing out from a full `shoes` gem installation since there seemed to be other issues at play there related to Bundler.

### DIFF
--- a/shoes-core/shoes-core.gemspec
+++ b/shoes-core/shoes-core.gemspec
@@ -18,5 +18,5 @@ Gem::Specification.new do |s|
 
   # Curious why we don't install shoes? See ext/Rakefile for the nitty-gritty.
   s.executables   = ['shoes-picker', 'shoes-stub']
-  s.extensions    = ['ext/install/Rakefile']
+  # s.extensions    = ['ext/install/Rakefile']
 end
--- a/shoes-swt/lib/shoes/swt/app.rb
+++ b/shoes-swt/lib/shoes/swt/app.rb
@@ -1,7 +1,7 @@
 class Shoes
   module Swt
     shoes_icon = ::Shoes::ICON
-    if shoes_icon.include? '.jar!'
+    if shoes_icon.include? 'uri:classloader'
       ICON = File.join(Dir.tmpdir, 'shoes-icon.png').freeze
       open ICON, 'wb' do |fw|
         open shoes_icon, 'rb' do |fr|


PragTob:
Thanks! :heart: 

FWIW I'd highly prefer it if we kept it somewhat small and amybe even merged in small incomplete fixes otherwise reviews will get really hard and hairy. I'd be inclined to try to use the mikado technique here but might be overkill of rework :)

jasonrclark:
Yeah you're probably right about smaller chunks. The hard bit is that things like the SWT fix can't be tested without the other half baked stuff. Will ponder, revise and read up on mikado!

jasonrclark:
Logged https://github.com/jruby/warbler/issues/404 with a minimal repro case https://github.com/jasonrclark/warbling-to-the-core. Will see what feedback we get there before I plow much more effort into it. Depending on the direction for that, will probably close this out and slice it up into separate smaller PRs from that point.

jasonrclark:
Ok, so have been thinking about it and I've got an approach that'll make things relatively simple to approach while we're getting the native extensions resolved.

Plan is to PR individual changes necessary to get things working (ala #1325 and #1326) and generally get those small changes to master quickly.

I can then use whatever hacked together current solution I have (like on this PR), cherry-pick/edit it into place for testing. In light of that, will close this out in favor of another PR to come eventually.

PR #1327, opened 2016-12-13
Re-raise if in SHOES_ENV test

### LINKED ISSUE #1222: Don't swallow painter exceptions while running tests
Talking about [this code here](https://github.com/shoes/shoes4/blob/e22db7b713b73136294721bbc4f8f67c026c8abe/shoes-swt/lib/shoes/swt/common/painter.rb#L24-L29)

```
     rescue => e
      # Really important to rescue here. Failures that escape this method
      # cause odd-ball hangs with no backtraces. See #559 for an example.
      #
      puts "SWALLOWED PAINT EXCEPTION ON #{@obj} - go take care of it: " + e.to_s
      puts 'Unfortunately we have to swallow it because it causes odd failures :('
```

Introducing it was fairly critical, but sadly it can (and has) masked failures when running our specs. So we shouldn't swallow it and raise instead when running the tests.

I'm thinking setting some environment variable for which it then raises (might also be nice for debugging) and then automatically setting it in `spec_helper.rb` - something like that.

Any ideas @jasonrclark ?

jasonrclark:
Fixes #1222

Rather than make a specific value for just the exception swallowing, I went with an "env" flag to say we're in testing. Might come in handy elsewhere.

What do you think? All the other ideas I had for flag names just got unwieldy.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -41,7 +41,7 @@ Metrics/BlockLength:
 # Offense count: 15
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 251
+  Max: 252
 
 # Offense count: 8
 Metrics/CyclomaticComplexity:
--- a/shoes-core/lib/shoes/slot.rb
+++ b/shoes-core/lib/shoes/slot.rb
@@ -194,6 +194,8 @@ class Shoes
 
       update_current_position(current_position, element)
     rescue => e
+      raise e if ENV['SHOES_ENV'] == 'test'
+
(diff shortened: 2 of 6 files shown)


jasonrclark:
Ugh, maybe should step away from the computer for the night. There were tests that I apparently missed running which checked explicitly that we are swallowing (which is right to do given how critical it is that we suppress them). These now temporarily clear out the ENV values that we're using to re-raise so we can check the unmodified behavior.

PragTob:
looks acceptable to me - I won't merge so you can have your after sleep look at it and maybe inspiration struck! :)

jasonrclark:
No bright ideas... merging!

Issue #1319, opened 2016-12-07
NullPointerException on second call to Shoes.app in same script on OS X


jayjlawrence:
I figured the my test script ( [display_npe.rb.txt](https://github.com/shoes/shoes4/files/636810/display_npe.rb.txt) ) would be acceptable. Basically it is (minus body code):

    app1 = Shoes.app { }
    app2 = Shoes.app { }

I get an NPE ( [exception.txt](https://github.com/shoes/shoes4/files/636826/exception.txt) ) when creating the 2nd app:

`Unhandled Java exception: java.lang.NullPointerException
java.lang.NullPointerException: null
    setNeedsDisplay at org/eclipse/swt/widgets/Widget:-1
         windowProc at org/eclipse/swt/widgets/Display:-1
       objc_msgSend at org/eclipse/swt/internal/cocoa/OS:-2
    setDocumentView at org/eclipse/swt/internal/cocoa/NSScrollView:-1
          setZOrder at org/eclipse/swt/widgets/Shell:-1
       createWidget at org/eclipse/swt/widgets/Control:-1
       createWidget at org/eclipse/swt/widgets/Scrollable:-1
             <init> at org/eclipse/swt/widgets/Shell:-1
             <init> at org/eclipse/swt/widgets/Shell:-1
Unhandled Java exception: java.lang.NullPointerException
        newInstance at java/lang/reflect/Constructor.java:423
  newInstanceDirect at org/jruby/javasupport/JavaConstructor.java:294
        newInstance at org/jruby/RubyClass.java:1008
               call at org/jruby/RubyClass$INVOKER$i$newInstance.gen:-1`

I poked around in the shoes code - my hunch is that I'm irritating a SWT glitch (maybe calling it a bug is too strong). There was a somewhat similar report (only I could find) here: https://www.eclipse.org/forums/index.php/t/205090/

I did check the value of ::SWT.display that is used when creating the second app and it is returning the same object id for Java::OrgEclipseSwtWidgets::Display for both the first and second runs.

a) is this correct  usage of Shoes.app?
b) is there a way to avoid touching this issue?

Thanks!
Jay

Environment info - jruby 9.1.6.0, swt gem 4.4, java version "1.8.0_74", OS X 10.11.6

jasonrclark:
Hi @jayjlawrence! We were actually discussing this over in #1298. It seems like we're leaning toward saying that having two top-level apps like that isn't currently supported.

[...]

PR #1328, opened 2016-12-14
Warn on second Shoes.app invocation

### LINKED ISSUE #1298: Can't run a second app after main app stops
Don't know whether we want to support this or not, but if you run this code, it crashes after you close the first app:

```
Shoes.app do
  para "first"
end

Shoes.app do
  para "second"
end
```

In Shoes 3 it behaves in a further different way--it will just start each app immediately, without blocking on the first app running.

jasonrclark:
Fixes #1298 and #1319.

As discussed in those issues, we aren't really looking to support the pattern of having two top-level `Shoes.app` calls in an app. If you need a second window, you can always do that by calling `Shoes.app` within your app block.

This PR tracks when we close out, and if another attempt to start an app is made, we give a friendly error message and exit (with non-zero status code for you POSIX fans).

### DIFF
--- a/shoes-swt/lib/shoes/swt/app.rb
+++ b/shoes-swt/lib/shoes/swt/app.rb
@@ -22,6 +22,21 @@ class Shoes
       attr_reader :dsl, :real, :shell, :click_listener
 
       def initialize(dsl)
+        if self.class.main_app_closed?
+          Shoes.logger.error <<-EOS
+Sorry! You can't start another Shoes.app at after the main one has already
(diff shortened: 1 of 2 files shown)


PragTob:
Could we test this in some way? Would be good to have imo?

Otherwise, looks good!

jasonrclark:
Specs, excellent idea! 😊 

It's a pretty mock-heavy test as such things go, but it validates what we're doing here so 👌

PragTob:
:+1:

PR #1329, opened 2016-12-15
Set overall app name

### LINKED ISSUE #1259: Look more like a Mac app
Currently the menu bar shows "SWT" when we're running a Shoes app. Also Command+Tab lists it as "java", and it shows as "java" in the dock.

It'd be nice to look more like a Mac app. Wonder whether other OS's have similar less-than-native-looking behavior too.

Couple references I dug up that look pertinent:

* http://stackoverflow.com/a/6700253 (confirmed in brief test this worked if added before [this point in the SWT loading]( https://github.com/shoes/shoes4/blob/d6c56ecbd1432bce9a18c419993cdcae4405efbd/shoes-swt/lib/shoes/swt.rb#L50)) 
* http://alvinalexander.com/java/java-application-name-mac-menu-bar-menubar-class-name

jasonrclark:
Fixes #1259 (as much as we can... see notes over there for more 😭 )

This must be done prior to initializing the SWT display, so we have to set it to something generic and can't use the DSL level title.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -55,8 +55,12 @@ Metrics/LineLength:
 
 # Offense count: 75
 # Configuration parameters: CountComments.
+#
+# Excludes long central SWT require file since they're long and that's fine!
 Metrics/MethodLength:
   Max: 86
+  Exclude:
+    - shoes-swt/lib/shoes/swt.rb
 
 # Offense count: 2
 # Configuration parameters: CountComments.
--- a/samples/README
+++ b/samples/README
@@ -67,6 +67,7 @@ simple_slide.rb
 simple_sound.rb
 simple_stack_flow_buttons.rb
 simple_stripes.rb
+simple_system_background.rb
 simple_translate.rb
 simple_text_movement.rb
 simple_timer.rb
--- a/samples/simple_system_background.rb
+++ b/samples/simple_system_background.rb
@@ -1,0 +1,4 @@
+Shoes.app do
+  fill system_background
+  title "This should be in the default system background color for your OS", align: 'center'
+end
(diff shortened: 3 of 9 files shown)


PragTob:
> This must be done prior to initializing the SWT display, so we have to set it to something generic and can't use the DSL level title.

Do we? I mean there is the `Shoes.app title: "lalala" do` or something that we could theoretically do before ever initialized SWT or what's the problem?

jasonrclark:
The `Shoes.app` call actually happens a long while after this--this is during the backend initialization, which is [called from the CLI runner](https://github.com/shoes/shoes4/blob/master/shoes-core/lib/shoes/ui/cli.rb#L11) before we even load the user's application code.

Maybe the font and system color loading could be lazily done later, once we've actually gotten into the `Shoes::Swt::App` to let us alter the title based on the options. Let me fiddle with it tonight and see how it goes.

jasonrclark:
Thanks for the nudge @PragTob... ended up being able to shuffle things around enough to enable the desired behavior. From the commit message:

> Once `::Swt.display` is touched, we can't set the title and have it apply
> to the Mac menu bar anymore (it does show in the window title for
> whatever reason). So, we need to postpone anything that touches the
> display until `::Shoes::Swt::App` actually has a chance to get started.
> 
> Turns out this is easy enough if we just postpone loading fonts and
> system colors. This also simplifies the font loading some, as it was
> making this weird transit through the DSL, despite being initiated in
> the SWT/backend side of the world. Now we just admit that the backend
> must load fonts on its own timeline and move on with our lives.

Also added a sample using `system_background` which was present nowhere else in the samples (although it is exercised on default window setup, but felt good to get it explicitly called out)

PragTob:
I love how little LoC/diff impact seemingly big changes have, well done! :heart:

PR #1330, opened 2016-12-16
Set size when updating position too

### LINKED ISSUE #1323: Button doesn't respect height on Mac
This program:

```
Shoes.app do
  button "haha", height: 100 do
    alert "whatever"
  end
end
```

results in this on the Mac:

image

Look like we're getting appropriate positioning to some extent, but the element just isn't the height it's supposed to be!

jasonrclark:
Fixes #1323

For reasons beyond my understanding, SWT on Mac seems to tweak button
heights when we `set_location` on them. Forcing the size again keeps
things happily proportioned as we'd like.

Have run through samples on the Mac and they're 👌 . ~~Checking on VMs for
other OSs to make sure this doesn't have negative side-effects there.~~
Windows and Ubuntu look good for at least a few samples with the effected
element types. Calling it 🙆

### DIFF
--- a/shoes-swt/lib/shoes/swt/common/update_position.rb
+++ b/shoes-swt/lib/shoes/swt/common/update_position.rb
@@ -7,6 +7,10 @@ class Shoes
         def update_position
           unless @real.disposed?
             @real.set_location dsl.element_left, dsl.element_top
+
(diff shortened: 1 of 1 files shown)


PragTob:
:'( but :+1:

PR #1336, opened 2016-12-20
Make arrow API compatible with Shoes 3

### LINKED ISSUE #1332: Arrow compatibility with Shoes 3
Reported via https://github.com/shoes/shoesrb.com/pull/78

This program runs fine with Shoes 3 (image below). The incomplete elements just don't paint (although if you later set all the necessary properties, they appear... this may be a further incompatibility with Shoes 4 drawing vs Shoes 3 to investigate!)

Shoes 4 just bails with `ArgumentError` problems 😢 

```
Shoes.app do
  arrow fill: blue
  arrow 50, fill: blue
  arrow 50, 50, fill: blue
  arrow 100, top: 100, width: 100, fill: green
  arrow 200, 200, width: 100, fill: yellow
  arrow 300, 300, 100, fill: red
  arrow left: 400, top: 400, width: 100, fill: orange
end
```

image

jasonrclark:
Fixes #1332 

As reported in that issue, we had some incompatibilities with the Shoes 3 API on `arrow`. In particular, tutorials called it with just an options hash which Shoes 4 wasn't allowing. ⬇️ 

Turns out on further testing Shoes 3 allows any number of arguments you want! Values that aren't provided explicitly are defaulted to what's in the options hash (and treated as 0/false if missing from there). :zero: 

I've allowed some of that flexibility, but deviated in one spot from precisely matching Shoes 3. If you provide *too many* explicit arguments, I'll give you an `ArgumentError`. It seems kinder than allowing blatantly bad data. 🔪 

I also chose in the method docs not to individually call out every variation that's actually supported, but just the full and options-only flavors. 📚 

Note that the behavior I saw when not giving enough args (where the arrow was "invisible") is just a consequence of the `width` being 0. I didn't even need to code anything explicitly to get Shoes 4 behaving the same--it just fell out of the implementation. 🎉 

This was partially missed because I neglected to add DSL specs for `arrow` (d'oh!) which I've added in all their spec-ish glory. 👼 

Based on @snood1205's fine footwork on the tutorial, I suspect that even the DSL methods we've got covered by specs have gaps though. Before `pre8` I'm going to do a comprehensive survey of all of them and get our 🏠 in order once and for all.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -66,6 +66,8 @@ Metrics/MethodLength:
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
   Max: 267
+  Exclude:
+    - shoes-core/lib/shoes/dsl.rb
 
 # Offense count: 13
 # Configuration parameters: CountKeywordArgs.
--- a/shoes-core/lib/shoes/dsl.rb
+++ b/shoes-core/lib/shoes/dsl.rb
@@ -308,15 +308,37 @@ class Shoes
       Shoes::Sound.new @__app__, soundfile, opts, &blk
     end
 
-    # Creates an arrow centered at (left, top)
+    # Creates an arrow
     #
-    # @param [Integer] left the x-coordinate of the element center
-    # @param [Integer] top the y-coordinate of the element center
-    # @param [Integer] width width of the arrow
-    # @param [Hash] opts Arrow style options
-    # @option opts [Integer] rotate (false)
-    def arrow(left, top, width, styles = {}, &blk)
-      create Shoes::Arrow, left, top, width, styles, blk
+    # @overload arrow(left, top, width, opts)
+    #   Creates an arrow centered at (left, top)
+    #   @param [Integer] left the x-coordinate of the element center
+    #   @param [Integer] top the y-coordinate of the element center
+    #   @param [Integer] width width of the arrow
+    #   @param [Hash] opts Arrow style options
+    #   @option opts [Integer] rotate (false)
+    # @overload arrow(opts)
+    #   Creates an arrow using values from the opts Hash.
+    #   @param [Hash] opts
+    #   @option styles [Integer] left (0) the x-coordinate of the top-left corner
+    #   @option styles [Integer] top (0) the y-coordinate of the top-left corner
+    #   @option styles [Integer] width (0) the width
+    #   @option styles [Integer] rotate (false)
+    def arrow(*args, &blk)
+      opts = style_normalizer.normalize pop_style(args)
+
+      left, top, width, *leftovers = args
+
+      message = <<EOS
+Too many arguments. Must be one of:
+  - arrow(left, top, width, [opts])
+  - arrow(left, top, [opts])
+  - arrow(left, [opts])
+  - arrow([opts])
+EOS
+      raise ArgumentError, message if leftovers.any?
+
+      create Shoes::Arrow, left, top, width, opts, blk
     end
 
     # Creates an arc at (left, top)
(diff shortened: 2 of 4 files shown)


.rubocop_todo.yml
```diff
@@ -66,6 +66,8 @@ Metrics/MethodLength:
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
   Max: 267
+  Exclude:
+    - shoes-core/lib/shoes/dsl.rb
```
PragTob: There might or might not be an issue for this, But I'd actually like to slice the dsl file up into multiple modules... like I dunno `Art`, `TextBlock` and other related/grouped methods. We can discuss some other place :)

shoes-core/lib/shoes/dsl.rb
```diff
@@ -308,15 +308,37 @@ def sound(soundfile, opts = {}, &blk)
...
-    # @param [Integer] width width of the arrow
-    # @param [Hash] opts Arrow style options
-    # @option opts [Integer] rotate (false)
-    def arrow(left, top, width, styles = {}, &blk)
-      create Shoes::Arrow, left, top, width, styles, blk
+    # @overload arrow(left, top, width, opts)
+    #   Creates an arrow centered at (left, top)
+    #   @param [Integer] left the x-coordinate of the element center
+    #   @param [Integer] top the y-coordinate of the element center
+    #   @param [Integer] width width of the arrow
```
PragTob: width width

shoes-core/spec/shoes/shared_examples/dsl/arrow.rb
```diff
@@ -0,0 +1,96 @@
...
+  let(:left)   { 40 }
+  let(:top)    { 30 }
+  let(:width)  { 12 }
+
+  context "no arguments" do
+    subject(:arrow) { dsl.arrow }
+
+    let(:left)  { 0 }
+    let(:top)   { 0 }
+    let(:width) { 0 }
```
PragTob: nice formatting :+1:

shoes-core/spec/shoes/shared_examples/dsl/arrow.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    include_examples "arrow dimensions"
```
PragTob: huh, never seen this. I think we always used to do `it_behaves_like` but this seems fine with me as well also I don't think we NEED conformity here :)

shoes-core/lib/shoes/dsl.rb
jasonrclark: Actually not a typo--first word is parameter, everything remaining is description. Could maybe add intervening "the" like other descriptions so it doesn't look mangled though.

shoes-core/spec/shoes/shared_examples/dsl/arrow.rb
jasonrclark: Oddly, I cribbed it from Shoes :) https://github.com/shoes/shoes4/blob/master/shoes-core/spec/shoes/shared_examples/dsl/rect.rb

Honestly, though, I don't think either of those really express these specs the way I'd like. Will fuss with it further tonight and see if I can get something I like better.

jasonrclark:
All right @PragTob, made a major revision of the specs that got rid of the weird example includes, but also ended up dropping the whole context and subject business.

I also ended up co-locating the values directly into the tests, as I feel it's easier to glance at a test and see that it's right, vs let's at the top for part of the values cascading down.

Happy to discuss further whether this is preferable or not before moving onto other DSL methods as I intend to make them more consistent as I fix regressions amongst them.

PragTob:
Love it how it is more concise now, less magic and more readable! :)

PR #1339, opened 2016-12-20
Fix order dependency in SWT font specs

jasonrclark:
Got a failure on Travis which was from order dependencies in the font testing after we changed things around. Bummer that earlier testing didn't catch it, but ¯\_(ツ)_/¯ 

Seed values which gave us problems: 38504, 24610

EDIT: Also, first branch after I changed Travis to not build on pushes, but only on PRs. I noticed that I was spending a lot of waiting time on duplicate builds against branches I pushed up that would then rebuild for the PRs. Lighten the load for Travis, shorten my waiting 🕐 👍


PragTob:
:+1:

Issue #1250, opened 2016-10-24
The download() method does not respect :headers, :body and :method styles


spidzio:
This is what manual says.

![image](https://cloud.githubusercontent.com/assets/2708775/19646644/aa18642a-99fa-11e6-98bc-f7f0ba196d10.png)

For making http requests `open` method is used from `open-uri`. That method accepts only additional http headers (which are not passed in this case) and several other options related to proxy and http basic authentication. So it doesn't even seems to be possible to pass these styles somehow.

The issue is checked on simple web service, and manipulation of the `:headers`, `:body` and `:method` styles does't give any effect.

PragTob:
Hi there,

thanks for pointing this out. Indeed, it's not implemented yet. For those interested it's implemented in [download](https://github.com/shoes/shoes4/blob/master/shoes-core/lib/shoes/download.rb).

Sadly, right now no one is actively working on shoes4 so it is also unlikely to make it in the foreseeable future. Sorry :(

Cheers,
Tobi

jasonrclark:
Closed by #1343

PR #1343, opened 2016-12-23
New Download implementation

### LINKED ISSUE #1026: Downloading doesn't work from an http:// URL that redirects to an https:// URL
As originally reported in #911, and made more visible to users in #1025. We should handle this case without raising an exception. Probably requires either custom http handler or incorporating an external http lib. Would be great to coordinate with asset downloading in shoes/furoshiki as well.

jasonrclark:
Ok, so this one seem a little crazy, but I ripped out our use of `open-uri` and replaced it with a custom implementation using `Net::HTTP` (only ~50 lines!)

## What's That?
This introduces a separate `Shoes::HttpWrapper` class with a basic interface of `yield`ing each chunk in the response (so we can mark progress).

If it encounters a HTTP error code, it raises per what `open-uri` did (so we didn't have to change the `Shoes::Download` code too much).

It also allows up to 5 redirects. Didn't make it customizable because, eh, 5 is probably enough.

## Why!!
First off, `open-uri` just plain doesn't support changing method, or passing headers or a body. Given that, we can't do what the Shoes docs claim is supported with `open-uri` (plus I honestly find the whole `open` thing a little 😲 ).

Searching, I found (as allude to in the description at the top of `Shoes::HtmlWrapper`) every gem either lacked the "streaming" interface for progress reporting, or showed minimal signs of JRuby testing (i.e. no JRuby in `.travis.yml` and/or flippant remarks in README/issues :sob:)

Is this madness? You tell me, but it was fun to write, passes the tests, and looks 👌 so far! I've had plenty of experience writing stuff straight against `Net::HTTP` in my work on `newrelic_rpm`, so I at least feel 🎉 about it.

## Sharing is Caring
#1026 mentioned custom redirection in `furoshiki` that could use consolidation. Not sure how I feel about that since we're looking to keep dependencies clean. Would that mean yet another gem... `shoes-http`?

Maybe `furoshiki` could just use an existing `Net::HTTP` wrapper gem since it doesn't need progress, and redirection support is totally easy to find all over the place... just a matter of getting one that's happy on JRuby.

Fixes #1026 and #1250

## TODO

* [x] Verify existing samples that use `download` call
* [x] Fix redirect scheme checking to disallow dangerous changes (https -> http)
* [x] Testing, testing, testing!


shoes-swt/lib/shoes/swt/download.rb
```diff
@@ -10,11 +10,16 @@ def initialize(_dsl, _app)
...
-          @busy = false
+          actually_run_block(blk, result)
         end
       end
 
+      # Why a separate method? So RedrawingAspect can target it!
+      def actually_run_block(blk, result)
+        blk.call result
+        @busy = false
+      end
```
PragTob: :heart:

shoes-core/lib/shoes/download.rb
```diff
@@ -100,26 +103,25 @@ def default_error_proc
...
+    def mark_progress
+      @gui.busy = true
+      eval_block(@progress_blk, self)
     end
 
-    def finish_download(_download_data)
+    def should_mark_progress?(size)
+      !content_length.nil? &&
+        (size - transferred) > (content_length / UPDATE_STEPS) &&
+        !@gui.busy?
```
PragTob: while you're in here - maybe you remeber/know I just have the vague feeling that something went wrong whenever we updated progress too fast and then stuff crashed... if you remember, maybe a little why style comment for the real reason/link to issue/whatever?

PragTob:
So thanks for all this work! :green_heart: 

Regarding the gem question... if it is mildly appropriate to also solve our problem at furoshiki, I'd be in favor of creating a separate gem that we maintain (as you created it, on your account or shoes whatever is more to your liking). I'd do that, if the effort is not big just so we don't have to fight with another dependency (and users who want to use both don't have to fight with another dependency as well). I guess we'd probably have to add another more general interface that doesn't call a block all the time so that it's generally usable.

I guess overall that might be less long time effort, but of course that depends on the extra effort we'd have to do to make it fit the furoshiki purposes.

shoes-core/lib/shoes/download.rb
jasonrclark: I actually don't remember that at all, but I did wonder why we had the `busy?` flagging business. As I'm doing some testing I might rip it out and see if I can duplicate what was happening and comment with my findings.

shoes-core/lib/shoes/download.rb
```diff
@@ -140,9 +142,23 @@ def save_to_file(file_path)
       open(file_path, 'wb') { |fw| fw.print @response.body }
     end
 
-    def download_started(content_length)
-      @content_length = content_length
-      @started        = true
+    def download_started(response)
+      unless @started
```
jasonrclark: Yeah, it's slightly cryptic this way. I had a version where `HttpWrapper` had tons of callbacks and it wasn't pretty, but I think doing one for started might work. Will check it and see.

shoes-core/lib/shoes/download.rb
```diff
@@ -140,9 +142,23 @@ def save_to_file(file_path)
...
+        @started = true
+        @content_length = read_content_length(response)
+
+        @response.status = [response.code, response.message]
```
jasonrclark: Yup, it's about compatibility. It was leaning on the [`OpenURI::Meta`](https://ruby-doc.org/stdlib-2.3.0/libdoc/open-uri/rdoc/OpenURI/Meta.html) object which has this odd form.

shoes-core/lib/shoes/download.rb
```diff
@@ -140,9 +142,23 @@ def save_to_file(file_path)
...
+        response.each_header do |key|
+          @response.headers[key] = response[key]
+        end
+
+        mark_progress
```
jasonrclark: This is another concession to compatibility, although one that I could be convinced into changing. `open-uri` invokes the progress proc once immediately when thing are starting up.

We had specs directly checking that. Open question whether that's just over-testing (since I wrote it, I suspect it might be), but it was reason enough for me to just match it to start with.

What do you think? Would we be happy with the slightly different progress triggering order of just doing the later progress calls?

shoes-core/lib/shoes/http_wrapper.rb
```diff
@@ -0,0 +1,55 @@
...
+#
+#   1) Reading in chunks to trigger progress
+#   2) Allow for graceful redirection
+#
+# Most of the gems out there do 2) but not 1), or do both but are untested on
+# JRuby.
+#
+class Shoes
+  class HttpWrapper
+    def execute(url, meth, body, headers = {}, redirects_left = 5, &blk)
```
jasonrclark: Funny to hear you mirror things I thought about precisely while doing the work ;)

An earlier rev of this had state with the object, so it started life as a true object. I left it when I considered a class method because it "might" want state, but after thinking a little longer, I think it's be better as a class method. Will revise.

shoes-core/lib/shoes/http_wrapper.rb
```diff
@@ -0,0 +1,55 @@
...
+      uri = URI(url)
```
jasonrclark: There is `URI.parse`, but for some reason this usage seems perfectly normal to me. Maybe I've been [watching too much Rubytapas](http://www.virtuouscode.com/2012/05/07/a-ruby-conversion-idiom/) as Avdi has some good material explaining these funky methods.

shoes-core/lib/shoes/http_wrapper.rb
```diff
@@ -0,0 +1,55 @@
...
+      Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == "https") do |http|
```
jasonrclark: I can buy it.

shoes-core/lib/shoes/http_wrapper.rb
```diff
@@ -0,0 +1,55 @@
...
+        http.request(request) do |response|
+          case response
+          when Net::HTTPRedirection
+            ensure_can_redirect(response, redirects_left)
+            next_location = URI(response["Location"])
+            execute(next_location, "GET", nil, headers, redirects_left - 1, &blk)
+          when Net::HTTPSuccess
+            response.read_body do |chunk|
+              yield response, chunk
+            end
```
jasonrclark: Yet again, great minds think alike :)

I originally had them in separate methods, but ran into issues with some of the earlier control-flow at the time (I had redirects happening via loop rather than recursion at the time). I see how to move them back now, so I'll do that.

shoes-core/lib/shoes/http_wrapper.rb
```diff
@@ -0,0 +1,55 @@
...
+          else
+            raise "#{response.code} #{response.message}"
+          end
+        end
+      end
+    end
+
+    def build_request(uri, meth, body, headers)
```
jasonrclark: Honestly everything except `execute` (probably should rename...) can be `private`. 👍

jasonrclark:
So `furoshiki`-wise, this does pretty much exactly the sort of redirection we need there, and adding an interface method to gather up the chunks into one output would be simple enough. Will see how things go and weigh whether to merge and extract later or not, mostly based on how I feel at the moment 😆

shoes-core/lib/shoes/http_wrapper.rb
jasonrclark: Ah, also just realized one of the nice bits about `URI()`--if you hand it something that's a `URI` instance already, it quietly hands it along. `URI.parse` raises if the thing isn't a string to parse.

Still cool with the change, but I think that's the idea behind those conversion methods.

jasonrclark:
Pushed up changes from your comments @PragTob... the file is now ~90 lines, but I think it's much clearer, thank you!

Want to do more testing that I don't have gumption for tonight, but happy with where it's headed.

shoes-core/lib/shoes/download.rb
PragTob: hm, if we wrote a spec directly for it then I'm allfor keeping it. Past us, both in writing and reading probably had some reason for it :)

shoes-core/lib/shoes/http_wrapper.rb
PragTob: It's probably fine... I know Avdi likes them `Array(...)` methods etc. and that's fine and I think we should use them more, as URI looks more like a constant (but it's probably a class/module) it looked odd/firghtening/why-ish to me but I'm good with it :)

shoes-core/lib/shoes/http_wrapper.rb
```diff
@@ -14,42 +16,72 @@
...
+      def schemes_mismatch?(original_uri, new_uri)
+        original_uri.scheme != new_uri.scheme
+      end
+
+      def allowed_scheme_change?(original_uri, new_uri)
+        original_uri.scheme == "http" && new_uri.scheme == "https"
+      end
+
+      def handle_success(response, started_proc)
+        started_proc&.call(response)
```
PragTob: nifty! :clap:

jasonrclark:
Totally enjoying the exchanges too! While it's nice when bug fixes are plain and obvious, it's fun to do something that we can both learn on ✨ 

A couple of your comments had me thinking about the shape of the interface. Specifically, I liked the `read_chunks` because it indicated how the reading happens, but I hear you that it's not as clear with the class name. I also wasn't loving the growing set of arguments, which lead me to this idea--how about changing back to an object :)

Specifically, I'm thinking we rename the class `Shoes::HttpRequest`. It gets instantiated with the `start_proc` and sets the redirect count at that point. From there, you call `read_chunks` (and later if we support furoshiki, we can provide a simple `read` that tallies the stuff up.

This deals neatly with the naming (for me at least) and gets the parameter list shortened down. What do you think?

Off to Christmas Eve party 🎄, and unlikely to get back around to this for a bit, but vacation is coming and 👟  work is my relaxation plan!

PragTob:
Sounds like a plan! :+1:

jasonrclark:
Latest revisions based on what I discussed earlier. Definitely like this better now that I've plumbed it through. Objects FTW!

Will do some further testing that I want to try, but in the next couple days should be ready to pull the WIP off this and :shipit:

jasonrclark:
All right, actually got to doing some testing and I'm 👌 with where it's at if you're good @PragTob.

PragTob:
:heart: 

Shall we file an issue to cross-check using it from furoshiki/putting it into its own gem?

jasonrclark:
Thanks! Filed #1346 to track it.

PR #1348, opened 2016-12-29
Make rotate direction compatible with Shoes 3

### LINKED ISSUE #1288: Rotate reversed vs Shoes 3?
In writing arrow, I noticed that we might have rotate flipped?

**Shoes 3**
image

**Shoes 4**
image

```
Shoes.app do
  rotate 45
  rect 50, 50, 100, 50
end

```

Negating the rotation gets it pointed the same way in Shoes 4 as in Shoes 3.

jasonrclark:
Fixes #1288

After experimenting further with `rotate`, it's pretty plain to me that we're going the opposite direction that Shoes 3 assumes consistently for every single art element.

Anecdotally, when I googled about "rotating in degrees", the top result was https://www.mathsisfun.com/geometry/degrees.html which shows an animation working precisely how Shoes 3 does (starting from middle-right, rotating counter-clockwise with increase in degrees).

While it may be opposite of what some of us were expecting, this seems like enough of a compatibility issue that I'm willing to flip it.

This fix addresses the 3 art elements that currently support `rotate` (`rect`, `oval`, and `arrow`). Other elements that don't respond yet to `rotate` will just get addressed as they support it.

### DIFF
--- a/shoes-swt/lib/shoes/swt/common/painter.rb
+++ b/shoes-swt/lib/shoes/swt/common/painter.rb
@@ -80,9 +80,9 @@ class Shoes
           if block_given?
             begin
               transform = ::Swt::Transform.new Shoes.display
-              reset_rotate transform, graphics_context, angle, left, top
-              yield
               reset_rotate transform, graphics_context, -angle, left, top
+              yield
+              reset_rotate transform, graphics_context, angle, left, top
             ensure
               transform.dispose unless transform.nil? || transform.disposed?
             end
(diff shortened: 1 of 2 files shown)


PragTob:
Thanks for giving this so much thought!

Only thing I'm missing here would be a spec as we are certain we want this behavior and we want it for compatibility etc. :)

I'd also appreciate a link to the website in the commit message and/or maybe above/in the spec.

jasonrclark:
Ah yes, specs! Turns out we had some specs that run through that path, they just weren't being very careful in their assertions. I tightened those up and we're 👌 

Also commented at the spot where we do the negation. Hopefully will clarify. Assuming I continued online for a bit, will probably merge this once it's 🍏 as I've got some follow-on rotate stuff to push up.

PR #1349, opened 2016-12-30
Apply rotate consistently to all art elements

### LINKED ISSUE #699: Fully Implement App#rotate (and make it available as a style option)
From what I gathered (not tried out yet) [App#rotate](http://shoesrb.com/manual/Art.html#rotate) isn't fully working.

Optionally, which could be moved to another issue/PR, it'd be nice if rotate were available as a style option. That would be an extension of the DSL which I'd like but could wait for 4.1.0

e.g.:

`rect 100, 50, 200, 300 rotate: 45`

cc: @ashbb @KCerb

This issue was "born" from #698 

Cheers,
Tobi

jasonrclark:
Fixes #699

Previously the kind of odd `Shoes::Common::Rotate` module would give you the required method `needs_rotate?`, but only to return false. You had to re-plumb that method yourself to actually rotate (which `rect`, `oval` and `arrow` did, but other art elements did not).

I've flipped that now so `Shoes::Common::Rotate` actually supports real rotation, and we have a stubbed `needs_rotate?` that returns false on `UIElement` as the fallback instead.

As part of this, I also did something I'd wanted to for a while--started a new `ArtElement` module which can apply common across, not-surprisingly, the art elements :)  Follow-on work will be coming to make better use of this, but for now it lets us easily plumb rotation across the elements.

### DIFF
--- a/shoes-core/lib/shoes/common/art_element.rb
+++ b/shoes-core/lib/shoes/common/art_element.rb
@@ -1,0 +1,8 @@
+class Shoes
+  module Common
+    module ArtElement
+      include Common::UIElement
+      include Common::Rotate
+    end
+  end
+end
--- a/shoes-core/spec/shoes/common/rotate_spec.rb
+++ b/shoes-core/spec/shoes/common/rotate_spec.rb
@@ -1,13 +1,21 @@
 require 'spec_helper'
 
 describe Shoes::Common::Rotate do
-  let(:test_class) { Class.new { include Shoes::Common::Rotate } }
-
   subject { test_class.new }
 
   describe '#needs_rotate?' do
-    it 'defaults to falsey value' do
-      expect(subject.needs_rotate?).to be_falsey
+    let(:test_class) do
+      Class.new do
+        include Shoes::Common::Rotate
+
+        def rotate
+          45
+        end
+      end
+    end
+
+    it 'defaults to truthy value' do
+      expect(subject.needs_rotate?).to be_truthy
     end
   end
 end
--- a/shoes-core/spec/shoes/common/rotate_spec.rb
+++ b/shoes-core/spec/shoes/common/rotate_spec.rb
@@ -6,15 +6,23 @@ describe Shoes::Common::Rotate do
   describe '#needs_rotate?' do
     let(:test_class) do
       Class.new do
+        attr_accessor :rotate
         include Shoes::Common::Rotate
-
-        def rotate
-          45
-        end
       end
     end
 
-    it 'defaults to truthy value' do
+    it 'is falsey if nil' do
+      subject.rotate = nil
+      expect(subject.needs_rotate?).to be_falsey
+    end
+
+    it 'is falsey if 0' do
+      subject.rotate = 0
+      expect(subject.needs_rotate?).to be_falsey
+    end
+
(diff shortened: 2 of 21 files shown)


shoes-core/lib/shoes/common/art_element.rb
```diff
@@ -0,0 +1,8 @@
+class Shoes
+  module Common
+    module ArtElement
```
PragTob: :heart:

shoes-core/spec/shoes/common/rotate_spec.rb
```diff
@@ -1,13 +1,21 @@
...
+      Class.new do
+        include Shoes::Common::Rotate
+
+        def rotate
+          45
+        end
+      end
+    end
+
+    it 'defaults to truthy value' do
```
PragTob: well, the method checks more than that, it checks if it is non zero and present. E.g. as for test cases to cover this appropriately I think we need

* some number
* nil
* 0

PragTob: Ok have just seen the object with rotate :D still lacking the `0` case imo!

shoes-core/spec/shoes/common/rotate_spec.rb
```diff
@@ -1,13 +1,21 @@
 require 'spec_helper'
 
 describe Shoes::Common::Rotate do
-  let(:test_class) { Class.new { include Shoes::Common::Rotate } }
-
   subject { test_class.new }
 
   describe '#needs_rotate?' do
-    it 'defaults to falsey value' do
-      expect(subject.needs_rotate?).to be_falsey
```
PragTob: this should probably be tested somewhere in UIElement?

jasonrclark:
All right, beefed up the rotation tests. Now the shared example and the direct spec both check nil, 0 and non-zero cases.

Also introduced a `UIElement` test to check the direct methods it provides. Intent is *not* to test all the various things it pulls in--that actually gets rough to do against a fake class without a lot of plumbing.

Our general approach of 1) test the module directly on a fake and 2) use shared examples to check use of the module in real elements feels right to me now that it's getting cleaned up and made more consistent.

What do you think @PragTob?

shoes-core/spec/shoes/common/ui_element_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    it { is_expected.to respond_to(:update_stroke) }
+  end
+
+  describe "#needs_rotate?" do
+    before do
+      # Add faux rotate accessor to prove it really doesn't support rotation
+      class << subject
+        attr_accessor :rotate
+      end
+    end
```
PragTob: minor thing but I feel like this is overdoing it, couldn't we just stub the `rotate` method to have the same effect with less code and more in line with the rest of the code base or am I missing something? :)

If we use the instance_variable we could revert to using the method or use instance_variable_set imo

jasonrclark: Yeah, you're totally right. Since I was in the mode of creating fake classes, got a little carried away 😨

jasonrclark:
Pushed up last spec change. Will merge on 💚

PR #1350, opened 2016-12-30
Consolidate shared modules in ArtElement

jasonrclark:
This builds off #1349 and the diff contains those changes as well for now. Will rebase before merging once that PR is in, but just couldn't wait to show the fruit off from making `Shoes::Common::ArtElement`

This pulls everything I could find that was common across `UIElement`, `BackgroundElement`, and `ArtElement` up into those modules.

It also then rolls up the various repeated `it_behaves_like` blocks for `ArtElement` includers so we only have one thing to put in a spec that says "I'm an art element, treat me as such."

Over in the `UIElement` part of the world, I saw less to collapse besides pulling `Common::Style` inclusion up since everyone wants that. Similarly for `BackgroundElement`, although it got a bit of love.

While not quite the deletion count I was expecting, this trims things down, and especially trims it down in the main classes and their specs. What do you think?

## TODO

* [x] Rebase after #1349 is merged
* [x] Thorough sample testing to make sure we haven't inadvertently broken something!


shoes-core/lib/shoes/arc.rb
```diff
@@ -1,15 +1,9 @@
...
-    include Common::Hover
-    include Common::Fill
-    include Common::Stroke
-    include Common::Style
-    include Common::Translate
+    include Common::ArtElement
 
     style_with :angle1, :angle2, :art_styles, :center, :common_styles, :dimensions, :radius, :wedge
-    STYLES = { wedge: false }.freeze
+    STYLES = { wedge: false, fill: Shoes::COLORS[:black] }.freeze
```
PragTob: I'm a fan of `.fetch` these days when we really need something to be there!

jasonrclark:
Yeah, I initially just included `Common::Style` in `UIElement`, expecting it to work, but it went 💥 and after a moment's thought it was clear why... hence the comment. It really does boil down to how modules work, so I don't see a way around it with our current structure.

An alternative possibility would be to change `UIElement`, `BackgroundElement` and `ArtElement` into base classes that we derive from. Then our module application stays a single level deep.

What do you think? 🤔

PragTob:
> An alternative possibility would be to change UIElement, BackgroundElement and ArtElement into base classes that we derive from. Then our module application stays a single level deep.

To my mind it has been borderline weird that there is nothing like a `ShoesElement` base class and then `BackgroundElement`/`ArtElement` classes. I mean, we sort of do alright with modules and ruby does give us that flexibility but it really does feel weird. So far it hasn't really gotten into our way, now that it seemingly has I'm all for getting that into base classes. We are very unlikely to need another base class (so that we ought to "save" our one time inheritance card). So I'd say... **yay** - _but in another PR_ :)

PragTob:
Was about to merge but saw that thorough testing was outstanding

jasonrclark:
Cool... so we'll just roll with this and because I'm in the mood I'll probably turn back around after and plumb in the classes. Thanks @PragTob!

jasonrclark:
Pushed rebase of this, so should be clear to just the diffs for this PR now. Will get down to my sample testing and report back once that's done.

PragTob:
:+1: [supergeil](https://www.youtube.com/watch?v=jxVcgDMBU94) :+1:

PR #1364, opened 2017-01-09
Arc compatibility with Shoes 3

jasonrclark:
Standard Shoes 3 compatibility work, plus a few "fun" extras.

Noticed that scoping on blocks in the `ArcPainter` spec meant they weren't actually running. 😬 

Then I noticed divergent behavior from Shoes 3 -> 4 on `arc`. If the start and end angles are reversed, Shoes 4 will refuse to display anything. Shoes 3 more gracefully lets those reverse.

We also get further incorrect displays like from `Shoes::PI` -> `Shoes::TWO_PI` (see below)

Comments indicate there's differences in SWT vs Shoes direction of rotation. I suspect our math isn't right for all cases here, but it's too 💤 at night for me to figure it out.

```
Shoes.app height: 300 do
  arc 100, 50, 100, 50, Shoes::PI, 0, fill: orange
  arc 100, 100, 100, 50, Shoes::PI, Shoes::TWO_PI, fill: yellow
  arc 100, 150, 100, 50, Shoes::PI, Shoes::TWO_PI + Shoes::HALF_PI, fill: green
end
```

**Shoes 3**
image

**Shoes 4**
image


PragTob:
Thanks for working on this!

Just as a minor thought, we shouldn't work too hard to replicate some of the error/edge case behavior of Shoes 3. Chances are they are just random implementation quirks that aren't even intended and we start to try to reimplement those rather than big things :)

jasonrclark:
I'm with you--I don't necessarily feel like we need to replicate all the behavior.

That said, putting in two numbers that used to work and getting a blank screen would be pretty frustrating, so I'll see if I can make the basics of this behave nicely. If it gets too tough, I'll table it in this PR.

jasonrclark:
Progress on understanding, although my brain isn't quite up to dig into the math deeply tonight for implementing...

I've had inklings something was more dramatically misaligned between 3 and 4 than I was spotting. Tonight this testing code finally brought it clear:

```
Shoes.app do
  arc 100, 50, 100, 50, Shoes::HALF_PI, Shoes::PI, fill: orange, wedge: true
end
```

Results are...

**Shoes 3**
image

**Shoes 4**
image

This finally made sense! Shoes 3 is treating `angle1/2` as explicit start/stop locations around the circle. Shoes 4 is treating `angle1` as the starting point and `angle2` as _how far to sweep through the angle_! Both totally valid _but incompatible_ models for representing arcing! (After the fact confirmed by [the SWT documentation](http://help.eclipse.org/kepler/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fswt%2Fgraphics%2FGC.html&anchor=fillArc(int,%20int,%20int,%20int,%20int,%20int)))

Upshot, although I can't do the math in my head just yet, I can see that it won't be too much trouble to code up the necessary transformations from Shoes' start/stop to SWT's start/sweep and get things behaving themselves nicely... another night.

ccoupe:
If I understand things in Shoes3 (don't bet the farm)  Shoes::TWO_PI and HALF_PI  are radians,  not angles.

jasonrclark:
Yep @ccoupe, Shoes' interface in both 3 and 4 is in radians. `angle1` and `angle2` are just the style/accessor names.

For added fun, SWT under us in 4 actually does its work in degrees and in the opposite rotation for added whiplash :rage4: I'm not surprised things haven't lined up nicely in this corner of the code between the versions.

jasonrclark:
All right, finally got down to getting the math figured (plenty of comments in the painter where that mostly lives) and better spec'ing around those calculations in the `ArcPainter`. Those include lots of pictures, since I'm sure within days all of those angles I can actually reason about at the moment will totally fade.

There's very few samples that use arc directly, but they look good with these, and I've run through a ton of combinations in a personal little sample, comparing Shoes 3 to 4, and this all looks good.

(Hrmm, while I've been writing CI appears to have failed on a gem issue. Will investigate that further, but once that's resolved this should be good to 👟 )

jasonrclark:
Wow, that test failure was spectacularly gnarly!

Failure was during installing the `rainbow` gem which `rubocop` uses for doing ANSI color codes to the terminal. It complained about modifying a frozen string.

The string in question was in the rake command-line that Rubygems builds up for running a "native" gem extension that `rainbow` has for Windows. This turned out to be [an already fixed bug in Rubygems](https://github.com/rubygems/rubygems/pull/1819) that hasn't been released yet.

But why didn't this happen before? Well, that's because Bundler between 1.13 and 1.14 started passing `args` to the rake builder (seen in the linked PR) as `[nil]` instead of `[]`, which then tried to modify the string, which then hits the bug.

So, dropping back to Bundler 1.13.7 for the moment deals with it and once there's a newer Rubygems out with the fix we can move back to current, with [this reminder issue to help us follow through on doing it.](https://github.com/shoes/shoes4/issues/1372).

shoes-swt/lib/shoes/swt/arc_painter.rb
```diff
@@ -26,7 +26,33 @@ def draw(graphics_context)
...
                                     @obj.element_width - sw,
                                     @obj.element_height - sw,
-                                    @obj.angle1, @obj.angle2 * -1)
+                                    start_angle, sweep)
+        end
+      end
+
+      # Shoes angles to run clockwise, with 0 at the 3 o'clock position. SWT
+      # instead runs counter-clockwise from 3 o'clock, hence we negate the
+      # start angle to translate from Shoes -> SWT
```
PragTob: :+1:

shoes-swt/lib/shoes/swt/arc_painter.rb
```diff
@@ -26,7 +26,33 @@ def draw(graphics_context)
...
+      # Hence we end up calculating a negative (clockwise) sweep. There's a
+      # small case difference depending on if we're crossing over the 0 point.
+      #
+      # See the specs for lots of good examples, with pictures, of this logic.
+      #
+      def sweep
+        if @obj.angle1 <= @obj.angle2
+          @obj.angle1 - @obj.angle2
+        else
+          @obj.angle1 - @obj.angle2 - 360
```
PragTob: maybe put 360 in a constant? It's sort of clear from the context but I still like to be more explicit about it :)

shoes-swt/spec/shoes/swt/arc_painter_spec.rb
```diff
@@ -9,72 +9,207 @@
...
-        path = double('path')
-        allow(::Swt::Path).to receive(:new) { path }
-        args = [100, 200, width, height, 180.0, -90.0]
-        expect(path).to receive(:add_arc).with(*args)
-        sw = 10
-        args = [100 + sw / 2, 200 + sw / 2, width - sw, height - sw, 180, -90.0]
-        expect(gc).to receive(:draw_arc).with(*args)
-        subject.paint_control(gc)
-      end
+      sw = 10
```
PragTob: what's sw?

shoes-swt/spec/shoes/swt/arc_painter_spec.rb
```diff
@@ -9,72 +9,207 @@
...
+    specify "draws arc" do
+      expect(gc).to receive(:draw_arc)
+      subject.paint_control(event)
+    end
+
+    specify "translates DSL values for Swt" do
+      args = [left, top, width, height, -180, -270.0]
+      expect(gc).to receive(:fill_arc).with(*args)
+
+      sw = 10
```
PragTob: what's sw?

shoes-swt/spec/shoes/swt/arc_painter_spec.rb
```diff
@@ -9,72 +9,207 @@
...
+  describe "half-way around" do
+    #
+    #   ---------
+    #  /         \
+    # |           |
+    # |...........|
+    # |...........|
+    #  \........./
+    #   ---------
+    #
```
PragTob: sad to admit, but I don't get what your beautiful ascii art wants to tell me :crying_cat_face:

shoes-swt/spec/shoes/swt/arc_painter_spec.rb
jasonrclark: Hmm, actually don't know why it's called that (just modified the spec here from before). In the main code it's called that too, and looks like the line height? Will update.

shoes-swt/spec/shoes/swt/arc_painter_spec.rb
jasonrclark: So they're trying to visualize the dotted region as the part of the arc that we'll sweep through. This ones starts over on the right midpoint, then runs over to the left midpoint.

Any thoughts on how to convey this that would be clearer? I've found these type of drawings to be critical to me understanding the behavior as I work through it, but I admit I'm a bit deep in it at the moment so not surprised my scrawlings aren't totally clear :)

PragTob: I like your little visualizations, also over with text block. I suspected that they do something like you said but was not sure. Maybe a little description about what it's meant to convey above the first one is all we need? :)

jasonrclark: Will get it a solid description and let you see how that works for you.

jasonrclark:
Updated based on feedback. What do you think @PragTob?

PragTob:
:shipit:

Issue #1366, opened 2017-01-09
Rotate problem in samples/simple_gradient_shapes.rb


jasonrclark:
Noticed this in testing things... probably due to something in #1349 or the neighborhood.

```
$ bin/shoes samples/simple_gradient_shapes.rb

Unfortunately we have to swallow it because it causes odd failures :(
SWALLOWED PAINT EXCEPTION ON #<Shoes::Swt::Shape:0x749ab7b4> - go take care of it: undefined method `nonzero?' for [45, 300, 300]:Array
```

jasonrclark:
FWIW, I can only find reference in the Shoes manual to `rotate` accepting a number for degrees. Traced the sample back to [its origin](https://github.com/shoes/shoes4/commit/bd6cbfe859f036406883b1a6bc512e7cefa76bda), which had the array but also notes in the comment that `rotate` wasn't implemented yet...

Given that, this seems like just an error in the sample with incompatible code that was quietly not doing anything until we enable rotation on shapes.

PragTob:
So, what you are saying is we can remove the rotate call and close this? :D

jasonrclark:
Yep. Just ran out of steam to actually do it last night.

PR #1367, opened 2017-01-09
Shape API compatibility with Shoes 3

jasonrclark:
Now standard Shoes 3 compatibility work.

Took opportunity to make `Shoes::Shape` take explicit parameters the same way other elements do instead of passing things through via styles alone set in `dsl.rb`.

Noticed #1366 along the way, but that was happening on master, not just this branch.


shoes-core/spec/shoes/shape_spec.rb
```diff
@@ -115,4 +115,39 @@
...
+      expect(shape).to have_attributes(left: 10, top: 20)
+    end
+
+    it "doesn't like too many arguments" do
+      expect { dsl.shape 10, 20, 666 }.to raise_error(ArgumentError)
+    end
+
+    it "doesn't like too many arguments and options too!" do
+      expect { dsl.shape 10, 20, 666, left: -1 }.to raise_error(ArgumentError)
+    end
```
PragTob: I'd like a "just hash" example e.g. top/left in the hash

PR #1368, opened 2017-01-09
Fix problem with rake spec:package

jasonrclark:
Noticed that if I ran `rake spec:package` instead of `bundle exec rake spec:package` we'd get a failure about finding the `Pathname` constant.

Easy enough fix as we were already requiring it later in the spec helper... just not early enough if we're not running through Bundler.


PragTob:
Ha that always bugged me but I was like... fine ok I'll use bundler THIS TIME if I must.

PR #1370, opened 2017-01-16
Remove unnecessary rotate

jasonrclark:
Fixes #1366

Format is wrong, always had been, and not a Shoes 3 compatibility thing

PR #1371, opened 2017-01-16
Fixes for YAML.safe_load from new Rubocop

jasonrclark:
https://github.com/bbatsov/rubocop/blob/master/CHANGELOG.md#0470-2017-01-16

Since we don't check in `Gemfile.lock` (appropriately for a gem), we get all the latest Rubocop changes whenever they post (hurray?) since we don't bother with a version range in the `Gemfile` (which seems fine... when will we ever remember to update otherwise?)

This was the only breakage, and will fix #1370 on a rebase once this is merged.

### DIFF
--- a/samples/good_reminder.rb
+++ b/samples/good_reminder.rb
@@ -143,7 +143,7 @@ Shoes.app title: "A Gentle Reminder",
 
   def load
     if File.exist?(data_path)
-      @todo, @completed = YAML.load(File.open(data_path, 'r'))
+      @todo, @completed = YAML.safe_load(File.open(data_path, 'r'), [Date])
     else
       @todo = []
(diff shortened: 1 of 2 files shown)


PragTob:
We might want to discuss fixing the rubocop version though (I check for updates every now and then) - it could be frustrating for new contributors if their PR fails on some failing cop that they are absolutely not responsible for and I'd like to avoid that :)

jasonrclark:
You know, you're absolutely right @PragTob, I hadn't quite thought through the frustration, particularly for new folks, that would bring. Along with just pinning the versions, I've had a few thoughts in the meantime that'll help keep the updates fresh. Thanks for mentioning it!

PR #1373, opened 2017-01-24
Allow for skipping coverage via ENV

jasonrclark:
I noticed that in running individual specs, I was spending 2-3 seconds each time running coverage that I never actually look at.

This provides a way to opt out of running coverage. It's not default to avoid breaking CI and/or removing coverage for folks less familiar, but it will keep my feedback cycles shorter.

🏃 💨 👌


jasonrclark:
Sigh, build failures are from the bundler changes I need applied by #1364. Will rebase once that's landed.

PR #1375, opened 2017-01-25
Don't pin bundler as new rubygems version resolve the problem

### LINKED ISSUE #1372: Revert pinning of Bundler to 1.13.7
Bundler 1.14 started passing values along to the Rake builder used by the `rainbow` gem in a way that caused it to break on appending to a string literal (same issue as https://github.com/rubygems/rubygems/pull/1819).

Given the Rubygems update hadn't been released with that change, the best we could do was pin to the older Bundler, which didn't pass along values that caused the append to happen.

Should check in periodically to see whether the coast is clear to unpin Bundler on both AppVeyor and Travis.

PragTob:
* Fixes #1372 
Let's hope this works :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,8 @@ notifications:
     on_success: change # [always|never|change] # default: change
     on_failure: change # [always|never|change] # default: always
 before_install:
-  # Pinning to Bundler 1.13.7 until Rubygems with https://github.com/rubygems/rubygems/pull/1819
-  - gem uninstall -i /home/travis/.rvm/gems/jruby-9.1.7.0@global bundler
-  - gem install bundler -v "1.13.7"
+  - gem update --system
+  - gem install bundler
 before_script:
   - "export DISPLAY=:99.0"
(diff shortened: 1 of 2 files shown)


jasonrclark:
Cool thanks! You all right with us holding this until I get pre8 testing done and release this weekend?

PragTob:
@jasonrclark I'm cool with it, although I don't understand - this only affects our own CI builds it should not interfere with the release in any form or way?!

jasonrclark:
@PragTob If there is any regression to fix before release, I'd rather not risk any changes to CI that I have to debug instead. Little short on time and energy, but really want pre8 out this weekend.

Fwiw, even with latest Rubygems + latest bundler locally I saw some (as yet not debugged) weirdness, hence my hedging.

PragTob:
@jasonrclark it's fine then, we wait :) FWIW I took this approach from over at simplecov where it's been running smoothly so far :)

jasonrclark:
New 👟 out, ready for a merge!

PR #1376, opened 2017-01-29
Reset alpha on graphics context

### LINKED ISSUE #1190: image inherits alpha from last background call
This is a bit odd, and I suspect it's not an intended behavior (plus I couldn't find it in the docs), so I'm gonna guess it's a bug. Take a look at this little program (to be executed from the shoes4 dir):

``` ruby
Shoes.app do
   stack margin: 40 do
      stack{ background rgb(1.0,1.0,1.0,1.0) }
      stack do
         title "Now I'm here"
         image("samples/avatar.png")
         para "zoot!"
      end
      stack{ background rgb(1.0,1.0,1.0,0.3) }
      stack do
         title "Now... sorta"
         image("samples/avatar.png")
         para "zap!"
      end
      stack{ background rgb(1.0,1.0,1.0,0.0) }
      stack do
         title "Now I'm not"
         image("samples/avatar.png")
         para "zouk!"
      end
   end
end
```

This how it looks like:
screen shot 2015-12-06 at 5 33 57 pm

What this means is that the only way to put an image with `alpha=1.0` inside a container with `alpha < 1.0` is by adding an "invisible" `stack{ background rgb(1.0,1.0,1.0,1.0) }` before the image.

jasonrclark:
Fixes #1190 

As @PragTob suspected, we weren't getting the alpha reset on the graphics context after it was getting modified while painting faded elements.

This had a natural home in the method that sets up graphic context defaults.

Noticed along the way we were still using an old-school, Java-ish `setTransform` call, so tidied that up too.

### DIFF
--- a/shoes-swt/lib/shoes/swt/common/painter.rb
+++ b/shoes-swt/lib/shoes/swt/common/painter.rb
@@ -98,7 +98,7 @@ class Shoes
           transform.translate left, top
           transform.rotate angle
           transform.translate(-left, -top)
-          graphics_context.setTransform transform
+          graphics_context.set_transform transform
         end
       end
     end
--- a/shoes-swt/lib/shoes/swt/common/resource.rb
+++ b/shoes-swt/lib/shoes/swt/common/resource.rb
@@ -2,6 +2,8 @@ class Shoes
   module Swt
     module Common
       module Resource
+        OPAQUE = 255
+
         def reset_graphics_context(graphics_context)
           dispose_previous_contexts
           set_defaults_on_context(graphics_context)
@@ -15,6 +17,7 @@ class Shoes
         end
 
         def set_defaults_on_context(graphics_context)
+          graphics_context.set_alpha OPAQUE
           graphics_context.set_antialias(::Swt::SWT::ON)
           graphics_context.set_line_cap(::Swt::SWT::CAP_FLAT)
           graphics_context.set_transform(nil)
--- a/shoes-swt/spec/shoes/swt/text_block/painter_spec.rb
+++ b/shoes-swt/spec/shoes/swt/text_block/painter_spec.rb
@@ -13,7 +13,7 @@ describe Shoes::Swt::TextBlock::Painter do
   let(:graphics_context) do
     double("graphics context", set_antialias: nil, set_line_cap: nil,
                                set_transform: nil, clipping: nil,
-                               set_clipping: nil)
+                               set_clipping: nil, set_alpha: nil)
   end
 
   subject { Shoes::Swt::TextBlock::Painter.new(dsl) }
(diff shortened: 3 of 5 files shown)


shoes-swt/lib/shoes/swt/common/painter.rb
```diff
@@ -98,7 +98,7 @@ def reset_rotate(transform, graphics_context, angle, left, top)
           transform.translate left, top
           transform.rotate angle
           transform.translate(-left, -top)
-          graphics_context.setTransform transform
+          graphics_context.set_transform transform
```
PragTob: full new style would be `transform =` ;P

shoes-swt/lib/shoes/swt/common/resource.rb
```diff
@@ -2,6 +2,8 @@ class Shoes
   module Swt
     module Common
       module Resource
+        OPAQUE = 255
```
PragTob: :heart_eyes:

shoes-swt/lib/shoes/swt/common/painter.rb
PragTob: Hm I see that the other cases also seem to use `set_` so I'd make this an all or nothing thing, doesn't have to hold up this PR though

shoes-swt/spec/shoes/swt/text_block/painter_spec.rb
```diff
@@ -13,7 +13,7 @@
   let(:graphics_context) do
     double("graphics context", set_antialias: nil, set_line_cap: nil,
                                set_transform: nil, clipping: nil,
-                               set_clipping: nil)
+                               set_clipping: nil, set_alpha: nil)
```
PragTob: don't we have a spec somewhere that expects the right defaults to be set?

shoes-swt/lib/shoes/swt/common/painter.rb
jasonrclark: Yeah noticed that as well. Probably will turn them all to = form as long as it goes smoothly next time I'm coding.

shoes-swt/spec/shoes/swt/text_block/painter_spec.rb
jasonrclark: Good point. Will get that covered for real somewhere, as these really just stub unnecessary stuff for painters to run.

jasonrclark:
So, just pushed an update with a real set of specs covering `Shoes::Swt::Common::Resource`. This gets us the checks for the default values that we'd like.

Also took a stab briefly at changing the `set_` to equals form, and it quickly devolved into way more changes than I'd like to get in this PR. Will keep it mind for future revision, but not nearly as clean as I'd hoped for.

PragTob:
:tada:

PR #1377, opened 2017-02-02
Fix visibility to be set through styles

### LINKED ISSUE #1176: Didn't we want to handle all styles setting through style?
Hey @KCErb @jasonrclark 

question: I thought we wanted to handle all styles changing through the `style` method although I vaguely remember that I might have been opposed to that.

i.e. in our visibility code we do:

``` ruby
      def hide
        style[:hidden] = true
        update_visibility
      end
```

which is arguably faster (might need/should do a benchmark on that) but also a lot harder to trace (no real entry point when changes occur).

Does anyone remember what we decided?

Thanks + cheers,
Tobi

jasonrclark:
Fixes #1176.

Per that issues, `Shoes::Common::Visibility` was writing to the styles hash itself directly, then triggering a visibility update. This ought to be done through the `style` method, which already handles the updates itself.

Hunted for other similar cases that we'd missed, but didn't turn any problematic candidates up, so 🌮 for that.

This also adds a more direct test for the module like I've been doing in other cases for these modules.

### DIFF
--- a/shoes-core/spec/shoes/common/visibility_spec.rb
+++ b/shoes-core/spec/shoes/common/visibility_spec.rb
@@ -29,7 +29,7 @@ describe Shoes::Common::Visibility do
     expect(subject.gui).to have_received(:update_visibility)
   end
 
-  it "shows" do
+  it "toggles" do
     subject.show
     expect(subject.style[:hidden]).to eq(false)
 
--- a/shoes-core/spec/shoes/common/visibility_spec.rb
+++ b/shoes-core/spec/shoes/common/visibility_spec.rb
@@ -1,0 +1,49 @@
+require 'spec_helper'
+
+describe Shoes::Common::Visibility do
+  class VisibilityTester
+    include Shoes::Common::Visibility
+    include Shoes::Common::Style
+
+    attr_reader :gui
+
+    def initialize(gui)
+      @style = {}
+      @gui = gui
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


shoes-core/spec/shoes/common/visibility_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  it "hides" do
+    subject.hide
+    expect(subject.style[:hidden]).to eq(true)
+    expect(subject.gui).to have_received(:update_visibility)
+  end
+
+  it "shows" do
+    subject.show
+    expect(subject.style[:hidden]).to eq(false)
+    expect(subject.gui).to have_received(:update_visibility)
```
PragTob: I was like "but where do we `update_visibility` now?" - but there's a spec that it happens (great speccing @jasonrclark :heart: ) so I'm confident that we do it, so I'm good :d

shoes-core/spec/shoes/common/visibility_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  it "shows" do
```
PragTob: probably "toggles" ?

PragTob: (show is the one above already)

jasonrclark: Good catch. Copy-paste error.

PR #1378, opened 2017-02-02
Fix setting clicks during initialization

jasonrclark:
While testing around for #1377, I noticed that I didn't seem to be able to set the clicks appropriately during initializing an element. Turned out, if you didn't pass a block to an element, we didn't call `register_click` which was where the style lookup for `:click` happened.

Luckily, if `register_click` exists on a class, it can be called with a `nil` block, so safe enough to call through with a new guard in place.

### DIFF
--- a/shoes-core/lib/shoes/common/initialization.rb
+++ b/shoes-core/lib/shoes/common/initialization.rb
@@ -58,8 +58,7 @@ class Shoes
       #
       # Override if DSL element uses that block for something else (i.e. slot)
       def handle_block(blk)
-        return unless blk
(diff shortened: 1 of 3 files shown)


jasonrclark:
Totally agree these deserve specs and happy to do it. Just ran out of steam last night... will see the what tonight brings!

jasonrclark:
How do those specs strike you @PragTob? Bit of fiddling to get a testing class to use `Shoes::Common::Initialization` (which actually proves it's doing what we think with the real `register_click` invocation), but gets it done.

PR #1379, opened 2017-02-02
Pin version of Rubocop

jasonrclark:
Will look after releases for updates to the gem and do it manually.

@PragTob this addresses the main pain-point that's been breaking builds when it updates. Do you think we should apply direct versions to all the other gems in the Gemfile too?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,12 +20,13 @@ group :development do
   gem "rspec-its", "~>1.2.0"
   gem "rake"
 
+  gem "rubocop", "0.47.1"
+
   gem "yard"
   gem "kramdown"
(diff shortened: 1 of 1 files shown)


PragTob:
Big :+1: on this one!

As for other versions, I'm inclined not to pin them. If we pinned them all - we could just check in Gemfile.lock - I know it's not a ruby thing to do but I chatted with José Valim and the point of having the same development environment for contributors without surprises seems valid. We could still run one build that deletes it to check with newest versions or something.

So, I don't like to pin them although we might give them overall `~> major.minor` pins

PragTob:
Disregarding the discussion this should go in :D

PR #1380, opened 2017-02-04
Provide support for readonly state

### LINKED ISSUE #1187: Implement "readonly" value for state on input elements (edit line, edit box)
manual:

> The :state style is for disabling or locking certain controls, if you don't want them to be edited.
> 
> Here are the possible style settings:
> 
>    nil - the control is active and editable.
>    "readonly" - the control is active but cannot be edited.
>    "disabled" - the control is not active (grayed out) and cannot be edited.

Please check what the SWT equivalent is, if there is one. I guess it'd break down to disallow change events on the widget if there is no explicit setting.

If this turns out to be too hard I'm ok with cutting it.

Also please note that the original shoes manual mentions the state style only for EditBox/Line while we have it for button, list box and check/radio boxes. For those readonly makes limited sense I guess but we can see about that.

jasonrclark:
Fixes #1187

Couple things going in this to get it all lined up:

* Obviously had to implement readonly for input boxes (doesn't make sense elsewhere, and Shoes 3 treats it like any other non-disabled state). This was accomplished via an SWT verify listener on keypresses 👌 
* Noted that setting `state` via style wasn't sticking because we weren't triggering the updates from the style module (like all the others). Fixed that up following the same pattern (although I'm trying to figure out a more elegant way to let the modules declare they should get called for certain style changes, but haven't gotten there yet).
* Removed entirely unused `state_options` method (woot!)
* Moved style merging above all the triggers in `Shoes::Common::Style`. A few spots were explicitly setting things in `@style`, and then the merge at the end sets it again. Looks like things are good with just merging first.

### DIFF
--- a/shoes-core/lib/shoes/input_box.rb
+++ b/shoes-core/lib/shoes/input_box.rb
@@ -33,6 +33,15 @@ class Shoes
     def caret_to(index)
       @gui.caret_to(index)
     end
+
+    def readonly?
+      state.to_s == "readonly"
+    end
+
+    def update_from_state
+      super
+      @gui.readonly = readonly?
+    end
   end
 
   class EditBox < InputBox
--- a/shoes-core/spec/shoes/shared_examples/state.rb
+++ b/shoes-core/spec/shoes/shared_examples/state.rb
@@ -13,6 +13,7 @@ shared_examples_for "object with state" do
   end
 
   it "should disable" do
+    subject.state = ""
     expect(subject.gui).to receive(:enabled).with(false)
     subject.state = "disabled"
     expect(subject.state).to eq("disabled")
--- a/shoes-core/spec/shoes/shared_examples/state.rb
+++ b/shoes-core/spec/shoes/shared_examples/state.rb
@@ -13,7 +13,7 @@ shared_examples_for "object with state" do
   end
 
   it "should disable" do
-    subject.state = ""
+    subject.state = nil
     expect(subject.gui).to receive(:enabled).with(false)
     subject.state = "disabled"
     expect(subject.state).to eq("disabled")
--- a/shoes-swt/lib/shoes/swt/input_box.rb
+++ b/shoes-swt/lib/shoes/swt/input_box.rb
@@ -20,6 +20,10 @@ class Shoes
         @real.add_modify_listener do |event|
           @dsl.call_change_listeners unless nothing_changed?(event)
         end
+
+        @real.add_verify_listener do |event|
+          event.doit = false if @readonly
+        end
       end
 
       def text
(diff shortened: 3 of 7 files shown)


shoes-core/lib/shoes/input_box.rb
```diff
@@ -33,6 +33,15 @@ def highlight_text(start_index, final_index)
     def caret_to(index)
       @gui.caret_to(index)
     end
+
+    def readonly?
+      state.to_s == "readonly"
```
PragTob: Freeze string literals at the top PLS :D

shoes-core/spec/shoes/shared_examples/state.rb
```diff
@@ -13,6 +13,7 @@
   end
 
   it "should disable" do
+    subject.state = ""
```
PragTob: isn't the default value, according to the manual, `nil` ?

shoes-swt/lib/shoes/swt/input_box.rb
```diff
@@ -20,6 +20,10 @@ def initialize(dsl, app, text_options)
         @real.add_modify_listener do |event|
           @dsl.call_change_listeners unless nothing_changed?(event)
         end
+
+        @real.add_verify_listener do |event|
+          event.doit = false if @readonly
```
PragTob: lol? :D There is an accessor in java called `doit` ? Surprises never cease, also nifty implementation!

shoes-swt/lib/shoes/swt/input_box.rb
```diff
@@ -35,6 +39,10 @@ def enabled(value)
         @real.enabled = value
       end
 
+      def readonly=(value)
+        @readonly = value
+      end
```
PragTob: couldn't we just make this an attr_writer?

shoes-core/lib/shoes/input_box.rb
jasonrclark: Ha yes. Will get to that and do the issue for applying it everywhere soon.

shoes-core/spec/shoes/shared_examples/state.rb
jasonrclark: Might be, but I just needed it to have a value change from initialization (which is setting "disabled") since style skips no-op sets.

shoes-swt/lib/shoes/swt/input_box.rb
jasonrclark: Good catch. Originally had it as just readonly (no =) like the enabled, so missed that a writer would be better. 👍

jasonrclark: Good catch. Originally had it as just readonly (no =) like the enabled, so missed that a writer would be better. 👍

shoes-core/lib/shoes/input_box.rb
jasonrclark: Doing the whole batch in #1381

PR #1382, opened 2017-02-06
Turn UIElement, ArtElement, BackgroundElement to classes

jasonrclark:
Spoken of earlier, but a few other things I've been ruminating on would benefit from having the UIElement and others as full classes.

In particular this should yield some fruit around how we do updates after style setting, which currently is 1) ugly 2) has rough edges if you set keys on an unsupporting element and 3) probably inefficient (benchmarks will prove whether this bit is true).

Got a plan for handling the `Hover` weirdness we still have, but can probably apply those in separate PRs.

### DIFF
--- a/shoes-core/lib/shoes/arc.rb
+++ b/shoes-core/lib/shoes/arc.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 class Shoes
-  class Arc
-    include Common::ArtElement
-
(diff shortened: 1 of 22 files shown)


jasonrclark:
Pardon the unclear comments... benchmarking is for a later thing I want to try _based_ off this. I don't think that module vs class should matter much since it's all startup definition time and a pretty small set of stuff anyway.

PR #1387, opened 2017-02-10
Stop doing Initalization as module

jasonrclark:
Now that we have a base class, just use it more directly.

Less magic 🌈  more better 👌


jasonrclark:
Probably overstated calling it  "magic". For me it's a little exotic though for a class to get its `initialize` method from a module (although the mechanism is just good old module included methods) and definitely unneeded indirection now.

PragTob:
That is definitely true :)

PR #1389, opened 2017-02-12
Provide more direct warning on jar startup

### LINKED ISSUE #1344: Better error message if packaged jar started on Mac without -XstartOnFirstThread
Noticed that when we generate a jar, you get a rough failure case if you don't use the magic `-XstartOnFirstThread` setting (unlike some other cases like `require 'shoes'` where we manage to tell you more)

We could do better about giving a meaningful message in this case by putting some checking into the packaged app's bootstrapping code.

jasonrclark:
On Mac if you run `java -jar my_app.jar` without `-XstartOnFirstThread`,
you'll get the usual thread access errors that we protect you from in
most other cases. This at least gives you a warning about what happened
directly rather than just dying with a cryptic error.

Fixes #1344

### DIFF
--- a/shoes-package/lib/warbler/traits/shoes.rb
+++ b/shoes-package/lib/warbler/traits/shoes.rb
@@ -29,13 +29,35 @@ module Warbler
         [Traits::Jar, Traits::NoGemspec].include? trait
       end
 
+      INIT_CONTENTS = <<~EOS
+        require 'shoes'
+        require 'shoes/swt'
+
+        Shoes::Swt.initialize_backend
+      EOS
+
       def after_configure
-        config.init_contents << StringIO.new("require 'shoes'\nrequire 'shoes/swt'\nShoes::Swt.initialize_backend\n")
+        config.init_contents << StringIO.new(INIT_CONTENTS)
       end
 
       def update_archive(jar)
         super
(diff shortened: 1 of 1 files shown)


shoes-package/lib/warbler/traits/shoes.rb
```diff
@@ -29,15 +29,37 @@ def self.requires?(trait)
...
+      EOS
+
       def after_configure
-        config.init_contents << StringIO.new("require 'shoes'\nrequire 'shoes/swt'\nShoes::Swt.initialize_backend\n")
+        config.init_contents << StringIO.new(INIT_CONTENTS)
       end
 
       def update_archive(jar)
         super
         add_main_rb(jar, apply_pathmaps(config, default_executable, :application))
```
PragTob: I... don't understand. We've been calling this method all along and just now we implement it? I'm... I don't know. Magic? Inheritance? What?

jasonrclark: It comes from the [base `Warbler::Traits` module](https://github.com/jruby/warbler/blob/master/lib/warbler/traits.rb#L82-L87)

PragTob: got ya, thanks!

Issue #1200, opened 2015-12-09
shoes3 to shoes4 items require changing


thebenjy:
shoes3 app all works fine. using shoes4 had to change the following

**didn't recognised**

```
   self.start do
        # do scrreen refresh
   end
```

**crashed the app if button pressed**

```
  $log_window has be initiated as a para ''
  button "Copy Log Content" do
     app.clipboard = $log_window.text unless $log_window.nil?
  end
```

**doing a widget update (list_box) within the change event**
in shoes3 the box always selected an item, in shoes4 it was blank - probably not good code anyhow so needed to change

```
  $ini_file_lb.change do
    # do something that included
    $ini_file_lb.items = <new list>
  end
```

**layout changes for stacks**
fonts, gaps etc were all smaller under shoes4
had a number of stacks which had a margin, drawing the objects
within the stack ended up displaying outside the stack 

**basic app**
under shoes 3 all widgets contained within the white rectangle of the stack
shoes4 the buttons displayed on the border/outside the bottom of the rectangle

```
   stack :width => "98%", margin: 10 do
      background white
      flow do
        image "1.png", :margin => 0
        title "Environment"
        flow do
             $env_r_staging = radio :server, checked: true
             para "Staging (Test)       "
             $env_r_production = radio :server
             para "Production"
           end
             para "" # added in shoes 4 to make sure stack white rectangle big enough
      end #flow
    end # stack
```

PragTob:
Thank you, we'll investigate those individually. I've never seen `self.start` - you never learn enough about the old shoes :D 

We can try to get text sizes and margins closer to shoes3, which we should, but they'll never be the same.

ccoupe:
In Shoes3 self.start{} probably gets to app.start{}. It's a block to run after the all the widgets in partial ly evaluated Shoes.app are on screen and able to accept state settings. Some folks use a timer(0) block to do the final gui setup. It's something of a standard practice when setting the check or radio states after the Shoes.app block is loaded/evaluated but the event loop isn't really running so there's not widget to change. Very much a Shoes huh! thing that bites everyone.

thebenjy:
i actually prefer smaller text sizes but prob acknowledge a clean migration is required. maybe a switch to "update look n feel".
@ccoupe timer(0) seems fair approach. perhaps an alias to handle migrations?

jasonrclark:
[...]

PR #1390, opened 2017-02-12
Support start block method in app

jasonrclark:
Runs after main block processing has concluded for app.

For example this app:

```
Shoes.app do
  para "before?"

  start do
    para "in the start!"
  end

  para "after!?"
end
```

Looks like this (not `"in the start!"` is after even the `para` that follows the `start` call):

image

First of the things in #1200.

### DIFF
--- a/shoes-core/lib/shoes/app.rb
+++ b/shoes-core/lib/shoes/app.rb
@@ -62,6 +62,11 @@ class Shoes
       self.class.new(options, &block)
     end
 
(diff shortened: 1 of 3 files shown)


PragTob:
Thanks a lot! =D

great test to think of that edge case! 

sorry for the tardiness, been sick :(

jasonrclark:
Thanks for the merge! Feel better and no rush!

PR #1391, opened 2017-02-12
refactor: Move DSL Text methods to DSL::Text

OrthoDex:
Shifted the text methods to another file within a new directory 'dsl'.

How does this look @PragTob @jasonrclark ?

### DIFF
--- a/shoes-core/lib/shoes/dsl.rb
+++ b/shoes-core/lib/shoes/dsl.rb
@@ -76,6 +76,8 @@ require 'shoes/text_block_dimensions'
 
 require 'shoes/color'
 
+require 'shoes/dsl/text'
+
 require 'shoes/common/attachable'
 require 'shoes/common/changeable'
 require 'shoes/common/clickable'
@@ -151,7 +153,7 @@ class Shoes
   #                      a hash of default styles for elements of Class,
   module DSL
     include Color::DSLHelpers
-
+    include DSL::Text
     # Set default style for elements of a particular class, or for all
     # elements, or return the current defaults for all elements
     #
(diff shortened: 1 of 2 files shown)


jasonrclark:
Hey @OrthoDex, the code here is looking good to me.

There is a failure in the build from Rubocop which wants a frozen string literal comment at the top and some slightly different spacing. You can run locally to see these failures by executing `rubocop`, but I've logged #1392 so the default Rake task also runs it so folks will have a better chance of catching these type of things before CI.

Once that's tidied up, this should be 👌 for merging.

OrthoDex:
Hey @jasonrclark ,

rake spec:core is running successfully on my local machine,I dunno why there is a NullPointerException. :thinking:

jasonrclark:
@OrthoDex That's totally an odd failure... looks like an underlying JRuby thing? I've bumped it to rebuild and if we're all good, will merge this evening when I get down to my shoes time. Thank you!

jasonrclark:
Looking good, thanks again @OrthoDex!

For future PRs of this type you can either do them one by one or group things together if you'd rather. I would appreciate the individual commits per move to make it easier to review in batch.

✨

OrthoDex:
Thank you so much @jasonrclark ! :smile: 

I was thinking of doing small commits, squashing them in one commit per extraction and then issuing a PR for the entire thing.

If you would like to merge the extractions as they come, I can issue individual PRs for each extraction. I think we can then track the code climate status as the extractions come.
Thoughts?

jasonrclark:
One PR with each extraction as a commit sounds 👌 for me.

If at any point you need to pause, don't feel like you need to hold the rest until it's all there, just push a PR.

Feels like a PR per extraction is just busy-work, but I deeply appreciate keeping the individual commits for diffing.

OrthoDex:
Got it!
I'll issue a PR soon, and add commits as I start extracting the methods. :smile:

PR #1392, opened 2017-02-13
Add rubocop to rake default set

jasonrclark:
Enough folks (including me) trip over this on first pushes, worth putting in the default testing set.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: ruby
 script:
-  - bundle exec rubocop -D
-  - bundle exec rake spec:all
+  - bundle exec rake
 rvm:
   - jruby-head
   - jruby-9.1.7.0
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ Shoes 4 is developed in a TDD style using RSpec. You should be writing and runni
 
 The simplest way to do this is with rake tasks.
 
-    $ rake spec
+    $ bundle exec rake
 
 (Try `rake --tasks` for a full list)
 
--- a/tasks/rubocop.rb
+++ b/tasks/rubocop.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 require 'rubocop/rake_task'
 
-RuboCop::RakeTask.new
+RuboCop::RakeTask.new do |task|
+  task.options << "--display-cop-names"
+end
 
 task default: :rubocop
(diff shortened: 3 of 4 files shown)


PragTob:
if we do this, shouldn't we maybe make the `rake` do whatever our whole travis build script does? E.g. also spec:all etc?

otherwise, good ;)

jasonrclark:
So default rake actually ends up running `spec` which runs `spec:all`... so maybe I should just change Travis to run `rake` as part of this too. What do you think?

jasonrclark:
On further consideration, made a few more tweaks:

* Added option to keep the cop name reporting in Rubocop, which I find useful
* Did indeed update `.travis.yml` to just run `bundle exec rake`. Keeps us honest on what our defaults are vs what CI runs
* Recommended for testing that people just run `rake` instead of `rake spec` so they'll get the full effect of what they're asked to pass in CI

README.md
```diff
@@ -200,7 +200,7 @@ Shoes 4 is developed in a TDD style using RSpec. You should be writing and runni
 
 The simplest way to do this is with rake tasks.
 
-    $ rake spec
+    $ rake
```
PragTob: bundle exec! (otherwise packaging specs fail for me)

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: ruby
 script:
-  - bundle exec rubocop -D
-  - bundle exec rake spec:all
+  - bundle exec rake
```
PragTob: I think travis runs `rake` as default but probably without `bundle exec` or maybe, not sure - also it's better to be explicit. So this is good. I don't know what I'm writing ;P

tasks/rubocop.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+require 'rubocop/rake_task'
+
+RuboCop::RakeTask.new do |task|
+  task.options << "--display-cop-names"
+end
+
+task default: :rubocop
```
PragTob: :+1:

README.md
jasonrclark: It still fails? I thought I'd fixed that.

Either way, I'll update this as my rule is ABB... "Always Be Bundling"

PragTob: I'm not sure, maybe you fixed it and I just haven't knowingly tried afterwards ever again :)

PR #1393, opened 2017-02-14
Get clipboard working again

jasonrclark:
Another day, another compatibility regression over in #1200!

This time, it appears from my testing that this was only failing on the Mac. When I tried the 3 parameter version instead of 2 parameter, it happily works now. Tested on VMs for Windows and Ubuntu, and those went swimmingly.

This is a bit in the weeds testing-wise. Could maybe try to wire something in new that actually round-trips text through the clipboard, but what's there right now is just mocking out that we get the call to the `clipboard` methods. If we do want tests, that'll be a task for another night, as it's time for 💤

### DIFF
--- a/samples/simple_clipboard.rb
+++ b/samples/simple_clipboard.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+Shoes.app do
+  stack do
+    @line = edit_line "Put text here"
+    flow do
+      button "Copy" do
+        self.clipboard = @line.text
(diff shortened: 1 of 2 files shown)


PragTob:
I think this will be super hard to test... and we don't run on Mac either way in CI.

Do we have a good sample to notice this earlier the next time? :)

jasonrclark:
Oh, great call. I'll make a clippy sample app before we merge this. :+1:

PR #1394, opened 2017-02-15
Restore selection after list changes

jasonrclark:
More compatibility goodness from #1200 

In this case, we found that anything which changed out the items list in a listbox would drop the current selection. To address this, we capture the value immediately before (easiest to just get the `@real.text` selection) and then put it back after the items change.


PragTob:
Thanks a lot! :D

Issue #1341, opened 2016-12-21
Split up `shoes-core/lib/shoes/dsl.rb`


jasonrclark:
Per https://github.com/shoes/shoes4/pull/1336#discussion_r93190641, this file has gotten pretty long. A sensible way of breaking it up would be 💯 .

OrthoDex:
Hi!

I would like to take this up if it's alright. :smile: 

Code Climate shows the duplication and complexity, and I understand that the DSL could be sliced up into Art, TextBlock etc according to @PragTob

PragTob:
Hi there @OrthoDex !

Thanks for wanting to help! Yes it should, although I'm not entirely sure on all the categories yet. Art/TextBlock seem sensible but there is other stuff, e.g. Timer related things.

Before you get all started in code I think it might be a good thing to have a look at the methods (I don't have that much tome atm) and see what these and the other categories could be so that we could talk about them at least a bit before we mold them into code.

What do you think? :)

OrthoDex:
On it @PragTob ! :smiley: 

I'll try and propose a splitup into specific modules after going through the methods.

OrthoDex:
Hi @PragTob ,

I went through the code and I could roughly identify a few potential groupings. These are based on the broad functionality they provide. The code comments helped out a **lot**, since I could identify the parts providing diff functionalities.

Here it goes:

- DSL::Setup (including DeprecatedShoesGemSetup, #load_backend, #setup )
- DSL::Style (housing all private methods for style_* apart from #create)
- DSL::Element (housing all major elements along with #create
- DSL::Animate (Not sure about this, since organizing #timer and #sound together doesn't sound like good practice, no pun intended.)
- DSL::Art or DSL::Shape (to include Shoes::Arrow, Shoes::Arc till Shoes::Star)
- DSL::AppStyle (for TextStyle, PATTERN_APP_STYLES and OTHER_APP_STYLES)
- DSL::Action (for mouse and key related actions like #hover, #scroll_top, #download etc)

What do you think?

PragTob:
Hey there!

Looks roughly good to me, couple of comments.

* load_backend/setup are class methods, e.g. not like the rest of the flock, they could still move to a module but would probably keep them out of the DSL namespace
* DSL::Style/style_* methods is indeed more private and not meant for "normal" users to call - I think this will end up as a module that is included in some other DSL modules. However, we might even find out that some methods are just called from some modules (if slices right) so they could just be put as private in there
[...]

PR #1395, opened 2017-02-21
Shoes DSL Extraction

OrthoDex:
Initial PR for #1341 

Created the Interaction module since it's more general.


jasonrclark:
According to [this comment](https://github.com/shoes/shoes4/blob/master/shoes-swt/lib/shoes/swt/app.rb#L153-L155), `gutter` is related to scrollbar stuff, so I'd say it fits neatly here.

Also agreed on `visit`... it's a little odd, but I can't think of a better home offhand.

If you're good with it @PragTob, I'm :ok_man: here.

PR #1397, opened 2017-02-25
refactor: Extract app and media elements

OrthoDex:
Part of #1341


PragTob:
LGTM.

Only concern is that I think image should probably live with Media if that becomes its own thing? :)

PragTob:
Thank you!

PR #1400, opened 2017-03-01
Pick up gems from Bundler

### LINKED ISSUE #1245: Packaged App
Hi Chaps,

Awesome work on Shoes4! I have a scrip that relies on one external gem, which runs well from the cmd line using shoes4. I can package the app into a .app but when I try to launch it it does nothing...

Any ideas?

Rich

jasonrclark:
Fixes #1245 

If we detect the presence of Bundler when packaging, we'll try to use
its list of gems (minus `shoes` which causes us problems and isn't
necessary to install). This allows Rubyists to use the typical means of
declaring their dependencies in a Ruby app, but works nicely even
without a Gemfile around too.

So what's up with that conditional test? Ewww, right? Well, it got even
worse when I tried to stub in Bundler's business if it was missing...
three times the code, super-tight binding. I think I'm willing to just let
the non-Bundler case go here but I'm open to discussion.

### DIFF
--- a/shoes-core/lib/shoes/packager.rb
+++ b/shoes-core/lib/shoes/packager.rb
@@ -6,6 +6,7 @@ class Shoes
     def initialize
       begin
         @backend = Shoes.backend_for(self)
+        configure_gems
       rescue ArgumentError
         # Packaging unsupported by this backend
       end
@@ -23,6 +24,17 @@ class Shoes
     def run(path)
       raise "Packaging unsupported by this backend" if @backend.nil?
       @backend.run(path)
+    end
+
+    def configure_gems
+      @backend.gems = []
+      return unless defined?(::Bundler)
+
+      @backend.gems = ::Bundler.environment.specs.map(&:name) - ["shoes"]
+    rescue Bundler::GemfileNotFound
+      # Ok to be quiet since we didn't even have a Gemfile
+    rescue => e
+      puts "Looking up gems for packaging failed:\n#{e.message}"
     end
 
     def help(program_name)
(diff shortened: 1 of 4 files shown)


shoes-core/lib/shoes/packager.rb
```diff
@@ -25,6 +26,17 @@ def run(path)
       @backend.run(path)
     end
 
+    def configure_gems
+      @backend.gems = []
+      return unless defined?(::Bundler)
+
+      @backend.gems = ::Bundler.environment.specs.map(&:name) - ["shoes"]
```
PragTob: might we need all our gems here or just shoes? Not sure what the direct problem is

PragTob: Okay I see from thes pecs that it should be there, nevermind

PragTob:
Also yeah the conditional test... seems fine to me :) We could tag it like we tag some tests not to run on OSX (or at least I believe we did) but seems good & pragmatic to me (Maybe I'm also just used to it now as simplecov tests are basically wrapped in `SimpleCov.enabled?` ) :)

shoes-core/lib/shoes/packager.rb
jasonrclark: FWIW, the problems with including `shoes` directly is that `warbler` still doesn't play nicely with gems that have native extensions (which ours does for the executable install).

jasonrclark:
So after reviewing, I pushed two more tweaks here. One was to make the error around not finding gems more realistic via Bundler's exception for missing a Gemfile. Second, I added a warning message when we're skipping the tests because Bundler isn't there.

PragTob:
:+1:

PR #1402, opened 2017-03-02
Package cleanup

jasonrclark:
In rooting around in packaging, found a number of things to clean up:

* Weren't using `Shoes::Package::Jar` and `JarApp`. Removed them as I don't see much value in the existing wrappers right now.
* Pulled some stuff from `shoes-swt` that interacted more directly with `furoshiki` into `shoes-package`. These weren't a problem per-se in `shoes-swt`, but I prefer to keep the line strong of "all packaging only stuff (except invoking it) goes into `shoes-package`.
* Removed some complexity we had for "supporting multiple backends" but buried in the SWT layer. If we need that back, we'll add back at a different place when we get there.
* Fixed test ordering issue around requires for `furoshiki`


jasonrclark:
Rebased and put back the bundler thing which I'd neatly rebased away 😆 

I hear you that there probably were some good ideas there @PragTob, but I'm confident that if they're needed something similar will surface, and right now less packaging code for me to wrap my brain around is 👌

Issue #1321, opened 2016-12-08
Fix up the Shoes console


jasonrclark:
Realized another gap with Shoes 3, and that's the old console. It could be brought up with a simple `Alt+/`, and provided visibility to the errors and output that happened in your Shoes app.

![image](https://cloud.githubusercontent.com/assets/130504/21001024/e6cd06d2-bcd2-11e6-8a3e-3111a1b1ce62.png)

Personally spitting problems to STDOUT matches my expectation, but these aren't necessarily exclusive. We could easily spot errors in block evaluations, and write to both locations.

If we intend to have a "Shoes" app like old Shoes shipped (maybe related to #125?), where people might be starting from OS app launchers instead of a terminal, the console might be a necessity to see when things go side-ways. Or perhaps there's a better/different paradigm we could do to help, especially beginners, when their apps hit an error.

What do y'all think❓❔❓

## TODO

* [x] Fix so logging works
* [x] Ensure (or fix?) so exceptions get written. Don't break writing to STDOUT too!
* [x] Better notification scheme for updates than polling

ccoupe:
I use both in Shoes 3. The console, log.rb, alt-/ to find the Ruby/Shoes syntax problems and also debugging puts in the code - none of those ruby puts will be seen if Shoes is launched as an app where stdout, stderr, and stdin are effectively /dev/null. The next Shoes 3 (3.3.2) has a Shoes.terminal that can be called if the end user really wants it and it redirects GUI launched stdin/out/err to it. Pain in the native ass. Lot's of C and Objective-C. But you can run things like mini-tests  and some other low level thing creatures from Shoes 3.3.2 and get color highlighting.

The debug,info,warn methods are part of Shoes "classic" and they go to log.rb (pure ruby, last I looked)

PragTob:
But... we have the shoes console already or what am I missing? Ever since 2013 afaik. Maybe the keyboard short cut is broken on mac but on Linux it works fine. It's [here](https://github.com/shoes/shoes4/blob/master/shoes-core/lib/shoes/logger.rb) - maybe some of the recent rubocop fixes broke it? Those are the last commits.

Meaning, it opens... now when I tried it out it seems to be broken, e.g. the log does not update when I call methods. I'll PR a simple sample. 

[...]

PR #1406, opened 2017-03-10
New Shoes::Logger and Shoes::Console

jasonrclark:
First major swipe toward getting #1321 better.

Quite a few changes, but they're all entangled together so easiest to tee this up as a batch. Further revision to the console can happen separately later on.

* Major revision to logging.
  * `Shoes.logger` is the main thing, no longer run through the `Shoes.configuration` object.
  * Axed the registration business as it wasn't being used
  * `Shoes::Logger` is now a simple composite that passes on to a `STDERR` logger and the Shoes console! (and could be wired into by others if we wanted)
* Revised Shoes console implementation!
  * Still looks the same, but receives updates directly through logging (facilitated by the new `Shoes::Logger` structure). Hence, no timer-based updates which is ✨ 
  * Separate `Shoes::Console` class from logging, for better findability
* Replaced a few `puts` statements with logging instead.
  * I didn't do this for all instances, as especially some of the bad exception swallowers in SWT event loops would crash the console holding too much junk because those type of error tend to churn out toooooons of stuff.
* Added method to detect if an app window is still `open?`, and to force `focus` to it (so we don't open multiple console windows on the key-press like we were before)


shoes-core/lib/shoes/builtin_methods.rb
```diff
@@ -4,7 +4,7 @@
 # However they can also be used from the normal Shoes.app block.
 class Shoes
   def self.p(message)
-    Shoes::LOG << ['debug', message.inspect]
+    Shoes.logger.debug(message.inspect)
```
PragTob: :+1: :rocket:

shoes-core/lib/shoes/logger.rb
```diff
@@ -1,66 +1,47 @@
...
+      @loggers = []
+    end
 
-      def unregister(name)
-        @loggers.delete(name)
-      end
+    def <<(logger)
+      @loggers << logger
+      self
+    end
```
PragTob: this interface tripped me up a bit. It looks a bit weird to do `logger_instance << a_logger` imo as well... it might as well add a log message, you know?

Not saying it needs changing, just a consideration that something like `add_logger` might be good...

But then again, why are loggers added to a logger? Seems weird. Maybe one is the logger and the others are logging_backends? 

Maybe, one is the `LogInterface`/`LogDistributor` and the others are the actual loggers? I hope my naming confusion gets across :D 

Big :+1: for the overall design though, e.g. having one logger thingie with multiple backends :rocket:

shoes-core/lib/shoes/logger.rb
```diff
@@ -1,66 +1,47 @@
...
-          update
+    def warn(message)
+      forward(:warn, message)
+    end
+
+    def error(message)
+      forward(:error, message)
+    end
+
+    def forward(meth, message)
```
PragTob: shouldn't we be private from here on out?

jasonrclark: Totally could be. I'm pretty lax about marking things private, but this'd be the spot in this class to start.

shoes-core/lib/shoes/logger.rb
jasonrclark: Agreed that the naming didn't end up quite as clear as I'd have liked on first pass, and I totally see what you mean. I see two possible renames that might address it:

* My fancy OO patterns would have me call `Shoes::Logger` a `CompositeLogger` which makes it clearer that it's purpose is to dispatch to other loggers.
* Alternatively, we could also just treat `Shoes::Logger` as a `LoggerCollection` instead. `<<` is more natural in that context, although I'm not opposed to just renaming to `add_logger` if you find that clearer.

What do you think? Either of those work for you @PragTob?

PragTob: Both work fine for me, CompositeLogger sounds fancier but I'll gladly leave the decision in your capable mind and hands :) :star2:

jasonrclark: Decided that `LoggerCollection` was a little more straight-forward and plain, so ran with that. Thanks for the nudge on it @PragTob... happier with where it landed.

PR #1409, opened 2017-03-13
Fail fast option

jasonrclark:
As discussed over in https://github.com/shoes/shoes4/issues/1321#issuecomment-284620535, I'm aiming to keep Shoes 3's error swallowing (for now), but allow those of us who like early failure to embrace that practice as well.

This PR adds in the configuration and the safe evaluation method necessary to support this work in a tidy fashion.

Here's the list of other spots to apply the wrappers... might do a couple in this PR to test out how the module is working. Will see what balance feels right to not get this one too bulked up.

* [x] `Slot`
* [x] App resize callback
* [x] `Common::Hover`
* [ ] `Common::Changeable`
* [ ] App mouse motion
* [ ] App hover/leave
* [ ] App keypresses
* [ ] Download callbacks
* [ ] `Link` click
* [ ] `Shape`
* [ ] `App#append`
* [ ] ???


jasonrclark:
All right @PragTob, I've applied this in three different cases: the main eval, a separate app-level (resize) and a single element method (hover). I'm pretty happy with how this is shaping up. If you're cool with it, I'll head this to a merge, get pre9 out, and handle the rest of the list in separate PRs/batches afterward.

PR #1410, opened 2017-03-13
Clean up some Cop values, exclusions

jasonrclark:
Thanks to @OrthoDex's fine work on #1341, I was able to remove a couple of direct references to excluding our `dsl.rb` file from Rubocop rules. ✨ 

Along the way cranked down a couple max values closer to where they are, and removed an exclusion that was around for a Rubocop bug.


PragTob:
:heart:

PR #1412, opened 2017-03-18
Change images blocks to clicks

jasonrclark:
Back in https://github.com/shoes/shoes4/pull/1388 we made passing a block to images show a deprecation message as it didn't do what it used to in Shoes 3 (and we aren't supporting it just now).

Shoes 4 had already picked up the behavior of supporting default click action, though, if you passed one of those blocks and a few samples were using it. D'oh!

I've changed those samples to simply pass the `click:` explicitly.


jasonrclark:
Sigh, rubocop. Will get to fixing that later today but gotta run out now.

PR #1414, opened 2017-03-24
New gems, new rubies

jasonrclark:
With pre9 out the door (:tada:) it was time to look at what needs version bumps. Rubocop, our perinneal favorite for introducing spec breaks when we didn't pin it... didn't release, so it's all good.

RSpec had a minor release, and everything looks 🍏  on it.


PragTob:
added latest ruby version to travis :)

jasonrclark:
Great catch @PragTob!

Issue #1320, opened 2016-12-08
Where're my samples?


jasonrclark:
In testing for `pre7` (:tada:) I did a gem install on an Ubuntu VM and promptly realized I don't have samples since I haven't cloned from Github!

Maybe we should package up the samples in `shoes` or `shoes-core` and provide a command to install them? Maybe `shoes --samples path/to/my/samples`?

PragTob:
I like the idea. First thought was only working samples should be packaged, but by release all of them will/should work so no need to overengineer that now.

API wise I'd rather it not be a switch but a "sub command" like `copy-samples`, `samples` or `samples copy` - which makes me realize that packaging is also just a switch. I like proper sub commands better but maybe I spent too much time in elixir/mix :)

jasonrclark:
I'm onboard with the subcommands when we get there. This also aligns with `gem` and `bundle`, and won't feel at all weird to Ruby folks. Will entail a change for packaging, but I'm totally good with that.

PR #1416, opened 2017-03-27
Move to subcommands vs flags

jasonrclark:
As discussed briefly over in https://github.com/shoes/shoes4/issues/1320#issuecomment-265961168, this PR moves us toward using subcommands instead of flags for everything.

It will move people's 🧀 but I'm cool with it for the nice doors it opens. The main revision will be for packaging. I'm proposing that we change the layout to something like this:

```
shoes package --jar --mac the_app.rb
```

Today there's some binding being done in the DSL/core layer around the backend naming on the options. In reality, we're working with a single backend at a time here today (and it'll be big revisions to change that down the line) so I'd rather drop that and let the backend feed these options up to the CLI system. With the new structure, that'll be totally easy to accomplish.

FWIW although there are new `shoes help` and `shoes version` commands, I kept the dashed flag versions of the commands too... principle of least surprise you know 😉 

## Todo

* [x] Rewire help (planning to do just one help dump for all the commands)
* [x] More testing (manual and specs), particularly around passing too many, too few, etc.
* [ ] ~~Revise packaging options (maybe move to separate PR?)~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,13 +148,13 @@ That's Warbler talking. Actually, we sneak the Shoes gem in anyway, but don't te
 
 Okay, now for real. The simplest thing is to put your script in a directory by itself and then:
 
-    $ bin/shoes -p swt:app path/to/directory-of/your-shoes-app.rb
+    $ bin/shoes package -p swt:app path/to/directory-of/your-shoes-app.rb
 
-This will produce a Mac app, which you can then find at `path/to/directory-of/pkg/Shoes App.app`.
+This will produce a Mac app, which you can then find at `path/to/directory-of/pkg/your-shoes-app.app`.
 
 You can also package a shoes app up as a jar through:
 
-    $ bin/shoes -p swt:jar path/to/directory-of/your-shoes-app.rb
+    $ bin/shoes package -p swt:jar path/to/directory-of/your-shoes-app.rb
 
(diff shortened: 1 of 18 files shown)


jasonrclark:
New help command output (will change slightly for packaging when I revise the options... maybe I'll do that as a separate PR, though, to keep this a little closer to slim):

```
Shoes is the best little GUI toolkit for Ruby.

shoes [options] file
    -v, --version                    Shoes version
    -h, --help                       Shoes help
    -f, --fail-fast                  Crash on exceptions in Shoes code

shoes help
    Displays this help text

shoes manual
    Run the interactive Shoes manual

shoes package [options] file
    -p, --package PACKAGE_TYPE       Package as BACKEND:PACKAGE

    Packages may be built either from a single .rb file, or a .yaml file with
    more options defined.

shoes select_backend [backend]
    Select a Shoes backend to use. A backend can be specified, or Shoes will
    attempt to auto-detect available backends to select from.

shoes version
    Prints the current Shoes version
```

jasonrclark:
Dropping the WIP tag. Any time you're ready for a review @PragTob would be much appreciated. No rush, I've got other things I can work away at in the meantime.

jasonrclark:
What, having the README reflect actual usage? That's not how _I_ do open source 😆 

Kidding of course. Don't have gumption tonight, but will get that tidied up and merge away once I do.

PR #1419, opened 2017-04-17
Mention explicitly that JDK9 is no good atm

PragTob:


jasonrclark:
Good to clarify. Thanks @PragTob!

Issue #1422, opened 2017-05-05
After Package; Shouldn't path/"pwd" be in resources?  path/"pwd" is currently home directory. 


MichaelDimmitt:
When I make a package "pwd"  is the home directory.
"pwd" is not Resources or app's Current location.

code: simple test.rb
<pre>Shoes.app do
    @s = stack {}
      button "go!" do
        @out = `PWD`
        @s.clear { para @out }
      end
end</pre>

![screen shot 2017-05-05 at 3 43 58 pm](https://cloud.githubusercontent.com/assets/11463275/25761740/e540f3d6-31a9-11e7-821e-ecfe450d55ba.png)
<br><br><br> Example when I click the "go" button:<br>
<b>When I do not make a package it tells where I ran the script from, which is fine:</b>
![screen shot 2017-05-06 at 11 17 32 am](https://cloud.githubusercontent.com/assets/11463275/25773593/c1abffc0-324d-11e7-90cc-cea1f34ee2e7.png)

<b>When I make the package it does not tell where I ran the .app from.</b>
![screen shot 2017-05-05 at 3 43 47 pm](https://cloud.githubusercontent.com/assets/11463275/25761737/e3c8a5da-31a9-11e7-9f98-5fc98e292ea8.png)

The program also does not move images to resources folder.
However, I should be able to manually move them to resources folder.
I just need the path to be at Resources inside the Test.app or the path to Test.app.
Having the app be a self contained entity will allow me to run it from other people's computers.

I am aware there are open issues on Packaging.
I did not see an issue solely committed to path location.
If I find the solution to my problem I will comment below.

Thanks for your time reading this issue.
Let me know if you need any further details.

MichaelDimmitt:
some notes so that you know what environment is causing this problem:
note: using mac environment

note: in order to get program working I had to add  gem 'furoshiki', github: 'shoes/furoshiki'
note: issue #291 helped me figure out to add gem 'furoshiki'
note: made pull request #1423

jasonrclark:
Hi @MichaelDimmitt! So I'm working on packaging at the moment, so thanks for the report. I haven't had a chance to try and replicate what you're describing in the main issue yet, but I've added it to the backlog I'm trying to tackle in the next month and will chime back with what I find. Will chat more about `furoshiki` over on the PR you made.

jasonrclark:
You're right about the working directory not getting set. Before we push on that, though, it sounds like your concern there is locating images. Is that right?

[...]

PR #1427, opened 2017-05-07
Fix image loading for packaged apps

### LINKED ISSUE #676: Packaged app fails to open with java.io.FileNotFoundException (accessing images)
Hi all,

I've successfully packaged my app on MacOS X 10.8.5, but the app fails to open, with the following exception:

```
LoadError: load error: META-INF/main -- org.eclipse.swt.SWTException: i/o error (java.io.FileNotFoundException: file:/Users/braincomb/my-app/pkg/myapp.app/Contents/Java/aw.jar!/shoes-app/images/header.png (No such file or directory))
  require at org/jruby/RubyKernel.java:1065
  require at file:/var/folders/md/c7trzt5n66s3p0mwlnkbz6n4jnhzb3/T/jruby21387417529286949extract/jruby-stdlib-complete-1.7.12.jar!/META-INF/jruby.home/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:126
  require at file:/var/folders/md/c7trzt5n66s3p0mwlnkbz6n4jnhzb3/T/jruby21387417529286949extract/jruby-stdlib-complete-1.7.12.jar!/META-INF/jruby.home/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:124
   (root) at <script>:3
error: org.jruby.embed.EvalFailedException: (LoadError) load error: META-INF/main -- org.eclipse.swt.SWTException: i/o error (java.io.FileNotFoundException: file:/Users/braincomb/myapp/pkg/myapp.app/Contents/Java/aw.jar!/shoes-app/images/header.png (No such file or directory))
```

I have this code that calls the images, which works fine in un-packaged state:

```
  @app_dir = File.expand_path(File.dirname(__FILE__))
  @image_dir = @app_dir + "/images/"
  background @image_dir + "light_grey.png"
  image @image_dir + "header.png", top: 10, left: 20
```

I also looked inside the jar file with `jar tf` and made sure that the images were there:

```
shoes-app/images/header.png
shoes-app/images/light_grey.png
```

I removed the call to the image, repackaged it and it's working fine. Also interesting to note that the background didn't work/show, even though there were no errors about it.

Any ideas how to fix this? 

I followed all packaging instructions from the shoes4 repo, which I cloned yesterday. 

ruby -v

```
jruby 1.7.12 (1.9.3p392) 2014-04-15 643e292 on Java HotSpot(TM) 64-Bit Server VM 1.6.0_65-b14-462-11M4609 [darwin-x86_64]
```

Thanks!

jasonrclark:
SWT doesn't understand the embedded file paths that warbler creates
(uri:classloader), and so even when you give a properly calculated
relative path to an image in your JAR, it'd fail noisily.

To address this, we ~~load the image from the file~~ copy those files in
Ruby-land (which knows how to interpret those file paths) and always
just create images directly ~~with the byte arrays.~~ from those temp files.

What's remaining? TESTING! Did a bit but not satisfied yet, and don't
have the gumption to finish tonight so thought I'd get the PR rolling
anyway.

Fixes #676. Potentially also does the same for #1422 as long as images
were the only concern there.

* [x] From `shoes` executable
  * [x] Downloading file by URL
  * [x] Image file from disk
* [x] From packaged jar/app (per OS)
  * [x] Windows
    * [x] Downloading file by URL
    * [x] Image file from disk
  * [x] Linux
    * [x] Downloading file by URL
    * [x] Image file from disk
  * [x] Mac
    * [x] Downloading file by URL
    * [x] Image file from disk
* [x] Samples (at least that have `image` calls in them)!


jasonrclark:
Yeah, as I've pondered it further I wonder how this will fare with memory and performance on large images. I'm planning to do more testing on this than usual...

Part of what I like about this approach is that it is _always_ executed, which gets us tons more natural testing than something that only happens in packaged apps.

That said, if we hit problems with it at all, we could do like the downloading does, and simply copy any image referenced from `uri:classloader` (i.e. inside the package) out to a temporary location and then let SWT read from that file instead.

Anyway, will see what the testing looks like and report back!

jasonrclark:
So, initial testing results (couldn't help myself 😀). Running with normal sized JPG images, I don't necessarily notice a huge difference, even when loading quite a few of them.

But to stress test things, I grabbed a [60MB JPG of the Hubble Deep Field](http://zebu.uoregon.edu/hudf/).

```
# packaged-image-loading
real	0m8.217s
user	0m12.597s
sys	0m0.687s

# master
real	0m6.173s
user	0m8.490s
sys	0m0.923s
```

So that looks like a 50% slowdown on image loading... feels like that's significant warrant trying out the temp file approach instead. What do you think @PragTob.

PragTob:
Hm. I'd like to see that benchmark to gauge it :D I.e. what are you measuring, the whole script? Just the image creation? :)

If possible I'd like to have a benchmark-ips of the creation of the image... but that might be hard-ish to get :|

Even then, 60MB is huge. If it really just occurs then it's fine imo as it's enough of an edge case for us to ignore atm (imo).

We could also roll with this and create a separate ticket to investigate performance, we should probably also add a sample to make sure that it keeps working even when we improve performance :)

jasonrclark:
Yeah, this was just a quick benchmark by doing `time` around the app startup and shutting down as soon as it was visible. Will see if I can get something more repeatable whipped up, especially so we can see if the impact is the same on smaller images.

PragTob:
Hm the problem might be graver then, if you really used `time script.rb` that'd include all of jruby/shoes startup which is already a couple of seconds and common to both. That'd mean that the image creation itself is much slower :(

jasonrclark:
All right, as expected it was a bit of effort, but worth it, to get benchmarking closer around the image creation. This creates `Shoes::Swt::Image` objects, which runs the image load during initialization. It required one line of code in the class to feature flag name vs byte loading. It also gloriously fails Rubocop in the benchmark code itself, but whatever path we'll choose I'll need to revise that significantly anyway, so :man_shrugging: for the time being.

As implied by the earlier `time`, there is a significant slowdown across all file sizes (small 1.3K PNG, medium 3.4M JPG, large 61MB JPG).

```
Warming up --------------------------------------
small.png by raw bytes
                       253.000  i/100ms
small.png by filename
                       618.000  i/100ms
small.png by raw bytes
                          2.791k (± 3.2%) i/s -     14.168k in   5.080999s
small.png by filename
                          6.544k (± 5.6%) i/s -     32.754k in   5.021881s

Comparison:
small.png by filename:     6543.7 i/s
small.png by raw bytes:     2791.4 i/s - 2.34x  slower

Warming up --------------------------------------
medium.jpg by raw bytes
                         1.000  i/100ms
medium.jpg by filename
                         1.000  i/100ms
medium.jpg by raw bytes
                         18.099  (± 5.5%) i/s -     91.000  in   5.036112s
medium.jpg by filename
                          5.582  (± 0.0%) i/s -     28.000  in   5.019234s

Comparison:
medium.jpg by raw bytes:       18.1 i/s
medium.jpg by filename:        5.6 i/s - 3.24x  slower

Warming up --------------------------------------
large.jpg by raw bytes
                         1.000  i/100ms
large.jpg by filename
                         1.000  i/100ms
large.jpg by raw bytes
                          0.793  (± 0.0%) i/s -      4.000  in   5.054145s
large.jpg by filename
                          0.292  (± 0.0%) i/s -      2.000  in   6.862805s

Comparison:
large.jpg by raw bytes:        0.8 i/s
large.jpg by filename:        0.3 i/s - 2.71x  slower
```

Looking at this, I think it's time to break out the temp file loading and see how that fares. Since I won't be able to run actual packaging, I'll test it out with just a straight file copy in the main executable path and see how that compares.

Off to hang with family in the sunshine... probably back to this tonight, since it's nerd sniping me hard :)

jasonrclark:
As expected it was a bit of effort, but worth it, to get benchmarking closer around the image creation. This creates `Shoes::Swt::Image` objects, which runs the image load during initialization. Minor shufflign in feature code to flag the three approaches--straight filename (which doesn't work with relative names in packages), copying to tmp, and reading bytes directly through Ruby.

Rubocop is unhappy, but that's great since it will keep things red until we pick which one and resolve it on the branch.

Full results are over in [this gist](https://gist.github.com/jasonrclark/2387ba3595e2c3260098211b119ceabb) but I'll pick out the salient details:

```
# 1.3K file
small.png by raw bytes:     5726.7 i/s
small.png by filename:     2390.3 i/s - 2.40x  slower
    small.png by tmp:      925.9 i/s - 6.19x  slower

# 194K
smallish.png by filename:       94.9 i/s
 smallish.png by tmp:       86.6 i/s - 1.10x  slower
smallish.png by raw bytes:       75.6 i/s - 1.26x  slower

# 3.4M
medium.jpg by filename:       16.3 i/s
   medium.jpg by tmp:       14.5 i/s - same-ish: difference falls within error
medium.jpg by raw bytes:        4.9 i/s - 3.32x  slower

# 61M
large.jpg by filename:        0.7 i/s
    large.jpg by tmp:        0.6 i/s - 1.16x  slower
large.jpg by raw bytes:        0.3 i/s - 2.52x  slower
```

Interesting result here, the smallest file size was consistently _faster_ with the byte reading approach this PR started with! But that clearly doesn't hold very long, as even up to about 200K for the file size the raw bytes is already much slower.

While the scaling isn't totally obvious, for anything of even moderate size, copying to a tmp location holds much closer to just letting SWT work off the filename (which we can't do within packages).

Upshot: I think we should take the tmpfile copying approach, but only do it when we detect that we're within a package and it's necessary. This keeps maximal performance with the normal execution, keeps as close as we can within packages, and the code is pretty straight-forward so I'm less worried about exercising it less often. That said, I need to extend the range of packaging samples that I use for releases this go around, since the existing single-file-no-images test I've been using isn't adequate anymore.

What say you @PragTob?

PragTob:
I'd be inclined to use tmp files all the way. Performance hit for very small files is sort of big but we could still do 900 of them in a second which is good enough. Afterwards the performance is almost on par and I think I'd rather make that tradeoff than implementing 2 code paths, I'd argue it's more pragmatic/easier to maintain. Should it turn out to be a big issue we could still optimize performance later on.

Thanks for benchmarking :heart:

jasonrclark:
I buy it... pushed up a largely untested (but Rubocop clear!) version of copying to tmp location along with cleanup. Done for the night with almost no testing, so that'll be what I pick up next. Ready to run the samples + packaging gauntlet!

PragTob:
Some test doesn't wanna play:

```
  1) Shoes::Image basic should update image path

     Failure/Error: subject.path = updated_filename

     

     Errno::ENOENT:

       No such file or directory - /tmp/__shoes4_1494306729.684769_shoes-icon.png

     # ./shoes-swt/lib/shoes/swt/image.rb:143:in `block in cleanup_temporary_files'

     # ./shoes-swt/lib/shoes/swt/image.rb:142:in `cleanup_temporary_files'

     # ./shoes-swt/lib/shoes/swt/image.rb:63:in `create_image'

     # ./vendor/bundle/jruby/2.3.0/gems/after_do-0.4.0/lib/after_do.rb:99:in `block in create_image'

     # ./shoes-swt/lib/shoes/swt/image.rb:103:in `display_image'

     # ./shoes-swt/lib/shoes/swt/image.rb:38:in `load_image'

     # ./shoes-swt/lib/shoes/swt/image.rb:27:in `update_image'

     # ./shoes-core/lib/shoes/image.rb:23:in `path='

     # ./shoes-core/spec/shoes/image_spec.rb:21:in `block in (root)'
```

jasonrclark:
Had wondered if I should wrap up that temp clean up in some rescues... looks like that's a "yes!"

jasonrclark:
Updated to fix the tests which look like they'll be happy now. Also extended to spec out the behavior was seeing on Travis.

Looking to get to the testing this weekend and will remove the WIP when that gets done.

shoes-swt/spec/shoes/swt/image_spec.rb
```diff
@@ -19,7 +19,7 @@
 
   subject do
     allow(dsl).to receive(:file_path) { image }
-    Shoes::Swt::Image.new(dsl, swt_app)
+    dsl.gui
```
PragTob: :)

jasonrclark:
Found two issues in testing. First, we weren't making a temp copy of the static 'downloading' image that we show for URLs, so that borked when packaged.

Second, cleanup happens after the SWT image is made. But for a downloading image, we actually end up making two separate images--the downloading one and the other. Since we didn't clear the list after the downloading temp was deleted, we get a false warning on the second. Easy enough to address.

Bit more testing and then I think I'll get this merged tonight. I've noticed some other things around images + packaging that we can improve, but they're grounds for another PR.

PR #1429, opened 2017-05-14
Image search paths

jasonrclark:
This further smooths the issue reported in #676.

Previously we were looking in `Dir.pwd` for relative file paths when showing an `image`. This had some big downsides:

* If you ran from another directory, relative image lookups could fail
* Packaged apps also didn't work nicely with this

Our workaround for the time being was to recommend that all images get loaded via a `File.expand_path` against a relative path constructed from `__FILE__`. This works, but is cryptic and not in the spirit of how Shoes does things.

This PR addresses that by introducing search paths for images. If you say `image 'my_image.png'` in your Shoes app, we'll now look in the following locations (in order):

1. `Dir.pwd` 
1. Shoes "app" dir, which we define as the directory where the starting file was located
1. `shoes-core/static`

This works across all three of the startup methods for Shoes (`shoes` executable, packaged, and running `ruby app.rb` with `require 'shoes'` in the file).

With this I think we've got a tidy, compelling story for handling images within packaged apps! 🎉


jasonrclark:
Extracted to `Shoes::Common::ImageHandling` so we can apply to image patterns as well as direct images.

jasonrclark:
Hmmm, build break here is actually because I just merged to master in furoshiki with the class renames and missed some things in `shoes-package` that needed updating. Will get that resolved, then rebase this once I get the chance... but not tonight.

jasonrclark:
Patched up the specs on master and rebased, so all happy now.

jasonrclark:
FWIW, likely to merge this one way or another tomorrow so I can move forward on buttoning up `pre10` with heaps of packaging goodness inside! 🎁

shoes-core/lib/shoes/common/image_handling.rb
```diff
@@ -0,0 +1,29 @@
...
+
+        unless found
+          raise FileNotFoundError, "#{path} not found. Searched #{locations.join(',')}"
+        end
+
+        found
+      end
+
+      def default_search_paths
+        [Dir.pwd, Shoes.configuration.app_dir, File.join(Shoes::DIR, "static")]
```
PragTob: Would be hard for this to be a constant, because load order and stuff... was thinking about memoizing it, but probably not worth the effort (just sharing thoughts!) :)

shoes-core/lib/shoes/color/dsl_helpers.rb
```diff
@@ -81,7 +83,11 @@ def image_file?(arg)
       end
 
       def image_pattern(path)
+        path = absolute_file_path(path)
         Shoes::ImagePattern.new path if File.exist?(path)
+      rescue FileNotFoundError => e
+        Shoes.logger.error(e.message)
+        return nil
```
PragTob: I wonder why we only rescure here and in no other place? Because we normally catch stuff elsewhere? But why here? Seems interesting to me.

jasonrclark:
Cora loooooves the bunnies! Thanks for sharing them ✨

shoes-core/lib/shoes/color/dsl_helpers.rb
jasonrclark: To be honest, this was just matching the existing spec'd behavior. It feels in line with Shoes' historical defaults that it just omits the background if it fails to find the file. Will ponder and dig a bit further.

shoes-core/lib/shoes/common/image_handling.rb
jasonrclark: At first blush I thought memoizing might be nice--just to express that this doesn't change--but then I considered that `Dir.pwd` _could_ change. Imagine building an image viewing app that works on directories... you could just `Dir.chdir` into those and then use the unexpanded filenames, which seems kinda nice.

PragTob: Definitely, we should keep it as is. Safest option. Make it work, make it beautiful, make it fast way later when actually needed :D

shoes-core/lib/shoes/color/dsl_helpers.rb
PragTob: Just seemed a bit out of place, but yeah I can imagine... all good will merge away :)

PragTob:
I heard someone say they love bunnies.....

img_20170523_192533

You/Cora can find more at https://www.instagram.com/berlin_bunny_trio/ :)

Anyhow enough with our cute little PR/review/catching up/sharing... I'll merge this :D

PR #1433, opened 2017-05-17
CI against JRuby 9.1.9.0

PragTob:
See if we run into the each_with_index problem outlined in jruby/jruby#4610

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ script:
   - bundle exec rake
 rvm:
   - jruby-head
-  - jruby-9.1.8.0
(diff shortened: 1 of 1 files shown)


jasonrclark:
Do we want to change the `.ruby-version` as well in lock step? Or just do this in CI until 9.1.10.0 is out?

PragTob:
ah most definitely, good catch.

jasonrclark:
Looks like I waited too long and 9.1.10.0 came out #1434 😆

PragTob:
Yeah 9.1.10 fixed the exact regression :)

Issue #1015, opened 2015-01-05
Make running a sample faster in the rake task


PragTob:
I just did my round of `rake samples` - as we do `system "bin/shoes #{sample_name}"` we pay the (significant) startup cost for EVERY sample - all 85 of them :( Takes pretty long... making it faster would be really nice.

If we just `require 'shoes'` and then `load` the samples it could be faster, although then we don't have a real isolation when running the samples... although I think there was some kind of argument to load...

Wouldn't it be nice... :)

jasonrclark:
That's a great idea. I would frequently use a faster solution, even if it loses some isolation.

Along those lines too, would be super-nice to have something that runs the samples against the gem installed copy of Shoes as well. I hacked up a batch file for it locally to test on Windows against the installed version, but would be awesome to have it built-in.

PragTob:
A naive implementation using `load` does not work due to....

```
Java::OrgEclipseSwt::SWTException: Invalid thread access
java.lang.reflect.Constructor.newInstance(java/lang/reflect/Constructor.java:526)
RUBY.initialize_shell(/home/tobi/github/shoes4/shoes-swt/lib/shoes/swt/app.rb:196)
```

I guess when quitting the app we don't shut it down correctly or SWT simply dislikes starting new apps in the same process. Probably worth ivnestigating.

jasonrclark:
I found similar problem in poking at this--if the sample file that we load is the first/main app, then things get really cranky when we shut down.

In lieu of that, though, I spun up this little Shoes app (hacky, copies some stuff from Rakefile for the moment) which gets the effect we want of fast loading samples. Single keypress loads the next sample, with the super+w (Apple+w on the Mac) to shut the spawned windows. Will try this out with my next round of sample testing as it is much, much faster.

```
Shoes.app do
  title "Press a key to load next sample", align: "center"

  def samples_from_file(filename)
    samples = File.read("samples/#{filename}").lines
    samples.map { |s| s.sub(/#.*$/, '') }.map(&:strip).select { |s| s != '' }.map { |s| "samples/#{s}" }
  end

  def working_samples
    samples = samples_from_file "README"
    puts "#{samples.size} samples are known to work"
    samples
  end

  Shoes::InternalApp.add_global_keypress(:super_w) do
    Shoes.apps.last.quit
  end

  samples = working_samples
  keypress do |*args|
    quit if samples.empty?

    load samples.shift
  end
end
```

jasonrclark:
[...]

PR #1439, opened 2017-05-31
Helper scripts to run samples faster

jasonrclark:
As noted over in #1015, startup time on samples is a drag.

I'd fiddled before with [my own suggestion in the issue](https://github.com/shoes/shoes4/issues/1015#issuecomment-262448017), but multiple samples coexisting and reloading into the same process has some gnarly side-effects. It just isn't the same as launching the apps by themselves per the normal process.

Fortunately, I found a way to get around it 🎉. The basic approach is to just keep two processes running at all times (until we're done): the current one, and the next one. The next one is held in a pending state until the current completes, then we signal next to proceed and spawn a new "next" process. Each process is fully independent, but the next one can hit JRuby boot-time while we're busy interacting with the preceding sample!

Marked as WIP because I don't have it hooked up to the Rake tasks and will give it a thorough run-through to get my first pass of `pre10` testing done. So far it's gotten things down to just 1 second or so between sample apps showing up vs 5-7 seconds previously.

TODO:
* [x] Testing
* [x] Rakefile hookup
* [x] Make Rubocop happy

### DIFF
--- a/bin/run-sample
+++ b/bin/run-sample
@@ -1,0 +1,26 @@
+#!/usr/bin/env ruby
+
+puts "STARTED"
+puts "Child process #{$$} started"
+
+sample = ARGV[0]
+parent = ARGV[1].to_i
+ready  = ARGV[2] == "true"
+
+require 'bundler/setup'
+require 'shoes'
+
+# Signal we expect from parent when we're set to go
+Signal.trap("HUP") do
+  ready = true
+end
+
+# Let parent know we're ready
+Process.kill("HUP", parent)
+
+# Wait until we're signaled
+until ready
+end
+
+puts "Running #{sample} in #{$$}"
+load sample
--- a/bin/run-sample
+++ b/bin/run-sample
@@ -1,14 +1,20 @@
 #!/usr/bin/env ruby
-
-puts "STARTED"
-puts "Child process #{$$} started"
+# frozen_string_literal: true
 
 sample = ARGV[0]
 parent = ARGV[1].to_i
-ready  = ARGV[2] == "true"
+index  = ARGV[2].to_i
+total  = ARGV[3].to_i
+ready  = false
 
 require 'bundler/setup'
 require 'shoes'
+
+# A few samples tamper with things so we need to force backend loading early
+$LOAD_PATH.unshift(File.dirname(sample))
+
+require 'shoes/swt'
+Shoes::Swt.initialize_backend
 
 # Signal we expect from parent when we're set to go
 Signal.trap("HUP") do
@@ -19,8 +25,7 @@ end
 Process.kill("HUP", parent)
 
 # Wait until we're signaled
-until ready
-end
+sleep 0.2 until ready
 
-puts "Running #{sample} in #{$$}"
+puts "Running #{sample} (#{index} of #{total})...quit to run next sample"
 load sample
--- a/bin/run-samples
+++ b/bin/run-samples
@@ -1,0 +1,37 @@
+#!/usr/bin/env ruby
+
+require 'open3'
+require 'thread'
+
+$signals = Queue.new
+
+Signal.trap("HUP") do
+  $signals << true
+end
+
+puts "Parent process #{$$}"
+
+def start(sample, ready = false)
+  return nil unless sample
+
+  Process.spawn("bin/run-sample #{sample} #{$$}")
+end
+
+def run(pid)
+  # Make sure child has signaled back before we tell it to proceed...
+  $signals.pop
+  Process.kill("HUP", pid)
+  Process.wait(pid)
+end
+
+samples = ARGV.dup
+
+current_child = start(samples.shift)
+next_child    = start(samples.shift)
+run(current_child)
+
+until next_child.nil?
+  current_child = next_child
+  next_child = start(samples.shift)
+  run(current_child)
+end
--- a/bin/run-samples
+++ b/bin/run-samples
@@ -1,6 +1,14 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+# Because of JRuby startup time, it can be a drag running all 100+ samples.
+# This helper alleviates some of the pain.
+#
+# The basic idea is to just have two processes running at all times--one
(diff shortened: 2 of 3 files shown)


bin/run-samples
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env ruby
+
+require 'open3'
+require 'thread'
+
+$signals = Queue.new
```
PragTob: huh, globals? Is there a reason we'd use a global here and not an instance variable?

bin/run-samples
```diff
@@ -0,0 +1,37 @@
...
+
+Signal.trap("HUP") do
+  $signals << true
+end
+
+puts "Parent process #{$$}"
+
+def start(sample, ready = false)
```
PragTob: we don't use `ready` ?

bin/run-samples
```diff
@@ -0,0 +1,37 @@
...
+require 'open3'
+require 'thread'
+
+$signals = Queue.new
+
+Signal.trap("HUP") do
+  $signals << true
+end
+
+puts "Parent process #{$$}"
```
PragTob: For the lazy... could we use the english magic globals, there was such a thing wasn't there? I have no idea what this is... I could look it up but... :man_shrugging:

bin/run-samples
```diff
@@ -0,0 +1,37 @@
...
+  Process.wait(pid)
+end
+
+samples = ARGV.dup
+
+current_child = start(samples.shift)
+next_child    = start(samples.shift)
+run(current_child)
+
+until next_child.nil?
```
PragTob: somehow something like `while samples.any?` somehow reads better to me, but yeah we'd potentially forget one then if it's just 2 samples right?

bin/run-samples
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env ruby
+
+require 'open3'
+require 'thread'
```
PragTob: where are we using thread?

bin/run-sample
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env ruby
+
+puts "STARTED"
+puts "Child process #{$$} started"
+
+sample = ARGV[0]
+parent = ARGV[1].to_i
+ready  = ARGV[2] == "true"
```
PragTob: I think this argument is never used?

bin/run-sample
```diff
@@ -0,0 +1,26 @@
...
+Signal.trap("HUP") do
+  ready = true
+end
+
+# Let parent know we're ready
+Process.kill("HUP", parent)
+
+# Wait until we're signaled
+until ready
+end
```
PragTob: we might wanna sleep in here to prevent the CPU from running hot?

PragTob:
ah rubocop is unhappy.. plus I think we should run rubocop first.

bin/run-samples
jasonrclark: Oddly enough, that's where `Queue` comes from.

bin/run-samples
jasonrclark: I was doing a lot of hacking around on the loop early on, and the global was expedient. Will definitely revise to something more tidy.

bin/run-samples
jasonrclark: Yep, another vestige of early attempts.

bin/run-samples
jasonrclark: Yeah, the check against `next_child` is necessary to get the loop counting right, but could do `while next_child` instead of what I've got. Will play with it.

bin/run-sample
jasonrclark: 🔪

bin/run-sample
jasonrclark: Will see how it feels with a short sleep there.

jasonrclark:
Significant tidying... still got some testing to do over the weekend, but I'm pretty happy with where this landed.

bin/run-samples
PragTob: such intuitive, many wow.

bin/run-sample
```diff
@@ -19,8 +25,7 @@ end
 Process.kill("HUP", parent)
 
 # Wait until we're signaled
-until ready
-end
+sleep 0.2 until ready
```
PragTob: :+1:

bin/run-sample
```diff
@@ -19,8 +25,7 @@ end
...
 
-puts "Running #{sample} in #{$$}"
+puts "Running #{sample} (#{index} of #{total})...quit to run next sample"
```
PragTob: :+1:

bin/run-samples
```diff
@@ -1,37 +1,76 @@
...
 
-require 'open3'
+# Because of JRuby startup time, it can be a drag running all 100+ samples.
+# This helper alleviates some of the pain.
+#
+# The basic idea is to just have two processes running at all times--one
+# current, the other in a pending state waiting to be signaled to start up.
+# While you're using the current sample, the next process is getting through
+# the JRuby load-time... winning!
+#
```
PragTob: good why comment, thanks!

bin/run-samples
```diff
@@ -1,37 +1,76 @@
...
-  return nil unless sample
+    @run_count += 1
+    Process.spawn("bin/run-sample #{sample} #{Process.pid} #{@run_count} #{@total_count}")
+  end
 
-  Process.spawn("bin/run-sample #{sample} #{$$}")
-end
+  def run_process(pid)
+    # When we're done, you'll get a nil pid so bail
+    return unless pid
```
PragTob: good why comment!

bin/run-samples
```diff
@@ -1,37 +1,76 @@
...
+  end
+
+  def signal_child_to_proceed(pid)
+    Process.kill("HUP", pid)
+  end
+
+  def wait_for_child_to_finish(pid)
+    Process.wait(pid)
+  end
 end
```
PragTob: yay encapsulation!

jasonrclark:
Testing went smoothly through, so 🚢 !

Issue #1407, opened 2017-03-12
Implement Copy and Save button for Console


jasonrclark:
Press `Alt+/` to bring up the Shoes Console. Back on Shoes 3 it looked like this:

![image](https://cloud.githubusercontent.com/assets/130504/21001024/e6cd06d2-bcd2-11e6-8a3e-3111a1b1ce62.png)

Shoes 4's console is missing the Copy and Save buttons, and it'd be nice to have those back!

estensland:
I am interested in helping out, so if this issue is fair game I will take a stab at it.

jasonrclark:
It's wide open @estensland, so have at it! Let me know if there's any help I can give you on getting set up and running.

estensland:
Thanks, I seem to have it all working. I have one dumb newcomer question, but where in the repo does it make the most sense to write specs for this?

jasonrclark:
Not a dumb question at all, especially given that this appears to lack existing specs :|

I'd recommend that we add a spec in `shoes-core`  at `shoes-core/spec/shoes/console_spec.rb` to get it covered. Sorry there isn't preceding work there to base things off, so hit me up if you need any help with approaching tests there.

estensland:
@jasonrclark Appreciate the offer of help! Any existing tests I can look at for pointers on how to test the code block for buttons?  Like the existing clear button:

    button "Clear", margin: 6, width: 80, height: 40, right: 10 do
       @messages.clear

       @message_stacks.each(&:remove)
       @message_stacks.clear
     end

If there are no existing tests to base off of elsewhere, I have an idea to it best covered

jasonrclark:
@estensland I honestly can't think of another case inside of Shoes itself where we have tests against something that's running DSL code in this fashion. @PragTob any ideas there?

Look forward to seeing what you come up with!

estensland:
@jasonrclark My first thought was just have the buttons call methods defined on the console class, so they can just be tested there. The code inside the block would not be covered, but the code called would be.

But...I am thinking now if the mock classes are only used in testing, we can assign the button's given block as an instance variable, then call the block in a spec to see if it works. It is a little more complicated, but as long as Mock::Clickable is only used by rspec, it seems to be working.

Sorry if overthinking it :man_shrugging:

jasonrclark:
@estensland Totally not overthinking it :)

[...]

PR #1447, opened 2017-06-16
Copy and save button for console issue#1407

estensland:
For issue #1407 

So I assumed that the output for copy and the save file were the same, formatted like the shown console. Since they shared, I broke that code out into an instance method on console. I based getting it into the app off how the @messages instance variable is passed in.

Testing was a pain, as you can see by ugly code as I dive through objects to check instance variables. Not a huge fan and sure there is a better way. I tried unit testing the different console methods in addition to the new code I wrote, this was mostly part of me figuring out how console works, so I included them but can remove if you deem them worthless. I added methods to mock classes to further test things, the dialog chooser was part of the follow issue:

Main issue I ran into is testing the save button. Below you can see I am having it briefly create a file in the home directory, check it, then delete it. Should I save it to a tmp directory? Is there a better way to handle this?

Let me know thoughts you have for fixes/improvements/rewrites. Thanks in advance for checking it over, hopefully I am at least on the right track.


jasonrclark:
Hey @estensland thank you for the PR!

I'm in the process of making our next release out, and that part-way done stuff is what's broken the build on this PR. Once I get the release out, I'll ping here and at that point if you rebase to master things should go 🍏 

I will endeavor to give the PR some reading/testing over the weekend. I love how thorough you've been on the cases for the specs!

I'm keen to find a structure that avoids the instance variable munging, but it isn't immediately obvious so I'll see what I can come up with to recommend. I appreciate all the effort you've put into this.

On the file writing, that's totally fine to touch the disk. You might consider putting it in `Dir.tmpdir` if possible, so it's really in a temporary location, but we have plenty of tests (especially in packaging) that hammer the disk so this is no big deal.

estensland:
@jasonrclark Thanks for feedback and checking it out. One thought I had was just passing the entire console object in, now that it wants @messages and a method, but not sure if that's overkill. 

I'll try changing to Dir.tmpdir, that sounds like a much better solution. Thanks!

jasonrclark:
Yeah, something passing an object into the `Shoes.app` that it interacts with might make things a lot easier.

Changes are on master that should get the build green again. Feel free to rebase and push at your leisure.

jasonrclark:
Got a partial approach up at https://github.com/estensland/shoes4/pull/1 that will likely make the testing easier here.

estensland:
@jasonrclark I have some warning to deal with before this is ready, but wanted to get your feedback on specs.

I fought a long time with them and finally this is where I ended up. The new test for the save method on the console object would actually open the file dialog and thus mess up the test when running the whole test suite. Similarly, the button mock clicking would fail on the whole test suite. Since running just that file in rspec they would pass...I eventually just added "no_swt: true" to them.

I also have the question if the whole mocking to test the button clicks is necessary in your opinion. Since the buttons just each call a single method on console, and I tested those, is testing the button clicks redundant to you?

jasonrclark:
Howdy @estensland. Sorry for the delay here... I'm moving, and off-hours coding time has gotten short :)

Reading over things and pondering, I believe I'd be totally happy with us just testing methods directly on `Shoes::Console` and not worrying about the `click` bindings. Those are so trivial that I'm not too concerned about them regressing, and doesn't seem worth bending the specs too far.

Have a few other inline comments to make, but as soon as I'm settled in after the move this'll be the first thing I take a closer look at.

shoes-core/spec/shoes/console_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+
+  let(:sample_message_array) { [[:debug, 'debug message'], [:warn, 'warn message'], [:error, 'error message']] }
+  let(:formatted_message_output){ "Debug\n debug message\n\nWarn\n warn message\n\nError\n error message\n\n" }
+
+  let(:mock_dialog) { Shoes::Mock::Dialog.new.dialog_chooser }
+
+  before(:each) do
+    # Depending on spec order, these can be leftover and contaminate other specs
+    console.instance_variable_set(:@app, nil)
+    console.instance_variable_set(:@messages, [])
```
jasonrclark: Rather than resetting the variables, could we just change `let(:console)` to return a totally new `Shoes::Console` instance? Then it'd be guaranteed to be clear?

shoes-core/spec/shoes/console_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+  end
+
+  describe '#initialize' do
+    it 'must reset @messages as []' do
+      console.instance_variable_set(:@messages, ['placeholder'])
```
jasonrclark: You shouldn't need to set the instance variable here directly. With the reader already on `messages`, a simple `console.messages << 'placeholder'` (assuming our startup is clean) would suffice.

shoes-core/spec/shoes/console_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+  describe '#create app' do
+    let!(:console_app) do
+      console.instance_variable_set(:@messages, sample_message_array.dup)
+      console.create_app
+      console.instance_variable_get(:@app).instance_variable_get(:@__app__)
+    end
+
+    let!(:console_app_nested_contents) { console_app.contents.first.contents.first.contents }
+
+    context 'the contents ordering' do
```
jasonrclark: Honestly would be inclined to not test the specific ordering of this, since it introduces friction if we want to shuffle the sequence/nesting of things around in the UI to get the appearance we're after. Thank you for being thorough, though!

shoes-core/spec/shoes/console_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+          returned = call_button_mock_block(console_app_nested_contents[2])
+
+          expect(returned).to eq(formatted_message_output)
+        end
+      end
+
+      context 'when @messages empty' do
+        it 'must return nil', no_swt: true do
+          console.instance_variable_set(:@messages, [])
+          console.instance_variable_set(:@message_stacks, [])
```
jasonrclark: Both of these could just call `clear` through their accessor.

jasonrclark:
Looks like Rubocop has some issues too--pretty straightforward stuff, but let me know if you've got questions on any of them. You can either run Rubocop directly, or our default `rake` task should include that along with the specs.

estensland:
@jasonrclark thanks for all the feedback! Just need to finish with the Dir.tmpdir, just wanted to check that what I had done so far was happy with the Integration checks as it was on my local. Hopefully the Dir.tmpdir will take care of AppVeyor's issue

estensland:
@jasonrclark Have you dealt with the AppVeyor permission issue before? If not I can do some more research

jasonrclark:
Thanks for the updates! I haven't dealt with temp files before on AppVeyor unfortunately. I was going to suggest that we look at how the `shoes-package` tests work (since they write a lot to disk), but I then I remembered we only run the `shoes-core` tests in that env :(

@PragTob any particular ideas?

If we need to mock some bits when on Windows (or even under AppVeyor specifically?), there's an example how to detect that over in https://github.com/shoes/shoes4/blob/master/shoes-package/spec/spec_helper.rb#L48. If we can't find anything about writeable paths with them, I'm fine with that.

@estensland, seriously the testing effort you've put on this PR has been above and beyond! So appreciate it! 💖

jasonrclark:
🎉

PR #1451, opened 2017-06-19
New rspec, rubocop

jasonrclark:
Punted on actually fixing all the new 👮  since oh my, so many fiddly changes. Like literally were almost 600 offenses. Anyone who wants to pick away at them can run on #945 quite a while longer.


PragTob:
:tada:

PR #1468, opened 2017-08-23
Enable focus on elements that need it

### LINKED ISSUE #1458: Focus method for edit_line not available
It appears that the focus method for the input_box class is not implemented. 

The following code will recreate the issue:
```ruby
class Focus < Shoes
  url '/', :index
  url '/text_input', :text_input

  def index
    para link("enter data", click: "/text_input")
  end

  def text_input
    @input = edit_line width: 400, top: 20, left: 20
    @input.focus()
  end
end
```

After clicking on "enter data" the following error will be reported:
```
ERROR: undefined method `focus' for #<Shoes::Swt::EditLine:0xff6077>
```

(If `edit_line` is called in the `:index` method, the input will still receive focus automatically despite the failure to focus, so it isn't immediately clear there is an issue.)

Untested solution: Add the following to `shoes-swt/lib/shoes/swt/input_box.rb`
```ruby
def focus
     @real.set_focus()
end
```

jasonrclark:
Fixes #1458

Adds focus to:
* `edit_line`
* `edit_box`
* `list_box`


shoes-core/lib/shoes/common/focus.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+class Shoes
+  module Common
+    module Focus
+      def focus
+        @gui.focus
+      end
```
PragTob: :+1:

PR #1471, opened 2017-09-19
Get samples/simple_accordion.rb working again

### LINKED ISSUE #1440: samples/simple_accordion.rb broken
* [ ] ~~Code includes reference to `Link` without qualifying `Shoes`. Would be nice to actually support that for compatibility.~~
* [ ] ~~Missing `app.slot` method once the styling's fixed~~

jasonrclark:
Fixes #1440 

Sample was busted on both Shoes 3 and 4, so this gets it back into working order on both.

Noticed that we weren't getting proper redraws after the flush when expanding the bottom element on the screen. Will keep a close eye in testing that this doesn't regress anything, but definitely appears to be the correct thing to do at this moment.

### DIFF
--- a/shoes-swt/lib/shoes/swt/redrawing_aspect.rb
+++ b/shoes-swt/lib/shoes/swt/redrawing_aspect.rb
@@ -35,8 +35,9 @@ class Shoes
                                                        :mouse_left],
                            ::Shoes::TextBlock      => [:replace] }.freeze
 
-      CHANGED_POSITION = { ::Shoes::Common::Positioning => [:_position, :displace],
-                           ::Shoes::Common::Hover       => [:eval_hover_block] }.freeze
+      CHANGED_POSITION = { ::Shoes::DimensionsDelegations => [:adjust_current_position],
+                           ::Shoes::Common::Positioning   => [:_position, :displace],
+                           ::Shoes::Common::Hover         => [:eval_hover_block] }.freeze
 
       attr_reader :app
 
--- a/samples/README
+++ b/samples/README
@@ -7,6 +7,7 @@
 # Curious to know more? See https://github.com/shoes/shoes4/wiki/Samples-Guide
 # for more information about writing samples!
 
+simple_accordion.rb
 simple_alert.rb
 simple_altered_para.rb
 simple_anim_shapes.rb
--- a/samples/simple_accordion.rb
+++ b/samples/simple_accordion.rb
@@ -1,50 +1,68 @@
 # frozen_string_literal: true
-#
-# Broken until https://github.com/shoes/shoes4/issues/1440
-#
 module Accordion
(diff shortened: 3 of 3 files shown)


shoes-swt/lib/shoes/swt/redrawing_aspect.rb
```diff
@@ -79,7 +79,10 @@ def add_redraws
       end
 
       def update_gui
-        app.flush unless app.disposed?
+        unless app.disposed?
+          app.flush
+          app.redraw
```
PragTob: huh, this doesn't seem right as we tell it to redraw the whole UI? IIRC flush was also just "ok now really do those redraws I been telling you to do" so order also seems a bit weird. It's been a long time, I'm not sure this is the appropriate fix, although I can see how it fixes things :)

jasonrclark: So I admit that this did strike me as a heavy hammer. `flush`, though, is [only telling the UI to do it's layout again](https://github.com/shoes/shoes4/blob/master/shoes-swt/lib/shoes/swt/app.rb#L120). That doesn't seem to always correctly trigger the redrawing area in SWT, especially when our shown region gets bigger as a result of a change (like the animation here).

I'm going to give this more reading and some solid testing through the heavier samples before moving forward. Will report back.

PragTob: My idea/hope would be that if we correctly redraw all areas/objects that are moved during the animation then we shouldn't need the whole app redraw... of course can't totally say :)

jasonrclark: Yeah, it does seem like something is probably being missed at the lower levels there. Looking it over, I grow a bit suspicious that maybe our redrawing aspect hook is missing something it needs on resizes for elements. I would have thought `_position` would handle it, but maybe not. Will investigate.

PragTob: One thing that springs to my mind, we only redraw the new location but maybe not the old position, but not sure how to get that :scream:

jasonrclark:
As discussed, this was actually triggered by us lacking the right hooks in redrawing aspect for certain slot resizing conditions (which an intermediate form of my accordion fixes revealed, although the full version doesn't.) Basically if you had something on the bottom of all the known slots, expanding it downward didn't trigger redraws for the element like it should have.

I've reverted the more aggressive redraw in favor of just hinging off a method we use during slot positioning. Will still give this some solid sample testing, as I can imagine this method getting hit a lot and maybe causing perf issues, but I'm not too worried.

jasonrclark:
Tested out a good portion of samples, including all the heavy expert ones (Game of Life, Tankspank) and no noticable lag from the additional redraws this presumably schedules. 💃

PR #1472, opened 2017-09-21
Report right/bottom if possible

### LINKED ISSUE #889: #right returns nil if unset - report it relative to parent like left?
Hello everyone,

this is somewhat of a discussion. I believe we just changed this behavior some time ago, probably I even did it. So just to know what I'm talking about:

``` ruby
Shoes.app do
  circle = oval 40, 40, 50
  button 'report' do
    p circle.left
    p circle.right
  end
end
```

`circle.right` will always report nil, while circle.left reports that value. This is also what shoes3 does. However it's kind of bad 0 sometimes I just want to know the right value of the thing I'm having. 

We _could_ use `absolute_right`/`element_right` for this but it's not the prettiest thing.

What if `right` were just to report the value relative to its parent like left if no explicit right value is set?

FYI I discovered this digression when I tried out a program I was writing in shoes4 a year ago:

``` ruby
Shoes.app title: 'bouncing ball' do
  strokewidth 5
  circle = oval 40, 40, 100
  SPEED = 10
  xspeed = SPEED
  yspeed = SPEED
  animate 30 do
    xspeed *= -1 if circle.right >= width || circle.left <= 0
    yspeed *= -1 if circle.top <= 0 || circle.bottom >= height
    circle.move circle.left + xspeed, circle.top + yspeed
  end
  click do
    circle.fill = rgb rand(255), rand(255), rand(255)
  end
end
```

Of course, that'd break shoes3 compatibility... so probably not. Maybe 4.1. Maybe 5.0. Maybe now? Ideas?

jasonrclark:
Fixes #889 

That issue raises the question of breaking compatibility with Shoes 3 over the calculation of right (and as it turns out bottom) on elements that don't explicitly set it. This seems like a totally reasonable thing to me, and easy enough to accomplish as you can see with this change.

From a user's standpoint, this is clearly the behavior I'd expect so I'm good with making it that way. Only a single test broke, and it didn't provide rationale as to _why_ this behavior was good--just documented that we do it.

Will keep an eye out for broken samples in later testing, but almost nothing actually refers to `right` so I'm not tromping through them all just yet.

### DIFF
--- a/shoes-core/lib/shoes/dimension.rb
+++ b/shoes-core/lib/shoes/dimension.rb
@@ -27,7 +27,7 @@ class Shoes
     end
 
     def end
-      @end || report_relative_to_parent_end
+      @end || report_relative_to_parent_end || element_end
     end
 
     def extent
--- a/shoes-core/spec/shoes/dimension_spec.rb
+++ b/shoes-core/spec/shoes/dimension_spec.rb
@@ -190,7 +190,7 @@ describe Shoes::Dimension do
       subject.extent = 45
     end
 
-    its(:end) { should be_nil }
+    its(:end) { should eq(67) }
     its(:start) { should be_nil }
   end
 


PragTob:
The whole of the positioning stuff has been some time ago... and I think we didn't report `element_end` because `element_*` is the total position app wise while `left/right/top/bottom` are often relative to something (parent slot or someting) and hence not absolute, so I think the value might be unexpected.

What do we do with left if it's not set? It looks like we might also end up with `nil` left/start values:

```
     def basic_start_value
       value = @start
       value = calculate_relative value if @start_relative
       value = report_relative_to_parent_start if value.nil?
       value
     end
```

so overall, I'm not entirely sure if this change is desirable :)

jasonrclark:
Yeah I'm on the fence. With left there I expect we might get the relative parent value in a lot of cases because left gets set by positioning where right doesn't.

The original intent does seem to be that the left/right is what's been set, element_* more calculated (with margins) and absolute_* the exact position.

If you're cool with it @PragTob we can just close this and the issue until someday someone can explain differently. Cool?

PragTob:
haha totally missed the issue xD Yeah I guess we can close it the issue was more of a question from past me

PR #1475, opened 2017-09-24
Fix widgets to behave like real elements

### LINKED ISSUE #641: Shoes::Widget dimensions don't behave like Shoes3
Noticed this in fixing up #528, but working around it in the sample for now.

Widgets in Shoes3 have dimension setting methods, and abide by the values. It also seems to pick them up from parameters initially passed into the widget creation DSL method.

Shoes4 doesn't allow setting dimensions (like width), and doesn't default the parameters on creation either.

Sample code:

```
class MyWidget < Shoes::Widget
  def initialize(color, args)
    background color
  end
end

Shoes.app do
  my_widget green,  width: 170, height: 120
end
```

Shoes3
screen shot 2014-04-11 at 11 58 38 am

Shoes4
screen shot 2014-04-11 at 11 58 48 am

jasonrclark:
Fixes #641

This PR changes `Widget` to behave essentially as a `Flow` so it gets all the standard dimensions and styling it wasn't obeying before.

This has a big challenge, though, as the contract for `Widget` derived classes in older Shoes didn't for calling `super` from `#initialize`. We must run `UIElement#initialize` (or an equivalent) to get everything wired up, but we also need to run the user-defined `initialize` so the user's code runs.

Ruby's flexibility to the rescue! ✨  We can take the odd step of actually skipping `new` calls for the widgets and just `allocate` and `initialize` (in the order and timing we want, along with our base initialization) directly. 😱 Never thought I'd find a use for that bit of trivia, but here we are!

Other tidbits:

* Widget delegation before was touchy because it tried to delegate EVERYTHING to the app except a small set of methods we'd called out. Turns out, that's a bad plan, so instead now we delegate everything that the widget class doesn't define itself already.
* Added a new way for DSL classes to declare their own backend class directly for cases where the class names don't necessarily line up (ala people's custom-named classes, which need to be backed by `Shoes::Swt::Flow`)

Been using this app, in addition to the two samples with widgets (`samples/simple_menu.rb` and `samples/simple_guess_game.rb`):

```
class MyWidget < Shoes::Widget
  def initialize(color, args)
    background color
    para "thing"
    para "one"
    para "thing"
    para "one"
  end
end

Shoes.app do
  @w = my_widget green, width: 170, height: 120
  click do
    @w.left += 10
    @w.top += 10
  end
end
```

### DIFF
--- a/shoes-core/lib/shoes/widget.rb
+++ b/shoes-core/lib/shoes/widget.rb
@@ -21,40 +21,38 @@ class Shoes
   #     say_hello 'Shoes'
   #   end
   #
-  class Widget
-    include Common::Inspect
-
+  class Widget < Shoes::Flow
     Shoes::App.subscribe_to_dsl_methods self
 
-    attr_accessor :parent
-    attr_writer :app
+    attr_accessor :original_args
 
-    class << self
-      attr_accessor :app
+    def handle_block(*_)
+      old_current_slot = @app.current_slot
+      @app.current_slot = self
+      initialize(*original_args)
+      @app.current_slot = old_current_slot
     end
 
-    # lookup a bit more complicated as during initialize we do
-    # not have @app yet...
-    def app
-      @app || self.class.app
+    def shoes_base_class
+      Shoes::Widget
     end
 
     def self.inherited(klass, &_blk)
       dsl_method = dsl_method_name(klass)
-      Shoes::App.new_dsl_method(dsl_method) do |*args, &blk|
-        # we set app 2 times because widgets execute most of their code
-        # straight in initialize. I dunno if there is a good way of setting
-        # an @app instance variable before initialize is executed. We could
-        # hand it over in #initialize but that would break the interface
-        # and people would have to set it themselves or make sure to call
-        # super so for not it's like this.
-        # Setting the ref on the instance is important as we might have
-        # instances of the same widget in different Shoes::Apps so each one
-        # needs to save the reference to the one it was started with
-        klass.app              = self
-        widget_instance        = klass.new(*args, &blk)
-        widget_instance.app    = self
-        widget_instance.parent = @__app__.current_slot
+      Shoes::App.new_dsl_method(dsl_method) do |*args|
+        # If last arg is a Hash, pass that to the underlying Flow
+        container_args = args.last.is_a?(Hash) ? args.last : {}
+
+        # What's this business? Because of the old contract about how widgets
+        # define initialize, we actually have TWO initializes we need to call--
+        # the base UIElement one and the user-defined one which doesn't super.
+        #
+        # Instead of using new, we take over for that by calling allocate
+        # ourselves, then invoking the base-initialize equivalent, which in
+        # turn (via handle_block above) will do the widget class's initialize.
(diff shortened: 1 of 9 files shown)


shoes-core/lib/shoes/widget.rb
```diff
@@ -21,40 +21,41 @@ class Shoes
...
+        # What's this business? Because of the old contract about how widgets
+        # define initialize, we actually have TWO initializes we need to call--
+        # the base UIElement one and the user-defined one which doesn't super.
+        #
+        # Instead of using new, we take over for that by calling allocate
+        # ourselves, then invoking the base-initialize equivalent, which in
+        # turn (via handle_block above) will do the widget class's initialize.
+        widget_instance = klass.allocate
+        widget_instance.original_args = args
+        widget_instance.actual_initialize(@__app__, @__app__.current_slot, container_args)
```
PragTob: Not sure I'm too happy with this magic of calling `initialize`/`allocate` ourselves - how do you feel about breaking compatibility and calling this method `initialize_widget` instead or something?

PragTob: Or just require calling super... one of the 2. I dunno, seems like a little breaking change but one worth making imo - don't wanna dabble with initialize and allocate tbh :)

shoes-core/lib/shoes/widget.rb
```diff
@@ -21,40 +21,41 @@ class Shoes
...
+  class Widget < Shoes::Flow
     Shoes::App.subscribe_to_dsl_methods self
 
-    attr_accessor :parent
-    attr_writer :app
+    attr_accessor :original_args
 
-    class << self
-      attr_accessor :app
+    def handle_block(*_)
```
PragTob: just one thing, do we have redrawing etc. correctly setup for this?

shoes-core/lib/shoes/widget.rb
```diff
@@ -21,40 +21,41 @@ class Shoes
   #     say_hello 'Shoes'
   #   end
   #
-  class Widget
-    include Common::Inspect
-
+  class Widget < Shoes::Flow
```
PragTob: why do we inherit from flow and not Slot? Cause flow is the default, makes sense, just checking though

jasonrclark: I checked the behavior on the old Widgets, and they acted like a Flow.

shoes-core/lib/shoes/widget.rb
jasonrclark: How do you mean? Because of where this is called it wires into the startup sequence correctly, and after that it's just a slot so everything's happy.

shoes-core/lib/shoes/widget.rb
jasonrclark: So calling `initialize_widget` would be the path (`super`'ing doesn't really work well with how we have `UIElement` set up). It would completely break anything using widgets, though, which seems like a bummer.

If I can suss out a way to give a more direct warning about the situation, I'd be happier about breaking compatibility. Let me see what I can find.

jasonrclark:
Closing this in favor #1476

PR #1476, opened 2017-09-24
Fix widgets without doing manual allocate

jasonrclark:
Per @PragTob's suggestion on https://github.com/shoes/shoes4/pull/1475 here's an alternate PR that changes our contract with child classes to expect a `initialize_widget` instead of `initialize`.

TODOs
* [x] When the widget method is called, warn on whether the class disobeys our new rules for `initialize`. This is the best time to do it, since our `inherited` hooks happen too early to ensure everything's right.
* [x] Ahem, fix the tests so they pass 😓 
* [x] Also will want some docs/wiki updates about the change, probably on https://github.com/shoes/shoes4/wiki/Compatibility-with-Shoes-3

### DIFF
--- a/shoes-core/lib/shoes/common/ui_element.rb
+++ b/shoes-core/lib/shoes/common/ui_element.rb
@@ -14,16 +14,6 @@ class Shoes
       attr_reader :app, :parent, :dimensions, :gui
 
       def initialize(app, parent, *args)
-        actual_initialize(app, parent, *args)
-      end
-
-      # Why is this separate? With Widgets, we need to invoke this base
-      # initialize, but ALSO a user-define initialize that historically didn't
-      # require calling super!
-      #
-      # Extracted like this, we can call actual_initialize to get base work
-      # done, and then initialize on the class ourselves to get the widget set.
-      def actual_initialize(app, parent, *args)
         blk    = args.pop if args.last.is_a?(Proc) || args.last.nil?
         styles = args.last.is_a?(Hash) ? args.pop : {}
 
--- a/shoes-core/lib/shoes/common/ui_element.rb
+++ b/shoes-core/lib/shoes/common/ui_element.rb
@@ -14,6 +14,16 @@ class Shoes
       attr_reader :app, :parent, :dimensions, :gui
 
       def initialize(app, parent, *args)
+        actual_initialize(app, parent, *args)
+      end
+
+      # Why is this separate? With Widgets, we need to invoke this base
+      # initialize, but ALSO a user-define initialize that historically didn't
+      # require calling super!
+      #
+      # Extracted like this, we can call actual_initialize to get base work
+      # done, and then initialize on the class ourselves to get the widget set.
+      def actual_initialize(app, parent, *args)
         blk    = args.pop if args.last.is_a?(Proc) || args.last.nil?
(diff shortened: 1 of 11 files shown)


jasonrclark:
Sample of the warning as it shows up (and the error I get when the widget's overriding `initialize` has the wrong number of arguments... other deeper errors are likely if you happen to take 3 arguments and then misuse them.)

```
~/source/shoes4:bin/shoes testing/widget.rb
WARN: You've defined an `initialize` method on class 'MyWidget'. This is no longer supported.
      Your widget likely won't display, and you'll potentially receive argument errors.

      Instead, define `initialize_widget` and we'll call it from your generated widget method.

ERROR: wrong number of arguments (3 for 2)
```

shoes-core/lib/shoes/common/ui_element.rb
```diff
@@ -14,6 +14,16 @@ class UIElement
...
+        actual_initialize(app, parent, *args)
+      end
+
+      # Why is this separate? With Widgets, we need to invoke this base
+      # initialize, but ALSO a user-define initialize that historically didn't
+      # require calling super!
+      #
+      # Extracted like this, we can call actual_initialize to get base work
+      # done, and then initialize on the class ourselves to get the widget set.
+      def actual_initialize(app, parent, *args)
```
PragTob: Do we still need this magic now?

shoes-core/lib/shoes/widget.rb
```diff
@@ -21,40 +26,47 @@ class Shoes
...
+    # subclasses have inappropriately overridden it or not.
+    def initialize(*_)
+      super
     end
 
-    # lookup a bit more complicated as during initialize we do
-    # not have @app yet...
-    def app
-      @app || self.class.app
+    def initialize_widget
```
PragTob: shouldn't this be `(*args)` ?

shoes-core/lib/shoes/widget.rb
```diff
@@ -64,5 +76,16 @@ def self.dsl_method_name(klass)
            .gsub(/([A-Z]+)([A-Z][a-z])/, '\1_\2')
            .gsub(/([a-z\d])([A-Z])/, '\1_\2').downcase
     end
+
+    def self.warn_about_initialize
+      if instance_method(:initialize).owner != Shoes::Widget
```
PragTob: nifty!

shoes-core/lib/shoes/common/ui_element.rb
jasonrclark: Nope, shouldn't be necessary

shoes-core/lib/shoes/widget.rb
jasonrclark: Actually I don't think so. If we made this `*args`, then widgets that don't define an `initialize_widget` would happily take any number of args to the widget method we generate, but then not do anything with them. Seems better that we accept none by default, and would give an argument count error if you tried to mis-call it. That make sense?

shoes-core/lib/shoes/common/ui_element.rb
jasonrclark: Updated!

shoes-core/lib/shoes/widget.rb
PragTob: makes sense yeah :)

PR #1477, opened 2017-09-27
Include samples with shoes gem

jasonrclark:
Fixes https://github.com/shoes/shoes4/issues/1320

Command format looks like a new subcommand:

```
shoes samples [-d DEST]
```

I only include the root samples and a few well-known subdirectories to support those. This excludes `unsupported` and the `packaging` apps as at least one of them has some huge images in it on purpose.

This PR also has a refactor to the command class hierarchy. We had an awkward split of some methods being class level (`help` and `options`) to support their use from elsewhere, while the main action is through an instance a the `run` method. Gotten rid of the class stuff, so now you just use nice, plain objects everywhere.

~~TODO: Haven't tested latest-and-greatest here with an actual built gem (did in the earlier going). Will want to kick the tires on that, but another night...~~ 💤


shoes.gemspec
```diff
@@ -13,7 +13,15 @@ Gem::Specification.new do |s|
...
 
-  s.files = Dir["LICENSE", "README.md", "lib/**/*", "ext/install/**/*"]
+  s.files = Dir[
+    "LICENSE", "README.md",
+    "lib/**/*",
+    "ext/install/**/*",
+    "samples/*",
+    "samples/lib/*",
+    "samples/potato_chopping/*",
+    "samples/sounds/*",
```
PragTob: huh, not too sure about packaging up those sounds, that's almost 4 MB. I mean, not huge but still seems a bit excessive starting at our nice 12.5kb

jasonrclark: Yeah, that's why I trimmed one of the looooong sounds down. Actually I've got an idea how we could easily exclude the sample that uses these from being packaged up but leave it for testing... stay tuned!

jasonrclark:
Pushed a change that moves the `simple_sounds.rb` into subdirectory with the sounds themselves, which we no longer package up. `samples/README` is graceful with subdirectories, even though we've never done it, so this keeps the sounds in the loop for dev testing without forcing them into the gem.

Gem size still up to 150K from 15K (images), but I can live with that.

PR #1482, opened 2017-10-02
Fix intermittent failure in integration_spec

jasonrclark:
Leftovers from app_spec.rb fails later on if you run this:

```
rspec   shoes-core/spec/shoes/app_spec.rb:500 shoes-swt/spec/shoes/swt/integration_spec.rb
```

Overall specs for swt failed on seed 46667 with this, which we hit once on an unrelated PR.


jasonrclark:
Merged so I can get other PRs to not fail so often from this

Issue #1224, opened 2016-02-29
Add #focussed? to retrieve whether or no an element has the focus


PragTob:
We should be able to ask elements whether or not they are currently focussed, this is especially true for our inputs (edit_line and button). Therefore a method `focussed?` seems fitting that just returns true/false. Could call it `focus?` or `has_focus?` but `focussed?` sounded best to my non native tongue ;)

Originated in this stackoverflow question: http://stackoverflow.com/questions/35415114/how-can-i-bind-function-with-object-in-ruby-shoes-gui

KCErb:
As long as we alias it with `focused` I'm in favor ;)

PragTob:
sure thing ;) (I believe we have gray and grey as well)

estensland:
Is this still wanted? I think I figured it out...

jasonrclark:
Absolutely @estensland, that'd be great!

jasonrclark:
Closed by #1489

PR #1489, opened 2017-10-09
issue 1224 added a #focused? method

estensland:
For Issue #1224 

I realized there are requests for aliases on the issue, this is essentially a question on the logic, then I will add aliases and other specs that we can think of, if the core of it makes sense.

I added the focused? method as a chain through focus modules just like the current #focus method, does this make sense?

### DIFF
--- a/shoes-core/spec/shoes/common/focus_spec.rb
+++ b/shoes-core/spec/shoes/common/focus_spec.rb
@@ -30,15 +30,15 @@ describe Shoes::Common::Focus do
     end
 
     it "has alias #focused" do
-      expect(subject.method(:focused) == subject.method(:focused?)).to be true
+      expect(subject.method(:focused)).to eq(subject.method(:focused?))
     end
 
     it "has alias #focussed" do
-      expect(subject.method(:focussed) == subject.method(:focused?)).to be true
+      expect(subject.method(:focussed)).to eq(subject.method(:focused?))
     end
 
     it "has alias #focussed" do
-      expect(subject.method(:focussed?) == subject.method(:focused?)).to be true
+      expect(subject.method(:focussed?)).to eq(subject.method(:focused?))
     end
   end
 end
--- a/shoes-core/spec/shoes/common/focus_spec.rb
+++ b/shoes-core/spec/shoes/common/focus_spec.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+require 'spec_helper'
+
+describe Shoes::Common::Focus do
+  class FocusTester
+    include Shoes::Common::Focus
+
+    attr_reader :gui
+
+    def initialize(gui)
+      @gui = gui
+    end
+  end
+
+  subject { FocusTester.new(gui) }
+
+  let(:gui) { double("gui", update_visibility: nil) }
+
+  describe "#focus" do
+    it 'must call #focus on gui' do
+      expect(gui).to receive(:focus)
+      subject.focus
+    end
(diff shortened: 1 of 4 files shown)


jasonrclark:
This looks good to me, thanks! (...and TIL that focused has alternate spellings outside the US. 📚 ✨)

@estensland feel free to add the aliases on if you'd like. I believe I also sent you an invite to team shoes and you'd be able to [merge on your own](https://github.com/shoes/shoes4/wiki/Merging-a-Pull-Request) if you're interested. If you'd rather not, I'm happy to click the big green button.

estensland:
@jasonrclark thanks I appreciate it!

I added some aliases and specs, I'd rather you take a look before merging to make sure I didn't miss an obvious one, and if you think the specs are worth it.

NOTE: I did not do #has_focus? because rubocop disapproved

shoes-swt/lib/shoes/swt/common/focus.rb
```diff
@@ -6,6 +6,14 @@ module Focus
...
           @real.set_focus
         end
+
+        def focused?
+          @real.has_focus
+        end
+
+        alias focussed? focused?
+        alias focussed focused?
+        alias focused focused?
```
jasonrclark: Looking it over, I'd actually be inclined to only provide the aliases on the shoes-core side, not on the GUI. I'm fine with the contract of the backends (which end-users never access directly) saying "you've got to use `focused?` as the primary one, while providing the convenience for Shoes apps themselves of the aliases. Mind making that change and then I'll get it merged?

estensland: Ah that makes sense, will change. Thanks!

estensland: The specs are for the gui and I don't see a spec for the shoes-core. Are the alias specs worth it enough?

jasonrclark: I'm fine with letting this stand if you're ready to be done with the issue.

That said, I've been slowly adding in specs for the various `common` modules that didn't get direct testing way back when they were created (or are tested indirectly through other element types). An example is `shoes-core/spec/shoes/common/visibility_spec.rb`.

I do think it's probably worthwhile to add at the shoes-core level, just to document the expected interaction between shoes-core and the backends, but it's not a huge problem to leave it like it is for now.

estensland: I'll give it a shot!

estensland:
I based the specs off visibility_spec, but let me know if a different formatting or style would be better

shoes-core/spec/shoes/common/focus_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  describe "#focused?" do
+    it 'must call #focused? on gui' do
+      expect(gui).to receive(:focused?)
+      subject.focused?
+    end
+
+    it "has alias #focused" do
+      expect(subject.method(:focused) == subject.method(:focused?)).to be true
```
jasonrclark: Could these be written more simply as `expect(subject.method(:focused)).to eq(subject.method(:focused?))`?

jasonrclark:
One last small comment, and I've re-kicked the Travis build. There's some intermittent test failure still happening that I haven't managed to track down, so the failure's nothing to do with your code.

jasonrclark:
Woot! Ready to ✨

Issue #853, opened 2014-08-14
Implement Slot Scrolling


jasonrclark:
Some of the basic structure was put in with #766, but not the guts of really implementing a scrolling view.

This will likely require quite a bit of changes regarding any classes that do their own painting (art shapes, text blocks).

Implementation of scrolling will help in addressing #255 and #519.

PR #1490, opened 2017-10-12
Factor parent visibility in

### LINKED ISSUE #905: Slots probably need a backend representation
Right now slots (stack/flow) just have a dummy backend representation without any effect - e.g. it doesn't really create an SWT element. We need a suitable SWT element that groups elements together but isn't actually displayed. 

Why do we probably need that?
- Scrolling #853
- Hiding all elements of a slot easily at once (without iterating through them and hiding each one of them, although that might not be the worst after all)

jasonrclark:
Fixes #905 (or at least what's gonna get fixed there)

When visibility changes happen, we use `update_visibility` to notify the
elements that might need to directly change their current visibility
state. This was only abiding by the direct visible styling for the
element, so when a parent container got hidden we had to hack around
with tromping visibility at the lower levels.

Instead we now take advantage of our `hidden?` method and have that factor
in the parent visibility. If direct visibility access is still needed,
it should go through the styling, but that's almost never what you want.

Note that this works without need a new backend element for the slots,
which is good because I'm convinced none of the SWT elements will
actually work since they don't support transparency. :sob: I'll put more about
where I'm at with that over on #853 in the near future.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -100,6 +100,8 @@ Metrics/BlockLength:
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
   Max: 250
+  Exclude:
+    - 'shoes-core/lib/shoes/slot.rb'
 
 # Offense count: 10
 Metrics/CyclomaticComplexity:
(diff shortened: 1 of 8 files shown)


jasonrclark:
Agreed, this one totally deserves a sample run-through. Will give that a kick before I merge so it's easier to spot if this has regressed.

jasonrclark:
Full run-through on samples and all looking clean visibility-wise. 🚢

Issue #1101, opened 2015-04-21
Gradient isn't applied correctly to shapes


PragTob:
The gradient seems to start somewhere in the middle of the shape in the middle. (that's sample 31)

![screenshot from 2015-04-21 07 45 47](https://cloud.githubusercontent.com/assets/606517/7246265/822c3e90-e7fa-11e4-9cb1-0afdc8aac5a9.png)

The code that does that is in SWT::Gradient, when playing around with it, adjusting the top value sort of helped but no idea why or how...

The fix might be in SWT::Gradient, it might also rest in the Shape class though, as maybe it reports its dimensions wrong for some reason.

Issue #1100, opened 2015-04-21
Gradient clipping with Line


jasonrclark:
With the following sample app:

```
Shoes.app do
  line 10, 10, 100, 100, strokewidth: 20, stroke: gradient(blue, green)
end
```

I see this result:

![image](https://cloud.githubusercontent.com/assets/130504/7245181/9e2211e2-e79b-11e4-9707-d6472c4154c8.png)

The strokewidth apparently should get taken into account when a line is being used with a gradient. Haven't been able to duplicate another element type having the same problem (even rectangles when rotated, which seemed most likely to be impacted).

PragTob:
I guess we could/should reuse some of the mathematical code of the `in_bounds?` implementation or treat it as a rectangle or something :)

jasonrclark:
The other idea I had was that we know based on the strokewidth the maximum that the line could extend below it's "bottom" point, so we could just set the gradient to start/stop slightly further out?

PragTob:
Certainly the more pragmatic solution :D 

I'd probably try an exact solution for a bit and if that doesn't work resort to that :)

jasonrclark:
Did a little further investigating, and seeing even more weirdness with the gradients. Here's the sample app which rotates a line around.

```
Shoes.app do
  @l = line 10, 200, 200, 200, strokewidth: 20, stroke: gradient(blue, green)
  dx = +8
  dy = -8

  animate do
    @l.left += dx
    @l.top  += dy

    dx = -dx if @l.left > 400 || @l.left < 10
    dy = -dy if @l.top  > 400 || @l.top < 10
  end
end
```

In addition to the clipping I noted before, when crossing the vertical point, the ends of the gradient switch. When crossing the horizontal, the pattern gets too small and repeats.

Still no precise solution--too much math for me this late at night--but more to consider.

jasonrclark:
Found this app in a commit comment, bc21f58124d37cd85dc1ffc8cadb3521a96eab46, which helped me finally understand what the whole angle business was doing:

```
Shoes.app height: 300 do
  background salmon..chartreuse, angle: -45
  oval 10, 10, 200, fill: green..yellow, angle: 90
  rect 300, 10, 200, fill: red..blue, curve: 10, angle: 45
end
```

This is also discussed at some length in #156. Not sure when I'll get to this, but I've got some ideas forming about what's going on and the best way to handle it.

KCErb:
[...]

PR #1491, opened 2017-10-12
Fix gradient angles across elements

### LINKED ISSUE #1354: Gradient angle incorrect across many art elements
Doesn't look like all elements are properly supporting the angle for gradient. Only the rectangle appears right to me.

With this application:

```
Shoes.app height: 800 do
  rect 100, 50, 100, 50, fill: blue..green, angle: 45
  oval 100, 150, 100, 50, fill: blue..green, angle: 45
  arrow 150, 250, 100, fill: blue..green, angle: 45
  line 100, 350, 200, 350, strokewidth: 50, stroke: blue..green, angle: 45
  star 100, 450, 5, center: true, fill: blue..green, angle: 45
  arc 100, 550, 100, 100, 0, Shoes::PI, fill: blue..green, angle: 45

  shape fill: yellow..red, angle: 45 do
    move_to 100, 650
    line_to 200, 650
    line_to 150, 750
    line_to 100, 650
  end
end
```

You get this result:

image

## TODO
* [ ] Also check with gradients on `stroke` instead of `fill`
* [x] rect
* [ ] oval
* [ ] arrow
* [ ] line (#1100)
* [ ] star
* [ ] arc
* [ ] shape (#1101)

jasonrclark:
Coulda probably busted things up, but was on a roll :)

Fixes #1354, #1100, and #1101  🎉 

There were a number of things wrong across the various element:

* Several were missing the DSL-level `angle` style they need to respect gradients
* Bounds on some different element types needed special handling, so provided a new `redraw_*` set of accessors that can be overridden properly
* `RedrawingAspect` also needs that same different redrawing boundaries used, so plumbed those in

Fortunately the base math seems to actually be right, we just weren't setting the "where" for the gradient correctly for a bunch of elements!

TODO:
* [x] Test with stroked gradients (been focused on fill)
* [x] Run samples for things that have gradients in them
* [x] Further specs for new functionality? While we exercise some bits based on the test updates needed, we aren't doing a lot of direct verification of how this stuff works

### DIFF
--- a/shoes-core/lib/shoes/arc.rb
+++ b/shoes-core/lib/shoes/arc.rb
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
 class Shoes
   class Arc < Common::ArtElement
-    style_with :angle1, :angle2, :art_styles, :center, :common_styles, :dimensions, :radius, :wedge
-    STYLES = { wedge: false, fill: Shoes::COLORS[:black] }.freeze
+    # angle is the gradient angle used across all art elements
+    # angle1/2 are the angles of the arc itself!
+    style_with :angle, :angle1, :angle2, :art_styles, :center, :common_styles, :dimensions, :radius, :wedge
+    STYLES = { angle: 0, wedge: false, fill: Shoes::COLORS[:black] }.freeze
 
     def create_dimensions(left, top, width, height, angle1, angle2)
       @style[:angle1] = angle1 || @style[:angle1] || 0
--- a/shoes-core/lib/shoes/arc.rb
+++ b/shoes-core/lib/shoes/arc.rb
@@ -3,8 +3,8 @@ class Shoes
   class Arc < Common::ArtElement
     # angle is the gradient angle used across all art elements
     # angle1/2 are the angles of the arc itself!
-    style_with :angle, :angle1, :angle2, :art_styles, :center, :common_styles, :dimensions, :radius, :wedge
-    STYLES = { angle: 0, wedge: false, fill: Shoes::COLORS[:black] }.freeze
+    style_with :angle1, :angle2, :art_styles, :center, :common_styles, :dimensions, :radius, :wedge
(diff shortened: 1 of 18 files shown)


jasonrclark:
Success! Doing a small bit of additional testing and will examine the PR in the new day to see how I feel about the need for more specs.

gradient

Sample code:
```
Shoes.app height: 800 do
  elements = []
  elements << rect(100, 50, 100, 80,   strokewidth: 5, stroke: purple..pink, fill: gradient(purple, orange))
  elements << oval(100, 150, 100, 50,  strokewidth: 5, stroke: purple..pink, fill: purple..orange)
  elements << arrow(300, 90, 100,      strokewidth: 5, stroke: purple..pink, fill: purple..orange)
  elements << star(100, 450, 5,        strokewidth: 5, stroke: purple..pink, fill: purple..orange, center: true)
  elements << line(100, 350, 200, 250, strokewidth: 50, stroke: purple..orange)
  elements << arc(100, 550, 100, 100, 0, Shoes::PI, fill: purple..orange, strokewidth: 5, stroke: purple..pink)

  elements << shape(fill: gradient(purple, orange), strokewidth: 5, stroke: purple..pink) do
    move_to 100, 650
    line_to 200, 650
    line_to 150, 750
    line_to 100, 650
  end

  keypress do
    @pause = !@pause
  end

  every 0.1 do
    unless @pause
      elements.each do |element|
        element.style(angle: (element.style[:angle] || 0) + 10)
      end
    end
  end
end
```

jasonrclark:
Pushed up final commits on this with two changes

1) fold `angle` into the default `art_styles` (because they all had it anyway), and default it appropriate at point of use so we avoid repetition and
2) adding some specs around particular art element redraw regions.

1 is something I would have wanted to write a bunch of tests around ("make sure all the art elements have this!") but much easier to just _make_ all the art elements have it :)

2 covers the other major chunk of functionality we're actually modifying here.

Will merge away a bit later once it's 🍏

PragTob:
:green_apple:

PR #1492, opened 2017-10-14
Auto Correct all RuboCop Offenses

ericboehs:
I've auto corrected the offenses in `.rubocop_todo.yml` via https://git.io/vd18a.


shoes-core/lib/shoes/common/ui_element.rb
```diff
@@ -41,8 +41,7 @@ def initialize(app, parent, *args)
       #
       # It is intended for performing any additions to the styles hash before
       # that gets sent on to style_init.
-      def before_initialize(_styles, *_)
-      end
+      def before_initialize(_styles, *_); end
```
PragTob: not personally a fan of this style gotta say, would still be ok with it but imo `;` shouldn't be in production code :)

.rubocop_todo.yml
```diff
@@ -99,9 +94,7 @@ Metrics/BlockLength:
 # Offense count: 16
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 250
-  Exclude:
-    - 'shoes-core/lib/shoes/slot.rb'
```
jasonrclark: Not sure why this got changed by the auto-corrector. I'm actually wanting to hold the line at 250 and specifically exempt that one file, so could this get changed back?

shoes-core/lib/shoes/common/ui_element.rb
jasonrclark: Also personally don't care for this. If we could revert this and potentially move `Style/EmptyMethod` with enabled false over to the main file, that'd be awesome. (If you don't want to make the move, I'm fine with doing that separately too.)

jasonrclark:
Applied these via #1511

Issue #79, opened 2012-06-21
Swt doesn't allow transparent backgrounds


wasnotrice:
I have been working on solving `#move` in #59 (see code at #78) and ran into this problem. Swt doesn't allow transparent backgrounds. I have been looking into this for a few days now, and can't seem to figure it out. So here are questions for you all:
1. Can we implement Shoes in a framework that doesn't provide transparent backgrounds?
2. If so, how?
3. What am I missing?

PR #1493, opened 2017-10-17
Set control visibility outside of slots + A PLAN!

### LINKED ISSUE #1036: Slots content should be hidden when the slot is too small to display them

jasonrclark:
Fixes #1036 by hiding elements which aren't painted (things like buttons and textboxes) when they're outside of slot boundaries. Painted elements all abide by the slot boundaries already.

This is a meaningful change on its own (even with the rough edges, which are as good as I suspect we can get on SWT), but it leads into to a broader discussion around scrolling (#853)

In particular, as I [discovered back in 2015](https://github.com/shoes/shoes4/issues/1036#issuecomment-162791833) (and then forgot) and others struggled with all the way back in 2012 (#79), SWT's `Composite` classes (which are the reasonable container objects for the backend) don't support transparency. That means they're unusable because we have to show the element layering for anything to work properly.

This leads me to conclude that the only way to get any sort of scrolling support outside the app frame (since the `Shell` can scroll fine for us) is to handle it ourselves. This hopefully won't be as bad as it sounds... the plan looks like this:

1) Get everything so it properly respects visibility when outside the slot frame
1) Make slots understand their scroll position and move contents around accordingly (much like `translate` already does)
1) See if we can get scrollbars actually displaying for our slots via `org.eclipse.swt.widgets.ScrollBar`
1) Implement necessary handlers/methods for accessing scroll information on slots
1) Yank the existing implementation which works against the app `Shell`. It has inconsistencies with how slot scrolling worked before, and once the other work is done we should be able to just scroll on the top slot.

This PR should stand on its own fairly well. ~~Deserves more specs and manual testing, which I'll get to, but~~ how's that plan sounds @PragTob? My only other thought is to patch up app-level scrolling and punt the whole business, but seems like a meaningful loss to basic functionality.

### DIFF
--- a/shoes-core/lib/shoes/dimensions.rb
+++ b/shoes-core/lib/shoes/dimensions.rb
@@ -82,6 +82,16 @@ class Shoes
       x_dimension.in_bounds?(x) && y_dimension.in_bounds?(y)
     end
 
+    def overlaps?(other)
+      return false unless other.element_left && other.element_top &&
+                          other.element_right && other.element_bottom
+
+      in_bounds?(other.element_left, other.element_top) ||
+        in_bounds?(other.element_left, other.element_bottom) ||
+        in_bounds?(other.element_right, other.element_top) ||
+        in_bounds?(other.element_right, other.element_bottom)
+    end
+
     def margin
       [margin_left, margin_top, margin_right, margin_bottom]
     end
--- a/shoes-core/lib/shoes/dimensions.rb
+++ b/shoes-core/lib/shoes/dimensions.rb
@@ -82,14 +82,14 @@ class Shoes
       x_dimension.in_bounds?(x) && y_dimension.in_bounds?(y)
     end
 
-    def overlaps?(other)
+    def contains?(other)
       return false unless other.element_left && other.element_top &&
                           other.element_right && other.element_bottom
 
-      in_bounds?(other.element_left, other.element_top) ||
-        in_bounds?(other.element_left, other.element_bottom) ||
-        in_bounds?(other.element_right, other.element_top) ||
-        in_bounds?(other.element_right, other.element_bottom)
+      element_left <= other.element_left &&
+        element_right >= other.element_right &&
+        element_top <= other.element_top &&
+        element_bottom >= other.element_bottom
     end
 
     def margin
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -95,6 +95,9 @@ Metrics/AbcSize:
 # Configuration parameters: CountComments, ExcludedMethods.
 Metrics/BlockLength:
   Max: 648
+  Exclude:
+    - 'shoes-core/spec/**/*.rb'
+    - 'shoes-swt/spec/**/*.rb'
 
 # Offense count: 16
 # Configuration parameters: CountComments.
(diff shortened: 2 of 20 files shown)


jasonrclark:
Wired up a little hard-coded scrolling with this to see what it would look like... I think that perhaps we need to hide non-painted elements as soon as they leave the slot boundaries, rather than waiting until they are completely outside. Evidence, as always best presented in gif form:

scrolling

jasonrclark:
Pushed change for visibility to work on leaving the parent bounds at all vs leaving it entirely, and the effect is much better:

scroll-more

Not the perfectly seamless thing I long for, but I'm willing to live with it.

jasonrclark:
Testing found that `samples/simple_button_animate.rb` seems to have an issue. The `displace` call that's moving the button up and down isn't taken into account when figuring the slot height apparently, so the button vanishes.

jasonrclark:
Applied that change, have run through the testing again. Done for the day, but will probably merge tomorrow unless I hear differently and head on with the scrolling plan!

shoes-core/lib/shoes/dimensions.rb
```diff
@@ -82,6 +82,16 @@ def in_bounds?(x, y)
       x_dimension.in_bounds?(x) && y_dimension.in_bounds?(y)
     end
 
+    def contains?(other)
+      return false unless other.element_left && other.element_top &&
+                          other.element_right && other.element_bottom
```
PragTob: shouldn't we always have these?

shoes-core/lib/shoes/dimensions.rb
```diff
@@ -82,6 +82,16 @@ def in_bounds?(x, y)
...
+
+      element_left <= other.element_left &&
+        element_right >= other.element_right &&
+        element_top <= other.element_top &&
+        element_bottom >= other.element_bottom
```
PragTob: partial contains aren't a it?

shoes-core/lib/shoes/dimensions.rb
jasonrclark: When we're in the midst of laying out the elements (i.e. contents alignment), they haven't necessarily been assigned yet. It's awkward, but definitely necessary with the current system. I've noticed it in other spots too and would like to revise, but it'd probably take some deep 🔪 on the slot placement which I don't have the brain-power for at the moment :)

shoes-core/lib/shoes/dimensions.rb
jasonrclark: I found that showing on partial containment looked way jankier than just hiding as soon as things got outside the box.

Should the method have a more precise name than `contains?`, like `fully_contains?`?

jasonrclark:
Have made some local progress on scrolling past this. I think it'll be valuable to just merge this and see how it's behaving in testing with the further scrolling work. We can revise it if we find the positioning/boundaries aren't quite right.

Issue #456, opened 2013-10-18
Get rid off warnings when running with -w


PragTob:
We should get rid off warnings when running with `ruby -w`. You can also run rspec with -w to figure those out.

Why do we want to get rid off those? Because it's the nice thing to do - lots of people develop their applications like that and if shoes causes errors then it disrupts their workflow.

Some of them are really unnecessary to fix imo (from a code perspective) but we should either way, cause we want to be nice and therefore run error free under -w. 

I think we can put a `.rspec` file with `-w` to ensure that it always runs like that.

It's not a high priority sort of issue, just something that we should keep in mind and do before rc1.

Cheers,
Tobi

PR #1496, opened 2017-10-18
Clean up warnings

alkalinecoffee:
For https://github.com/shoes/shoes4/issues/456

### DIFF
--- a/shoes-core/lib/shoes/app.rb
+++ b/shoes-core/lib/shoes/app.rb
@@ -101,7 +101,7 @@ class Shoes
     end
 
     def method_missing(name, *args, &blk)
-      if @__additional_context__
+      if defined?(@__additional_context__) && @__additional_context__
         @__additional_context__.public_send(name, *args, &blk)
       else
         super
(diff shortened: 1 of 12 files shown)


jasonrclark:
Thanks @alkalinecoffee! This looks  reasonable to me. Not a huge fan of the `defined?` pattern, but with our use of modules that rely on instance variables set elsewhere I'm not sure there's a better way to do things. Appreciate that in the concrete classes you've just set things to `nil`

Looks like you've got some Rubocop problems to clear up before we can get it merged. `bundle exec rake rubocop` will run things for you with the right settings.

jasonrclark:
Excellent, thank you @alkalinecoffee!

Quick question--what command were you running to find the particular warnings to fix? I'm seeing that `bundle exec rake spec:core` is clean on warnings, but `bundle exec rake spec:swt` isn't. (You'll get both if you just run `bundle exec rake spec`)

alkalinecoffee:
@jasonrclark Good question.  Not sure what I was running... it's possible it was just `spec:core`.  I'll create a new PR and clean up the rest.

PR #1502, opened 2017-10-29
Scrollbars!

### LINKED ISSUE #853: Implement Slot Scrolling

jasonrclark:
Fixes #853 at long last!

TODO
* [x] Sample run
* [x] Multi-OS testing (appears good on initial pokes, but deserves more attention)
* [x] Add sample with better scrolling slot?

### DIFF
--- a/samples/README
+++ b/samples/README
@@ -67,6 +67,7 @@ simple_random_bubbles.rb
 simple_require.rb
 simple_sample_executor_all.rb
 simple_sample_executor.rb
+simple_scroll.rb
 simple_sesame_street_shoes.rb
 simple_shoes_intro.rb
 simple_slide.rb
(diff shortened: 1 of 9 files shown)


jasonrclark:
Run through on samples (`good_vjot`) found that (at least on Mac) you can't actually scroll to the last 10 pixels on the scrollbar as set... making concession for that at the moment and testing on other OSs to see if we get similar behavior there.

jasonrclark:
Also added new sample that does scrolling. vjot and reminder do as well, but it's a bit more buried.

image

image

Issue #125, opened 2012-09-10
Implement Installer for Shoes.{exe,app,sh}


pjfitzgibbons:
This will be some form of build for Shoes Installer (a self-extracting zip/exe, DMG, Install4J, or other).

This installer will (probably) be only able to be run by Shoes Developers (this github codebase).
This installer would be targeted toward public availability on the "Downloads" page of http://shoesrb.com

PR #1510, opened 2017-11-05
Shoes launcher application

jasonrclark:
First steps along the way, but it's a start!

# What's It Do?

* Select a Ruby file and run it
* Packaging apps (see caveats below)
* Spawns manual

Someday we may want actual full installers per #125, but for initial release I'm planning to just go with packaged versions of this app for each OS.

## Initial view
image

## Packaging
image

image

image

image

# Packaging Woes
Packaging is the big caveat. I ran into major problems with getting warbler to function from within the package (yo dawg) that I'll be looking for further help on.

For now, this simply shells out to run `shoes package ...` for you. That doesn't work right from all packaged formats (Mac app doesn't have the right `PATH` set to find my rbenv installed JRuby executables for instance), but it gives a feel for what I'm aiming at.


jasonrclark:
Deeper thoughts occurring to me... we may not want to do this via the "just package an .rb from the repo" approach at all.

While the current warbler struggles I mentioned are an issue, I see deeper problems looming. Our launcher app needs to pick up potentially different gems from the user's packaging directory. That calls for running a separate process, but in our packaging we don't have JRuby around to just script that up. I can't see quite how we'd run that separate thing from within the package, lacking an actual Ruby installation to start it from.

I'm leaning toward making the Shoes app a separate repo which just drops JRuby alongside our code. This gives us more freedom for secondary scripts vs trying to force it all inside the package. The launcher would be shaped a lot like our packages, but without warbler in the picture. 

Don't know if this makes any sense to anyone who hasn't been staring at this for the last two days... short story maybe hold on this? Maybe roll with this for now, under the plan that will roll it back once a real app's ready? Thoughts?

jasonrclark:
Pushed an updated version that takes the approach I alluded to of bundling up JRuby alongside our code to form the app. This has made the packaging _much_ simpler, and I think will also deal with issues we might have had around running apps with Gemfiles in the original approach (although need more code and testing for that)

`app/build-mac` will create an example of this... working on versions of this for Windows and Linux next. I've uploaded a version of the generated app (and will for the others) on [this release](https://github.com/shoes/shoes4/releases/tag/shoes-app-4.0.0.rc1).

jasonrclark:
`app/build-windows` appears, along with a binary on the [Shoes app release tag](https://github.com/shoes/shoes4/releases/tag/shoes-app-4.0.0.rc1)! Perhaps Linux tomorrow...

jasonrclark:
And then there was `app/build-linux`...

What remains? A pile of testing against each of the OSs. I think there's almost certainly a bug right now just in the plain app when trying to run in a directory with an alternative Gemfile (it should shell out instead like we do when packaging). Fix up Rubocop crankiness, make sure tests pass... all that good stuff.

PragTob:
Looks great :heart_eyes:

jasonrclark:
Final problems to resolve from cross-OS testing:

* [ ] Mac doesn't pass the thread access flag on child processes, so crashes
* [ ] Cancel not handled on file opens, so crashing there too

jasonrclark:
Testing's finally coming up clean! Assuming CI passes, will give this another day or so in case anyone wants to chime in, but this is ready to 🚢

PR #1515, opened 2017-11-13
Issue #778 basic oval formula for area

estensland:
For #778 

@jasonrclark this appears to work, but wanted to open up PR to ask if I am missing something obvious. I am not as experienced with the whole app, I did tried rotating and the bound still seem to work.

If you can't think of a use case I am missing, I'll write some specs and think about refactoring code.


jasonrclark:
This looks great from my testing, thanks @estensland!

PR #1518, opened 2017-11-24
Rubocop: Style/SafeNavigation

jaredmoody:


shoes-core/lib/shoes/color/hex_converter.rb
```diff
@@ -26,8 +26,7 @@ def pad_if_necessary(hex)
       # Returns a 3- or 6-char hex string for valid input, or nil
       # for invalid input.
       def validate(hex)
-        match = /^#?(([0-9a-f]{3}){1,2})$/i.match(hex)
-        match && match[1]
+        /^#?(([0-9a-f]{3}){1,2})$/i.match(hex)&.[](1)
```
jaredmoody: This is the only line of this PR I don't like  - explicitly calling `[]` seems weird and perhaps unclear that it's equivalent

Alternatively, we could disable this cop on this line

jasonrclark: Agreed, that makes an already dense snippet even more obscure.

What's the cop say if we try to use a post-fix if instead?

```
       match = /^#?(([0-9a-f]{3}){1,2})$/i.match(hex)
       match[1] if match
```

PragTob: Could also move the regex into a constant imo :)

however yes, `&.[](1)` doesn't seem nice

jaredmoody:
Rubocop still complained on the Post-fix - updated with the regex pulled out into a constant and the cop disabled inline.

PragTob:
Merge conflict now and soemhow a test failed but travis doesn't load for me

jasonrclark:
Looking 🍏  now. Thanks!

PR #1519, opened 2017-11-24
Rubocop: Style/SymbolArray ?

jaredmoody:
I’m unsure about if I like this syntax, what do you guys think?


jasonrclark:
Thanks for pulling this out into a single PR to discuss @jaredmoody !

Personally I find the `%i` more obscure. @PragTob any thoughts? I'd vote to just move it to disabled in the main file vs apply this particular change.

PragTob:
:+1: on disabled, I can live with both styles - agreed though that `%i` is more magical

jaredmoody:
Updated rubocop config to disable this cop

PR #1520, opened 2017-11-25
add fail fast wrapper to link:click

cursande:
Mainly I'm looking to work out a satisfactory way to test it, and then work out the right approach from there.


jasonrclark:
Hey @cursande! I'm glad you put up the PR, because it looks like your additions are around the registration for clicks (i.e. when someone assigns the block to the element), rather than the actual evaluation when a click happens.

The class where the main action happens is [`Shoes::Swt::ClickListener`](https://github.com/shoes/shoes4/blob/master/shoes-swt/lib/shoes/swt/click_listener.rb#L41-L58). Upside here is that I think applying your changes there will get not only links, but other text and potentially art elements too!

As you suggest, starting with tests is a good idea, and those should probably fit in [around here](https://github.com/shoes/shoes4/blob/master/shoes-swt/spec/shoes/swt/click_listener_spec.rb#L70). In particular, you'll be looking to add specs that flip the fail-fast on and off, then then stub the `block` to raise when called and confirm the behavior. You could probably draw some good inspiration from [the Hover specs around failing](https://github.com/shoes/shoes4/blob/master/shoes-core/spec/shoes/common/hover_spec.rb#L55-L93)

Hope that helps, and do let me know if there's any more I can do to get you moving.

cursande:
Thanks a lot for your feedback, makes much more sense now. I'll get on to fixing it up

PragTob:
? the diff here is 2 lines but 11 commites and no fail safe wrapper? Did we accidentally implement this double? :see_no_evil:

jasonrclark:
I think @cursande had taken a direction originally which didn't quite pan out, and where the branch is sitting was basically after reverting most of the changes... might be easier to restart this from master if they want to continue with it.

cursande:
Sorry, this is my issue. I had removed the changes I had made because they were redundant, and was going to start a fresh pull request without all the extra commits. In any case I've actually been waiting for an opportunity to meet up with a mentor in the new year to go through a few things around testing, including this. If it's not a matter of urgency I do intend to add this and other fixes, otherwise apologies for the lack of movement on this

PragTob:
sounds good, I'll just close this then so it doesn't seem like we have loads of open PRs that we don't take care of. New PR more than welcome :)

PR #1521, opened 2017-11-25
Rubocop: Style/FormatStringToken

jaredmoody:


PragTob:
this is still weird to me but I think it's good :D

PR #1523, opened 2017-11-28
Default to GTK 2 still

### LINKED ISSUE #1506: Contents off all apps escape the window: elementary OS 0.4.1
image

jasonrclark:
Fixes #1506

Continuing to see problems on GTK 3 for some Linux distros. Haven't dug into it yet.

Takes advice we made in https://github.com/shoes/shoes4/pull/880#issuecomment-55183970.


PragTob:
but... :'(

maybe we could try upgrading SWT? I know it probably won't work/help... might be worth a shot now or in the future though.

jasonrclark:
Fair enough. I've entered that separately to track at #1524, and will see get to it eventually, although I don't know whether I'll be able to have enough testing time before I'd like to get 4.0 stamped out.

I share your lack of expectation about it @PragTob--it has rarely helped our issues in the past--but it'll be worth getting up to date for the longer-term health of the project.

Issue #996, opened 2014-12-25
Give star and arc a center point accessor


jasonrclark:
With the changes in Shoes 4 to treat `left`/`top` as actually `left`/`top`, it would be convenient to have a center point accessor.

Note that we'll not want to call it center, to avoid colliding with the existing `center` option on dimensions.

Also will need to ensure it works bi-directionally, so changes to `left`/`top` will reflect in the center point, and changes to the individual values of the center point should modify `left`/`top`.

PR #1525, opened 2017-11-29
Star Center Point

nicholalexander:
A first pass at issue #996 - Give Star and arc a center point accessor.

Thought I'd start with a simple method to get the center point.  Then could work on setting the center_point and updating the corresponding Top and Left coordinates.  Then could recreate with arcs under separate PR.

## Open Questions
1. Not even sure this is the right way to add the functionality or if Shoes was looking for a style to use in the same way as `style_with`?
2. Should it return a Point? <- Seems like no, given 3.
3. I added the sample app mostly for my own benefit - can easily delete!
4. When setting a point, do you want to pass in a Point or just use simple x and y's? <- seems like this would be more in line with how coordinates are used in Shoes.
5. Wasn't fully sure about the difference between element_width and width...

Thanks!

### DIFF
--- a/shoes-core/spec/shoes/star_spec.rb
+++ b/shoes-core/spec/shoes/star_spec.rb
@@ -207,13 +207,11 @@ describe Shoes::Star do
     subject { Shoes::Star.new(app, parent, 100, 100, 5, 50, 30, center: false, strokewidth: 0) }
 
     it "should return the center point of the star" do
-      expect(subject.center_point.to_a).to eq([150,150])
+      expect(subject.center_point.to_a).to eq([150, 150])
     end
 
     it "should return the coordinates as a Point" do
       expect(subject.center_point.class).to eq(Shoes::Point)
     end
-
   end
-
 end
--- a/shoes-core/spec/shoes/star_spec.rb
+++ b/shoes-core/spec/shoes/star_spec.rb
@@ -209,5 +209,10 @@ describe Shoes::Star do
     it "should return the center point of the star" do
       expect(subject.center_point).to eq(Shoes::Point.new(175, 225))
     end
+
+    it "should handle stars initialized with nil dimensions" do
+      nil_star = Shoes::Star.new(app, parent, 100, nil, 0, nil, nil, center: false, strokewidth: 0)
+      expect(nil_star.center_point).to eq(Shoes::Point.new(200, 100))
+    end
   end
 end
--- a/shoes-core/spec/shoes/star_spec.rb
+++ b/shoes-core/spec/shoes/star_spec.rb
@@ -215,4 +215,13 @@ describe Shoes::Star do
       expect(nil_star.center_point).to eq(Shoes::Point.new(200, 100))
     end
   end
+
+  describe "center_point=" do
+    subject { Shoes::Star.new(app, parent, 125, 175, 5, 50, 30, center: false, strokewidth: 10) }
+
+    it "should set a new center_point" do
+      subject.center_point = Shoes::Point.new(80, 90)
+      expect(subject.center_point).to eq(Shoes::Point.new(80, 90))
+    end
+  end
 end
--- a/shoes-core/spec/shoes/star_spec.rb
+++ b/shoes-core/spec/shoes/star_spec.rb
@@ -204,14 +204,10 @@ describe Shoes::Star do
   end
 
   describe "center_point" do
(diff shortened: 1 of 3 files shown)


jasonrclark:
Welcome @nicholalexander! Thanks for the PR.

1. This seems like a fine approach to just use a method instead of `style_with`. Another example where we have `Point` based access is [over in `Line`](https://github.com/shoes/shoes4/blob/master/shoes-core/lib/shoes/line.rb#L7), and in that case it isn't surfaced through styles, just methods so 👌 
2. Actually, given the example we have in `Line` for points, I'm happy to return `Point` objects for this new method, since it's not there for compatibility
3. Adding a sample seems like a great idea!
4. So the `Line` example finally breaks down here... it doesn't allow setting via the point attribute at all. Give that, I'd be inclined to allow `center_point=` to take a `Point`, and its implementation would set the underlying dimensions.
5. There's a detailed explanation of the element_width vs width properties [in this file](https://github.com/shoes/shoes4/blob/master/shoes-core/lib/shoes/dimensions.rb) which may help. Basically, the differences revolve around whether values factor in the parent positions and margins. I'd recommend applying some margins in your sample and seeing (maybe with a quick `click` handler?) whether your center point still comes up right when clicking.

Hope that helps, and good plan to just do star here and handle arc separately. Thanks again!

shoes-core/spec/shoes/star_spec.rb
```diff
@@ -202,4 +202,16 @@
       expect(subject.redraw_height).to eq(108)
     end
   end
+
+  describe "center_point" do
+    subject { Shoes::Star.new(app, parent, 100, 100, 5, 50, 30, center: false, strokewidth: 0) }
+
+    it "should return the center point of the star" do
+      expect(subject.center_point.to_a).to eq([150, 150])
```
jasonrclark: Any particular reason to not just `eq(Shoes::Point.new(150, 150))`?

Also minor point, but you might want to alter dimensions so the center x and y don't match just in case the code got them flopped.

shoes-core/spec/shoes/star_spec.rb
```diff
@@ -202,4 +202,16 @@
...
+    end
+
+    it "should return the coordinates as a Point" do
+      expect(subject.center_point.class).to eq(Shoes::Point)
```
jasonrclark: If we do the earlier comparison via `Shoes::Point`, then this check might become unnecessary.

shoes-core/spec/shoes/star_spec.rb
```diff
@@ -202,4 +202,16 @@
       expect(subject.redraw_height).to eq(108)
     end
   end
+
+  describe "center_point" do
+    subject { Shoes::Star.new(app, parent, 100, 100, 5, 50, 30, center: false, strokewidth: 0) }
```
jasonrclark: One other test that might be worth adding is explicitly setting some of the dimensions to `nil` values. This mimics some scenarios during app startup where we might not be fully positioned yet. I typically make derived values like your `center_point` just do something sensible (`nil` or `0` return) if their underlying values are missing.

shoes-core/spec/shoes/star_spec.rb
nicholalexander: No reason at all except my general lack of paying attention.  Thanks for pointing out!  And also, totally good idea about the different coordinates.  Will Fix.

shoes-core/spec/shoes/star_spec.rb
nicholalexander: Agree

shoes-core/spec/shoes/star_spec.rb
nicholalexander: Ok, cool!  Thanks.

nicholalexander:
Great - thanks for the feedback.  Will make changes and proceed based on above feedback.

jasonrclark:
Excellent! Really like the sample changes too--they look nice, and do a great job of showing off how the `center_point` functionality works!

Question, did you see any trailing redrawing effects when running the sample? This isn't due to your code, but I get some single pixels "left behind" when the star moves:

image

I'm entering a separate issue since it isn't related to your changes, but curious if you saw it or it's something local for me.

nicholalexander:
Thanks!  Comments on the star dust on #1526

Issue #620, opened 2014-03-26
Document ALL the things


PragTob:
As we want to be a nice easy to use toolkit we need documentation. Lots of it.
- [ ] Write rdoc for classes/public methods
- [ ] Appropriately mark private methods as such so they are not included in the rdoc
- [ ] Update the manual
- [x] Ensure Yard generation to http://www.rubydoc.info/github/shoes/shoes4 is looking 👌 

This is mostly here to remind us of it for ~rc1~ 4.0. Some methods already have rdoc comments which is great :+1: 

The most important thing to document is the DSL module.

As a side note, I'm on the pro side of code should be so readable that it does not need comments. Documentation, especially for a library is different though. Especially a beginner library. Lots of people out there wouldn't even dare to look at source code to see how something works - I was just like that when I started and years after that. And people shouldn't have to do that. Good up to date documentation is vital to the success of a project imo.

ybur-yug:
@PragTob I would love to start working on this, and also commented w.r.t Rubocop violations. I am working on a project for a nonprofit that needs cross-platform support and this looks more promising than electron for my needs (quite simple). So, since I'll hopefully be using shoes4 (so far have used 3.x, but want to upgrade) I would love to start contributing.

jasonrclark:
:heart: @ybur-yug that sounds great! Let us know if you need any help in getting started on either of those. They both seem like great points to start from.

FWIW, the manual has been extracted to it's own separate project, so if you're wanting to modify things in that you'll want to PR over there. https://github.com/shoes/shoes-manual

ybur-yug:
Thanks @jasonrclark! I've got a bit of a busy early week w.r.t my regular work, but should be able to dive in relatively soon. I was thinking of starting with Rubocop because it will force me to go through the guts of the codebase (which is obviously quite large). With the other simple 'newcomer friendly' pieces like tests on somewhat isolated classes would get a lot easier with that done.

Thoughts? I'm pretty open to being instructed, this is obviously yall's baby :)

jasonrclark:
You're absolutely right that Rubocop will get you some broad exposure to the code, and get you running tests and learning where things stitch together.

I'd recommend finding a set of changes--either fixing the same problem across the library, or fixing all the problems in a single class--and then PR'ing that individually. Smaller changesets will be easier to get reviewed, tested and merged.

Shoes is a labor of love for all of us involved, so no worries on timing. We all just do what we can when we can, and are excited for every contribution!

[...]

PR #1527, opened 2017-12-01
Arc Center Point

nicholalexander:
following up on 2nd half of #996 

Same implementation, but notice the commented out test for centered arcs.  This doesn't work, and I'm not sure what the expected behavior of it should be.  It is also a problem on stars.  I wasn't sure if I should file this and then open a bug or no?  

Also, I tried looking at where centering gets applied but have not yet been successful in understanding how those dimensions actually get created - except within the SWT code?

### DIFF
--- a/shoes-core/lib/shoes/arc.rb
+++ b/shoes-core/lib/shoes/arc.rb
@@ -22,5 +22,16 @@ class Shoes
     def wedge?
       wedge
     end
+
+    def center_point
+      center_x = left + (element_width * 0.5).to_i
+      center_y = top + (element_height * 0.5).to_i
+      Point.new(center_x, center_y)
+    end
+
+    def center_point=(point)
+      self.left = point.x - (width * 0.5).to_i
+      self.top = point.y - (height * 0.5).to_i
+    end
   end
 end
--- a/shoes-core/lib/shoes/arc.rb
+++ b/shoes-core/lib/shoes/arc.rb
@@ -30,8 +30,13 @@ class Shoes
     end
 
     def center_point=(point)
-      self.left = point.x - (width * 0.5).to_i
-      self.top = point.y - (height * 0.5).to_i
+      if style[:center]
+        self.left = point.x
+        self.top = point.y
+      else
+        self.left = point.x - (width * 0.5).to_i
+        self.top = point.y - (height * 0.5).to_i
+      end
     end
   end
 end
--- a/samples/simple_rainbow.rb
+++ b/samples/simple_rainbow.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+# a simple shoes app to show how to use the star.center_point method
+Shoes.app width: 600, height: 600 do
+  # Roy G. Biv!
+  the_colors_of_the_rainbow = {
+    red: Shoes::Color.new(255, 0, 0),
+    orange: Shoes::Color.new(255, 127, 0),
+    yellow: Shoes::Color.new(255, 255, 0),
+    green: Shoes::Color.new(0, 255, 0),
+    blue: Shoes::Color.new(0, 0, 255),
+    indigo: Shoes::Color.new(75, 0, 130),
+    violet: Shoes::Color.new(148, 0, 211)
+  }
+
+  # we will set some variables to describe the first band of our rainbow,
+  # the red one.
+  band_thickness = 10
+  draw_left = 100
+  draw_top = 200
+  band_width = 400
+  band_height = 400
+
+  # tell Shoes how we're going to draw
+  nofill
+  strokewidth band_thickness
+
+  # draw a band for each of the colors of the rainbow.
+  the_colors_of_the_rainbow.each do |_color_name, color_rgb|
+    stroke(color_rgb)
+    arc draw_left, draw_top, band_width, band_height, Shoes::PI, 0
+
+    # update variables for next run.  the left and top point increase because
+    # we're drawing further in and down and the width decreases because we're
+    # because each successive band has to be contained within the one prior.
+    draw_left += band_thickness
+    draw_top += band_thickness
+    band_width -= (band_thickness * 2)
+    band_height -= (band_thickness * 2)
+  end
+
+  # add to docs
+  # 0 is 3 o'clock
+  # unit of angle expected in arguments is RADIANS!  DAWG!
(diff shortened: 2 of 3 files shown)


jasonrclark:
Ah, that's a good catch on centering. The centering gets applied down at the [`Shoes::Dimensions`](https://github.com/shoes/shoes4/blob/master/shoes-core/lib/shoes/dimensions.rb#L146) level (which is the class that the various elements use to track their actual dimensions).

It is likely that you'll just need a conditional around whether you're centered or not in your accessors--[that's effectively what the core dimensions do](https://github.com/shoes/shoes4/blob/71d5ec76f5be60cfdae08b80020171b5b0aee9a3/shoes-core/lib/shoes/dimension.rb#L62).

If you want to address the centering with the Arc code here, that'd be fine. In either case, an issue documenting the problem for Star (and if you don't want to do it here, Arc) would be appreciated!

jasonrclark:
Another thought as well... if the implementation of this center point is precisely the same, I wonder if it could be extracted to a `Shoes::Common` module, or implemented on the `Shoes::ArtElement`? Then when we have to do funky things about centering for instance, it'd be all in one place to fix (centralized... hahaha).

nicholalexander:
@jasonrclark - thanks!  I fixed the problem with arc here and opened another issue for the star.  If this fix is right, I'll go back and apply to star.  Tests passing locally.

Also, love the idea of centralizing all centerable things - let me get this and the star working and then i can open a new PR to refactor the methods into the common module - does that sound reasonable?

shoes-core/lib/shoes/arc.rb
```diff
@@ -22,5 +22,21 @@ def create_dimensions(left, top, width, height, angle1, angle2)
     def wedge?
       wedge
     end
+
+    def center_point
+      center_x = left + (element_width * 0.5).to_i
+      center_y = top + (element_height * 0.5).to_i
```
jasonrclark: Does the read side of this accessor need to factor in whether we're centered or not as well?

jasonrclark:
Had a question on the current implementation, but the plan to 1) get this merged, 2) apply it to star 3) refactor out to common module sounds 👌  to me.

shoes-core/lib/shoes/arc.rb
nicholalexander: Actually, I'm not sure.  I think this logic is behaving as we would want.  Here's a sample pair of shoes I was running around in and the center_point accessor match the x and y cursor - which I think is what it should be doing?

```
Shoes.app width: 420, height: 420, resizable: false do
  stroke rgb(0, 0.6, 0.9)
  cap :curve
  nofill
  strokewidth 10
  
  @arc = arc 210, 300,
             200, 200,
             -(Shoes::HALF_PI / 3),  -(Shoes::HALF_PI / 3) + Shoes::TWO_PI - Shoes::HALF_PI,
             center: true

  motion do |x, y|
    @arc.center_point = Shoes::Point.new(x,y)
    puts "cursor position: x: #{x} y: #{y}\r"
    puts "arc center_point: x: #{@arc.center_point.x} y: #{@arc.center_point.y}\r"
  end
end
```

I think this is because the `style[:center] = true` is correctly manipulating the shape, but it was only when i was directly setting the top and left I had to account for the center style.

Also, in looking at the test for the centered arc, it seems like that would catch it if this behavior changed in the future.

Sorry, does any of this seem right?  Also, do you think I should include the above sample in this PR?

jasonrclark: So here's a sample app that hopefully clarifies where I think there's an issue:

```
Shoes.app do
  @s1 = star 100, 100, 5, 90, 50, fill: orange
  @s2 = star 100, 100, 5, 90, 50, fill: pink, center: true

  stack do
    @info = para ""
    para @s1.center_point
    para @s2.center_point
  end

  click do |*args|
    @info.text = args.inspect
  end
end
```

In this sample, these two stars get set with the same location x, y values at creation, but one's `center: true` and the other isn't. These obviously draw in different spots on the screen.

The `para`'s at the top report the `center_point`, and with the current logic they both report the same value even though they should be different. Does that make sense?

Would love a nice centered arc example. `samples/simple_arc.rb` is already taken, but feel free to add one!

nicholalexander: Cool, I totally see that.  But with arc's I think it seems to work.  I added to your program and I think they are correctly reporting their center points.  Since we have #1531 opened, and since I wasn't sure about refactoring the center_point into the common module, I figured I'd address the stars under a separate PR meant to address 1532.  If you think that still makes sense.

```
Shoes.app do
  @s1 = star 100, 100, 5, 90, 50, fill: orange
  @s2 = star 100, 100, 5, 90, 50, fill: pink, center: true

  @a1 = arc 200, 200, 200, 200, 3.141592654, 0, fill: blue
  @a2 = arc 100, 300, 200, 200, 3.141592654, 0, fill: red, center: true
  
  stack do
    @info = para ""
    para @s1.center_point
    para @s2.center_point
    para @a1.center_point
    para @a2.center_point
  end

  click do |*args|
    @info.text = args.inspect
  end
end
```

Also, I added a sample program under simple_rainbow.rb - a nice happy rainbow!  I had a question about the documentation for shoes4 and I'm not sure where the right place to ask that is?  

Thanks for all your feedback!  I hope I'm not actually creating more work for you!  Let me know what you think about the arc's working and if I can fix the Star accessor in that separate PR.

jasonrclark: Oh, that's super interesting! I thought the center point vs left/top thing was handled by underlying code, so I'm surprised that they're different between the two.

We can totally carry on and address the star issues with #1531 since arcs aren't impacted.

And no worries at all about creating work. I love helping folks get involved and get to know the code, and you're making good changes that it's fun to follow along with. Shoes on!

jasonrclark:
So with my question about `center_point` out of the way, is this ready to roll? Love the new sample BTW 🌈

nicholalexander:
Great!  I really liked making the rainbow so I'm glad its a good sample!  And yes, ready to roll on this PR, thanks!  

Also just as a follow up on the documentation question, is there an open active issue or pr for that or should i start a new one?

jasonrclark:
Cool. Will get that merged.

I seem to have missed the doc question. If you’re interested in general documentation we do have https://github.com/shoes/shoes4/issues/620 our there, along with a few other specific tickets for particular things that need docs.

PR #1529, opened 2017-12-03
Rubocop: Lint/IneffectiveAccessModifier

jaredmoody:


PragTob:
I think the auto fix here fails hard as we might be relying on these class methods not being private, travis won't load the logs tough so I'm not sure :)

shoes-core/lib/shoes/app.rb
```diff
@@ -149,5 +130,26 @@ def self.new_dsl_method(name, &blk)
       define_method name, blk
       @method_subscribers.each { |klazz| klazz.def_delegator :app, name }
     end
+
+    private_class_method :subscribe_to_dsl_methods, :new_dsl_method
```
jasonrclark: I think the test problems are from this line--these are both explicitly called elsewhere, so they shouldn't be private. [`subscribe_to_dsl_methods` use](https://github.com/shoes/shoes4/blob/71d5ec76f5be60cfdae08b80020171b5b0aee9a3/shoes-core/lib/shoes/widget.rb#L31), [`new_dsl_method` use](https://github.com/shoes/shoes4/blob/71d5ec76f5be60cfdae08b80020171b5b0aee9a3/shoes-core/lib/shoes/widget.rb#L54)

jaredmoody: OK cool, that's exactly what this lint was supposed to pick up - these previous methods were under `private` but effectively public.  I made them _actually_ private and this failed so I will instead make them explicitly public.

jasonrclark:
:+1:

PR #1532, opened 2017-12-10
Docs for Arc and Working Rake Task

nicholalexander:
Trying to also make progress against #620.

Two things here: 
1. when I ran `rake yard` the yardoc generation didn't generate any files.  Maybe this was because I wasn't running it correctly?  I was able to generate some of the documentation by modifying the rake task as here.  However, we may want to add a .yardopts file so that running yardoc from the root works without requiring a rake task?  This might also be required to get the docs on rubydoc - entirely not sure.  I also added --no-private flag as per the comments in #620.

2. Made a first pass at adding in some docs for the Arc class where I feel sufficiently able to explain what's going on.  I also moved a comment into the notes so that it is featured in the generated docs.

Completely not sure if this is helpful or not but thought I'd start on it and welcome any and all direction change!

best,
n

### DIFF
--- a/shoes-core/lib/shoes/arc.rb
+++ b/shoes-core/lib/shoes/arc.rb
@@ -9,7 +9,6 @@ class Shoes
   #   arc 200, 200, 100, 100, Shoes::PI, 0
   # @note Angle is the gradient angle used across all art elements. Angle1/2 are the angles of
   #   the arc itself!
-  # @author Jason Clark
   # @param left [Integer] The number of pixels from the left side of the window.
   # @param top [Integer] The number of pixels from the top side of the window
   # @param width [Integer] The width of the arc element.
--- a/shoes-core/lib/shoes/arc.rb
+++ b/shoes-core/lib/shoes/arc.rb
@@ -3,6 +3,13 @@
 class Shoes
   # Arc.  A basic element representing a curve of a circle or an oval.
   #
+  # @example A simple arc which describes the bottom half of a circle and uses the centered style.
+  #   arc 200, 200, 100, 100, 0, Shoes::PI, center: true
+  # @example An arc which describes the top half of a circle.
+  #   arc 200, 200, 100, 100, Shoes::PI, 0
+  # @note Angle is the gradient angle used across all art elements. Angle1/2 are the angles of 
+  #   the arc itself!
+  # @author Jason Clark
   # @param left [Integer] The number of pixels from the left side of the window.
   # @param top [Integer] The number of pixels from the top side of the window
   # @param width [Integer] The width of the arc element.
@@ -11,14 +18,7 @@ class Shoes
   #   starting from the 3 o'clock position.
   # @param angle2 [Float] The second angle of the arc to the center point, in Radians.
   # @param [Hash] Style hash
-  # @example A simple arc which describes the bottom half of a circle.
-  #   arc 200, 200, 100, 100, 0, Shoes::PI, center: true
-  # @example An arc which describes the top half of a circle.
-  #   arc 200, 200, 100, 100, Shoes::PI, 0
-  # @author Jason Clark
   class Arc < Common::ArtElement
-    # angle is the gradient angle used across all art elements
-    # angle1/2 are the angles of the arc itself!
     style_with :angle1, :angle2, :art_styles, :center, :common_styles, :dimensions, :radius, :wedge
     STYLES = { wedge: false, fill: Shoes::COLORS[:black] }.freeze
 
--- a/shoes-core/lib/shoes/arc.rb
+++ b/shoes-core/lib/shoes/arc.rb
@@ -7,7 +7,7 @@ class Shoes
   #   arc 200, 200, 100, 100, 0, Shoes::PI, center: true
   # @example An arc which describes the top half of a circle.
   #   arc 200, 200, 100, 100, Shoes::PI, 0
-  # @note Angle is the gradient angle used across all art elements. Angle1/2 are the angles of 
+  # @note Angle is the gradient angle used across all art elements. Angle1/2 are the angles of
   #   the arc itself!
   # @author Jason Clark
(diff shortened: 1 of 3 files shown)


shoes-core/lib/shoes/arc.rb
```diff
@@ -1,9 +1,24 @@
...
 class Shoes
+  # Arc.  A basic element representing a curve of a circle or an oval.
+  #
+  # @example A simple arc which describes the bottom half of a circle and uses the centered style.
+  #   arc 200, 200, 100, 100, 0, Shoes::PI, center: true
+  # @example An arc which describes the top half of a circle.
+  #   arc 200, 200, 100, 100, Shoes::PI, 0
+  # @note Angle is the gradient angle used across all art elements. Angle1/2 are the angles of
+  #   the arc itself!
+  # @author Jason Clark
```
jasonrclark: Don't know that I deserve specific credit here--I probably just shuffled the class around from it's original location :)

shoes-core/lib/shoes/arc.rb
```diff
@@ -1,9 +1,24 @@
 # frozen_string_literal: true
 
 class Shoes
+  # Arc.  A basic element representing a curve of a circle or an oval.
+  #
+  # @example A simple arc which describes the bottom half of a circle and uses the centered style.
+  #   arc 200, 200, 100, 100, 0, Shoes::PI, center: true
+  # @example An arc which describes the top half of a circle.
+  #   arc 200, 200, 100, 100, Shoes::PI, 0
```
jasonrclark: Love the examples, very helpful!

shoes-core/lib/shoes/arc.rb
```diff
@@ -1,9 +1,24 @@
...
+  #   the arc itself!
+  # @author Jason Clark
+  # @param left [Integer] The number of pixels from the left side of the window.
+  # @param top [Integer] The number of pixels from the top side of the window
+  # @param width [Integer] The width of the arc element.
+  # @param height [Integer] The height of the arc element.
+  # @param angle1 [Float] The first angle of the arc to the center point, in Radians,
+  #   starting from the 3 o'clock position.
+  # @param angle2 [Float] The second angle of the arc to the center point, in Radians.
+  # @param [Hash] Style hash
```
jasonrclark: I believe yard also supports describing the `@option` values that are supported in a hash (http://www.rubydoc.info/gems/yard/file/docs/Tags.md#option). Those seem worthwhile to consider listing out too.

tasks/yard.rb
```diff
@@ -4,7 +4,9 @@
   require 'yard'
 
   YARD::Rake::YardocTask.new do |t|
-    t.options = ['-mmarkdown']
+    t.files = ['shoes-core/**/*.rb', 'README.md', 'LICENSE', 'CONTRIBUTING.md',
+               'CODE_OF_CONDUCT.MD']
+    t.options = ['-mmarkdown', '--no-private']
```
jasonrclark: So these are a good start, but I found generating locally that this gives me a bit better result:

```
  YARD::Rake::YardocTask.new do |t|
    t.files = [
      'shoes-core/lib/**/*.rb',
      '-',
      'README.md',
      'shoes-core/README.md',
      'LICENSE',
      'CONTRIBUTING.md',
      'CODE_OF_CONDUCT.MD'
    ]
    t.options = [
      '-mmarkdown',
      '--no-private'
    ]
```

It does a couple things:
* Only generates for the `shoes-core/lib` folder, since we were picking up specs
* Works correctly with the markdown files by putting them after a `-`

How does that look to you @nicholalexander?

jasonrclark:
Sorry for dropping off for so long! Had some business travel, followed by holidays, but I'm back in the Shoes action in earnest now. Getting yard working again is really helpful--it was on my short list to look at soon, and this is a great first step.

Few comments, nothing major, and this should be :cool: to go after those.

nicholalexander:
No worries at all, I was away for the holidays as well.  Will make suggested changes!
Happy new year!
n

nicholalexander:
@jasonrclark - I took your suggested configuration and moved it into a yardopts file, that way you can either run `rake yard` or you can just run `yard`.  I think this is helpful for just jumping into yard as well as having the docs generated on rubydoc.info.  What do you think about that?

shoes-core/lib/shoes/arc.rb
nicholalexander: Cool, will fill out the options for the style hash.

nicholalexander: Hmm, a couple of problems on this one.  Firstly the params aren't showing up in the yard doc because it wants them to be next to an initialize method.

Additionally, the options hash needs to be named in order for it to pick up and match against it.

```
class Blurgh
    # @param thing [Integer] a thing!
    # @param thing_hash [Hash] a thing hash which we want to have.
    # @option thing_hash [String] :blurgh the name of blurgh
    def initialize(thing, thing_hash)
    end
  end
```

I will have to dig deeper into yard to see if I can figure this out...

jasonrclark: Ah, didn't know that about it. I wouldn't worry on it for too long--if you can't find an easy way, we can table the options hashes until later (or never) and just roll forward with this now.

jasonrclark:
@nicholalexander `.yardopts` seems perfect!

Is the only thing outstanding seeing what can be done about the hash options? Excited to get this landed so I can join in writing some docs as I go. 📚

nicholalexander:
Hey @jasonrclark - sorry for the delay, I was away for the long weekend.  This should be good to go as it is now.  I removed the params from the docs because those also weren't showing up.  As it currently stands, yard is configured and working and the class has some initial documentation which generates correctly.  If we want to solve the params and options problem, that could be done in another PR.  

Thanks!

jasonrclark:
No apologies necessary--we're all volunteers here. Thanks for the tidy up, and here comes the merge 🚄 !

PR #1535, opened 2018-01-05
Give more room in redrawing stars

### LINKED ISSUE #1526: Star redrawing missing pixels
Per https://github.com/shoes/shoes4/pull/1525#issuecomment-348401940, seeing some left behind pixels in `samples/simple_star.rb` that are almost certainly related to the redrawing box we use for stars (which is custom, and probably just off by a pixel)

image

jasonrclark:
Fixes #1526

Ended up that after some testing it wasn't just a single pixel--it was related to the strokewidth for the drawing which wasn't being accounted for perfectly.

This may be slightly more generous than it needs to be, but generally redrawing a few more pixels around the edges is better than artifacts!

### DIFF
--- a/shoes-core/lib/shoes/star.rb
+++ b/shoes-core/lib/shoes/star.rb
@@ -37,26 +37,31 @@ class Shoes
         element_top - dy <= y && y <= element_bottom - dy
     end
 
+    # Redrawing needs a bit of extra room. We offset by this factor, then
+    # extend our size by twice that to evenly surround the whole thing.
+    REDRAW_OFFSET_FACTOR = 4
+    REDRAW_SIZING_FACTOR = REDRAW_OFFSET_FACTOR * 2
+
     def redraw_left
       return 0 unless element_left
       calculated_left = element_left
       calculated_left -= width * 0.5 if center
-      calculated_left - strokewidth.ceil * 4
+      calculated_left - strokewidth.ceil * REDRAW_OFFSET_FACTOR
     end
 
     def redraw_top
       return 0 unless element_top
       calculated_top = element_top
       calculated_top -= width * 0.5 if center
-      calculated_top - strokewidth.ceil * 4
+      calculated_top - strokewidth.ceil * REDRAW_OFFSET_FACTOR
     end
 
     def redraw_width
-      element_width + strokewidth.ceil * 8
+      element_width + strokewidth.ceil * REDRAW_SIZING_FACTOR
     end
 
     def redraw_height
-      element_height + strokewidth.ceil * 8
+      element_height + strokewidth.ceil * REDRAW_SIZING_FACTOR
     end
 
     def center_point
--- a/shoes-core/lib/shoes/star.rb
+++ b/shoes-core/lib/shoes/star.rb
@@ -39,28 +39,24 @@ class Shoes
 
     def redraw_left
       return 0 unless element_left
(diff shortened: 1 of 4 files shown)


shoes-core/lib/shoes/star.rb
```diff
@@ -39,28 +39,24 @@ def in_bounds?(x, y)
...
     def redraw_left
       return 0 unless element_left
-      if center
-        element_left - width * 0.5 - style[:strokewidth].to_i
-      else
-        super
-      end
+      calculated_left = super
+      calculated_left = element_left - width * 0.5 if center
+      calculated_left - style[:strokewidth].to_i * 2
```
PragTob: don't we have a `strokewidth` method or something defined that already returns the integer?

jasonrclark: We do indeed have a method and [a style default to 1](https://github.com/shoes/shoes4/blob/master/shoes-core/lib/shoes/common/style.rb#L15), so switching this over. Great catch @PragTob!

jasonrclark:
Hmm, with further testing this approach doesn't seem to be actually working as I thought it was last night--still got a trail on the moving stars. Appears to potentially be because the star moves more than just 1 pixel at a time because of when mouse events fire.

jasonrclark:
So, after further testing noticed several things here:

* Stars actually just needed a bit more. I think it's related to the polygon drawing, but as before--a few more pixels around redrawing is fine
* Tests weren't setting the app level styling appropriately, so fixed that to get tests together
* Stars weren't the only one with the problem, so logged #1536
* Noticed that links aren't picking up the default blue styling they're trying to apply. I tweaked a test for this in f18457ab44f65ed6166a395c9c6bbf2a25281e0c so things pass as they stand right now, but should get fixed with #1537 (and the test changed back).

shoes-core/lib/shoes/star.rb
```diff
@@ -39,28 +39,24 @@ def in_bounds?(x, y)
...
     end
 
     def redraw_width
-      element_width + style[:strokewidth].to_i * 2
+      element_width + strokewidth.ceil * 8
     end
 
     def redraw_height
-      element_height + style[:strokewidth].to_i * 2
+      element_height + strokewidth.ceil * 8
```
PragTob: the 8s and 4s are a bit magical to me here :angel:

shoes-core/spec/shoes/star_spec.rb
```diff
@@ -170,7 +170,7 @@
   end
 
   describe "redrawing region" do
-    subject { Shoes::Star.new(app, parent, 100, 100, 5, 50, 30, center: false, strokewidth: 0) }
+    subject { Shoes::Star.new(app, parent, 100, 100, 5, 50, 30, center: false) }
```
PragTob: as strokewidth now factors in directly, setting it here might help discover what value it is at? (I know there's another test that specifically tests it I thought it might be good for the general case)

jasonrclark: Yeah, that would make it more obvious. I think I was trying to show what the default does, but it makes more sense with the strokewidth's default value (1) called out.

shoes-core/lib/shoes/star.rb
jasonrclark: :+1:

PR #1538, opened 2018-01-09
Fix link default styling

### LINKED ISSUE #1537: Links don't default to blue text
Regression from Shoes 3

```
Shoes.app do
  para link("yO")
end
```

**Shoes 3**
image

**Shoes 4**
image

jasonrclark:
Fixes  #1537

Links weren't defaulting to blue, despite having the color specified in their defaults for `stroke`. The problem was the app-level styling. Order of application for styles was:

* Element class defaults
* App-instance styling
* etc.

But the app instance styles picked up values from the app defaults, and so any key in the element defaults that was also defined in app defaults got tromped!

This establishes the new, and I believe correct, order:

* App defaults (directly, rather than expecting instance to set it)
* Element class defaults
* App instance styles, but only that don't match app defaults!
* etc.

Oh, and the links show blue by default now :)

Deserves some deeper sample testing since this is pretty fundamental, but don't have time for it this evening... will do before merge.

### DIFF
--- a/shoes-core/lib/shoes/common/style.rb
+++ b/shoes-core/lib/shoes/common/style.rb
@@ -35,11 +35,10 @@ class Shoes
       end
 
       def style_init(arg_styles, new_styles = {})
-        default_element_styles = {}
-        default_element_styles = self.class::STYLES if defined?(self.class::STYLES)
         create_style_hash
-        @style.merge!(default_element_styles)
-        merge_app_styles
+        @style.merge!(DEFAULT_STYLES)
+        @style.merge!(self.class::STYLES) if defined?(self.class::STYLES)
+        @style.merge!(applicable_app_styles)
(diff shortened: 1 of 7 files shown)


jasonrclark:
So, a few problems, all look arc related.

`samples/good_arc.rb`
image

`samples/simple_arc.rb`
image

`samples/simple_curve.rb`
image

jasonrclark:
Turns out to be related to the `nofill` and `nostroke` methods, which our changes to the styling order mixed up. Fix on the way that handles explicitly removing those styles since those are the only two that have these type of removal methods at the app level.

jasonrclark:
Sigh... so that borked the clock

`samples/good_clock.rb`
image

jasonrclark:
Clear at last!

Issue #1494, opened 2017-10-18
Certain methods don't work (alert, confirm...)


Jeytas:
I've been having some issues with my Shoes. Certain methods just don't seem to work at all. When I try to create a confirm? it gives me the following error

`NoMethodError: undefined method 'confirm?' for (Shoes::App:0x00000fa0 "Shoes 4"):Shoes::App
Did you mean?  confirm
              method_missing at org/jruby/RubyBasicObject.java:1657
                                    method_missing at /home/marvin/.rvm/gems/jruby-9.1.13.0/gems/shoes-core-4.0.0.pre12/lib/shoes/app.rb:107
                                    method_missing at /home/marvin/.rvm/rubies/jruby-9.1.13.0/lib/ruby/stdlib/forwardable.rb:189
                               block in postscreen at rerutwi.rb:95
                                        eval_block at /home/marvin/.rvm/gems/jruby-9.1.13.0/gems/shoes-swt-4.0.0.pre12/lib/shoes/swt/swt_button.rb:26
  block in _after_do_redefine_method_with_callback at /home/marvin/.rvm/gems/jruby-9.1.13.0/gems/after_do-0.4.0/lib/after_do.rb:99
                                    block in click at /home/marvin/.rvm/gems/jruby-9.1.13.0/gems/shoes-swt-4.0.0.pre12/lib/shoes/swt/swt_button.rb:31
                                    run_event_loop at /home/marvin/.rvm/gems/jruby-9.1.13.0/gems/swt-4.6.1.1/lib/swt/minimal.rb:107
                                        event_loop at /home/marvin/.rvm/gems/jruby-9.1.13.0/gems/swt-4.6.1.1/lib/swt/minimal.rb:90
                                              open at /home/marvin/.rvm/gems/jruby-9.1.13.0/gems/shoes-swt-4.0.0.pre12/lib/shoes/swt/app.rb:71
                                          open_gui at /home/marvin/.rvm/gems/jruby-9.1.13.0/gems/shoes-core-4.0.0.pre12/lib/shoes/internal_app.rb:125
                                        initialize at /home/marvin/.rvm/gems/jruby-9.1.13.0/gems/shoes-core-4.0.0.pre12/lib/shoes/app.rb:52
                                               app at /home/marvin/.rvm/gems/jruby-9.1.13.0/gems/shoes-core-4.0.0.pre12/lib/shoes/app.rb:20
                                            <main> at rerutwi.rb:108`

[...]

PR #1542, opened 2018-01-15
also delegate builtin methods for widgets + URLs

PragTob:
Wowzies, that seems like a pretty big oversight on our part.
Glad we found it finally... maybe we should wip up some sample?

Minimal reproduction script:

```ruby
class AppComponent < Shoes
  url '/', :index
  def index
    alert("doesn't work")
  end
end
Shoes.app width: 800, height: 600, title: 'testing page'
```

side note: It's amazing after a lot of work on big applications and bigger features on benchee how productive one can feel with a system knowledge, not so a big system (but well designed and well tested if I might say so!) and get a critical bugfix in in <20 minutes.

### DIFF
--- a/shoes-core/lib/shoes/app.rb
+++ b/shoes-core/lib/shoes/app.rb
@@ -112,6 +112,7 @@ class Shoes
     # class definitions are evaluated top to bottom, want to have all of them
     # so define at bottom
     DELEGATE_METHODS = ((Shoes::App.public_instance_methods(false) +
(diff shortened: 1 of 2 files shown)


PragTob:
Oh yeah, fixes #1494

jasonrclark:
I'd be 💯  in favor of adding that sample or something like it. I feel like part of the low testing (at least on my part) around widgets and visit style apps is the lack of good samples around it.

PR #1545, opened 2018-01-17
Handle center_points for stars with centered styling.

### LINKED ISSUE #1531: Bug - Centered Stars and center_point=
Setting the center_point of stars that have a 'style[:center] = true' attribute does not act as expected.  This is because when the star is centered the new top and left points should not incorporate the width of the element as that is already being done by the centering style.

See implementation in `Arc#center_point=` as a possible fix.  Introduced in #1525.

nicholalexander:
First I updated the center_point= method to match how it is done with Arcs.  Stars, however, also required changing the reader of the center_point based on styling.

I also updated the simple_star.rb example to include a star that was centered so we can see the correct behavior as well as updated the tests.

I'm not sure why reading the center point is different for stars and arcs with the styling, but it definitely appears to be!

fixes #1531.

### DIFF
--- a/samples/simple_star.rb
+++ b/samples/simple_star.rb
@@ -16,9 +16,23 @@ Shoes.app width: 600, height: 600 do
   radius = 3
   oval (center_point.x - radius), (center_point.y - radius), radius: radius
 
-  # do you want a star to follow you around?
-  @churyumov_gerasimenko = star 200, 100
+  # do you want a well centered star to follow you around?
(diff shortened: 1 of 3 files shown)


jasonrclark:
Looks great here!

I don't understand this well and haven't spent the time to dig into it, but there is some plumbing down in the `Shoes::Dimensions` class around treating start as centered. I suspect that one of the types (probably `Shoes::Arc`) is taking advantage of that, while the other isn't.

I'm not too hung up on it if it's well tested, though. :shipit:

PR #1551, opened 2018-01-28
Yard warnings

jasonrclark:
More for the #620 bucket!

This gets the `bundle exec rake yard` running without warnings.

In addition, I think this will inform the style that I want the various DSL methods to take. Some have previously provided separate overload definitions in the docs, but that gets repetitive, _and_ isn't being done exhaustively. I'd recommend we do like `arc` and `line` in this PR, which looks like this:

image

Seem legit @PragTob ?

### DIFF
--- a/shoes-core/lib/shoes/configuration.rb
+++ b/shoes-core/lib/shoes/configuration.rb
@@ -26,12 +26,15 @@ class Shoes
         end
       end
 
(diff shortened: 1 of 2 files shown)


jasonrclark:
Sounds good @pragtob. Haven’t really found myself waiting much, but will hit you up if I do!

PR #1556, opened 2018-02-19
Curate samples

### LINKED ISSUE #353: Curate samples

jasonrclark:
Fixes #353

At long last, I've plowed through the samples and done a bit of grooming.

Each change is a separate commit... could squash when we merge, but seemed easier to allow for backing anything out folks don't like.

At this point, all samples in the directory are good, and anything not is in `unsupported` so I stopped using the README as a list and started using it as a README :)

Made some minor color tweaks, dropped duplicates, and moved a few things into a `development` folder (for things like positioning as we go, which no one except me cares about the details of)

Also revised naming on the `good` samples so now we're in `simple`, `expert` and `nks` for all the files.

This gets the main sample load under 100, and looking a bit less random... ready for prime time!


samples/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+bundle exec rake samples:all
+```
+
+We have a couple directories:
+
+* `development` contains samples of interest to the Shoes team, but not much for everyone else
+* `packaging` contains subfolders so we can easily test packaging without picking up all the samples in the bundle.
+* `sounds` contains some samples with large audio files we also don't want bundled up in the `shoes` gem
+* `unsupported` contains a small set of samples that older versions of Shoes allowed but aren't anymore. These are kept for historical reference and with hope to someday move them out!
```
PragTob: :+1:

samples/simple_displace.rb
```diff
@@ -1,17 +1,100 @@
...
+            info "#{name} displaces to #{position}"
+            @objects[name].displace(*position)
+          end
+          button "replace" do
+            position = @layout.recall(name)
+            info "#{name} replaces to #{position}"
+            @objects[name].displace(*position)
+          end
+        end
+      end
```
PragTob: this... uhhh exploded in size?

jasonrclark: Odd. I suspect I removed the original simple_displace for some reason, then renamed over it later on. Will review the commits to make sure that was intentional.

jasonrclark: Confirmed... what's now `simple_displace.rb` was `good_displace.rb` before, and the old `simple_displace.rb` was a bouncing button of which we have another similar copy. Relevant commits:
* https://github.com/shoes/shoes4/pull/1556/commits/df8e791d50b7308ec54347c32b4ff86561fc5a51
* https://github.com/shoes/shoes4/pull/1556/commits/59866b3c970fcc7d79a80adfed2bb8af059123ed

PragTob: thanks!

PragTob:
Good from my side, not sure if @jasonrclark still wants to do something here :)

PragTob:
I'm gonna declare he doesn't and if not we can always make a new PR :sweat_smile:

jasonrclark:
I was all set--just been a busy week!--so thank you for landing the PR @PragTob. ✨

PR #1557, opened 2018-02-24
Fast failure handlers

### LINKED ISSUE #1413: Apply fail fast wrapper to all eval blocks in Shoes
Doesn't have to be a single PR. In fact, would be super happy with individual PRs against different spots.

A good example of the basic mechanics of the changes are in https://github.com/shoes/shoes4/pull/1409/commits/d01a6d4a02b065085c8bf102d5812e22af99a798

Will keep this list up to date with the various locations as we find more of them:

* [x] `Common::Changeable`
* [x] App mouse motion
* [x] App hover/leave (handled by underlying Hover module already)
* [x] App keypresses
* [x] Element hover/leave
* [x] Download callbacks (already covered because slot-bound async callbacks are wrapped!)
* [x] click handlers on `Link`, art elements, etc.
* [x] ~~`Shape`~~ Looked fine in re-testing!
* [x] ~~`App#append`~~ Looked find in re-testing!

jasonrclark:
Fixes #1413

Caught all the spots listed in the other issue... possible there are other places, but I don't see them. 💥 

the politics of failure have failed


jasonrclark:
Yeah, if it shows up in performance there are definitely plenty of ways to tackle it. I'm not that concerned because most of the locations aren't super hot paths--they're responding to individual user actions vs things happening during layout.

Thanks for the look @PragTob and will keep those in mind if we regress!

PR #1558, opened 2018-02-26
More docs!

jasonrclark:
More for #620 

This started generating docs for `shoes-swt` and `shoes-package` which we were leaving out. Deals with warnings as a result of those additions.

Also documents the `Shoes::Dimensions` and `Shoes::Dimension` class. Chose not to plumb in every single individual method since the pre-amble description really captures it all more clearly than the method-by-method callouts would without huge amounts of duplication.

### DIFF
--- a/shoes-core/lib/shoes/dimensions.rb
+++ b/shoes-core/lib/shoes/dimensions.rb
@@ -1,51 +1,55 @@
 # frozen_string_literal: true
 
-# Dimensions is a central class that most Shoes classes use to represent their
-# dimensions, e.g. where they are and how much space they are taking up there.
-# All the different position types might be confusing. So here is a little list:
-#
-# Position (left, top, right, bottom)
-# plain (left, top, right, bottom)
-#   An offset relative to the parent (parents mostly are slots e.g.
-#   flows/stacks), e.g it isn't fully positioned/doesn't flow anymore when set
-#
-# absolute (absolute_left, absolute_top, absolute_right, absolute_bottom)
-#   The absolute position of an element in the app, set by positioning code (in
-#   slot.rb). Might not be the beginning of the element as it also takes margins
-#   into account, so it could be the beginning of the margin. Is also used in
-#   the positioning code.
-#
-# element_* (element_left, element_top, element_right, element_bottom)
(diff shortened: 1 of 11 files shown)


shoes-core/lib/shoes/dimensions.rb
```diff
@@ -1,51 +1,55 @@
...
+  # = Position (left, top, right, bottom)
+  #
+  # *plain* (+left+, +top+, +right+, +bottom+): An offset relative to
+  # the parent (parents mostly are slots e.g.  flows/stacks), e.g it isn't
+  # fully positioned/doesn't flow anymore when set
+  #
+  # *absolute_** (+absolute_left+, +absolute_top+, +absolute_right+,
+  # +absolute_bottom+): The absolute position of an element in the app, set
+  # by positioning code (in +slot.rb+). Might not be the beginning of the
+  # element as it also takes margins into account, so it could be the beginning
```
PragTob: >takes margins into account

sounds weird and more like `element_*`  - it sort of doesn't take the margins into account. Dunno what a good description is. "Beginning of the general slot for the element" ?

shoes-core/lib/shoes/dimensions.rb
```diff
@@ -1,51 +1,55 @@
...
+  #
+  # *plain* (+width+, +height+): The whole space taken up by this
+  # element with margins and everything. Used for positioning/by the user.
+  #
+  # *element_** (+element_width+, +element_height+): Just the space
+  # taken up by the element itself without margins. Used by drawing.
+  #
+  # Note that this is NOT how margins work in the CSS box model. We diverge for
+  # reasons mentioned
+  # {here}[https://github.com/shoes/shoes4/pull/467#issuecomment-27655355]
```
PragTob: good reference!

jasonrclark:
You were totally right about that one bit being confusing @PragTob. I've updated it with this which I think is more representative of what's really going on (and contrasts it with element values)

```
  # *absolute_** (+absolute_left+, +absolute_top+, +absolute_right+,
  # +absolute_bottom+): The absolute position an element begins positioning
  # from in the app (in +slot.rb+). Margins are not included in absolute
  # position values--that's the main difference between +absolute+ and
  # +element+ values.
  #

```

PR #1561, opened 2018-04-10
Clickable arc issue#778

estensland:
Opening this for some feedback

For Issue #778

I have it working with most cases I have found. It took a between my schedule, the complexity, and how much of math I have forgotten.

I wanted it to work with samples/simple_arc.rb (if you put a click event on them) and it mostly does. Since those use nofill, but the default is not, I built out the code to handle both cases.

At opening of PR, I think the main code could use some more comments (similar to those added for #center_point). Also, the specs are just what I have been using to get coverage while I refactored. They could use some work to be clearer at what is being tested. However, I wanted to open this early because I am not sure where the code should live. Should it stay in arc.rb or be broken out somewhere else?

Besides just being a large chunk of logic within arc, it could apply to other shapes. Since at its core an arc is an oval, a good part of the methods would be useable in expansions to Oval#in_bounds?


shoes-core/lib/shoes/arc.rb
```diff
@@ -55,5 +58,221 @@ def center_point=(point)
...
+
+    def inner_oval_axis_fraction(axis, input)
+      oval_axis_fraction(axis, input, inner_oval_difference)
+    end
+
+    def inside_oval?(x, y)
+      # https://math.stackexchange.com/questions/76457/check-if-a-point-is-within-an-ellipse
+      #    (x - h)**2    (y - k)**2
+      #     -------    +  -------   <=   1
+      #      Rx ** 2      Ry ** 2
```
jasonrclark: Love love love the reference!

shoes-core/lib/shoes/arc.rb
```diff
@@ -55,5 +58,221 @@ def center_point=(point)
...
+    def adjust_angle(input_angle)
+      # Must pad angle, since angles in standard formulas start at different location than in shoes
+      adjusted_angle = input_angle + 1.5708
+
+      # Must make angle 0..6.28318
+      normalize_angle(adjusted_angle)
+    end
+
+    def y_adjust_negative?(input_angle)
+      ((input_angle >= 0) && (input_angle <= 1.5708)) || ((input_angle >= 4.71239) && (input_angle <= 6.28319))
```
jasonrclark: Could probably benefit from some constants for these values

jasonrclark:
Hey @estensland! Sorry I've been out of touch. Had some eye problems that have cut back on after-hours coding.

This looks awesome, though! While I don't follow all the math perfectly, the way you've broken it up and named the methods I feel confident I could with a little time and reading.

I'll pull this down as soon as I have some time and test a few things out, but this looks well on its way to being ready.

estensland:
@jasonrclark All good I've been there too. It too a lot of iterations and various testing to get it like this...it is possible I didn't check a certain case so please let me know if anything looks off!

estensland:
@jasonrclark I noticed that I had not pushed up everything, I had added a bit more comments. If they are too detailed or what not let me know. I added using constants to them off your comment.

PR #1565, opened 2018-05-24
Travis Updates

PragTob:


PragTob:
Seems the new JRuby has java 8 as a minimum requirement. Seeing that it's from 2011 understandable so should just exclude it when testing Jruby 9.2

PR #1568, opened 2018-10-03
Fix some rubocop offenses

submitnine:
This PR should fix Lint/HandleExceptions and Lint/RescueWithoutErrorClass rubocop offenses.

### DIFF
--- a/app/shoes_app.rb
+++ b/app/shoes_app.rb
@@ -55,7 +55,7 @@ class ShoesApp < Shoes
         destination = ShoesApp.navigation.pop(true)
         app.visit destination
       rescue ThreadError
-        # Empty queue, no worries...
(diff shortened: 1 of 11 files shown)


app/shoes_app.rb
```diff
@@ -55,7 +55,7 @@ def home
         destination = ShoesApp.navigation.pop(true)
         app.visit destination
       rescue ThreadError
-        # Empty queue, no worries...
+        Shoes.logger.info "Empty queue:\n#{e.message}"
```
jasonrclark: This is actually going to log a ton of messages as this queue is more often empty than not. Could we do something like https://github.com/rubocop-hq/rubocop/blob/master/manual/configuration.md#disabling-cops-within-source-code to ignore this specific case from code?

